PR #1, opened 2016-09-27
Report halconfig contents

lwander:
halerror1

Notice that it validates your halconfig & reports failures with line & column numbers pointing to the incorrect fields.

@duftler @jtk54 @ttomsu PTAL


lwander:
@danielpeach

lwander:
https://github.com/spinnaker/halyard/issues/2

Issue #2, opened 2016-09-28
Surface halconfig contents under `/v1/config/*` endpoints


lwander:
This involves:

1. Creating models for each initially supported provider & webhook.
2. Writing controllers that return provider & webhook details at the desired level of granularity.

PR #6, opened 2016-09-28
Provide fields used to configure Kubernetes

lwander:
@duftler @jtk54 @ttomsu @danielpeach  PTAL

Also added a pair of unit tests.

#2

### DIFF
--- a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/DeploymentConfiguration.java
+++ b/halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/DeploymentConfiguration.java
@@ -16,21 +16,18 @@
 
 package com.netflix.spinnaker.halyard.model.v1;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import lombok.Data;
-import lombok.Getter;
-import lombok.Setter;
+import com.netflix.spinnaker.halyard.model.v1.providers.Providers;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 /*
- * Maps the entire contents of ~/.hal/config
+ * A DeploymentConfiguration is an installation of Spinnaker
  */
-@Data
-public class Halconfig {
-  private String halyardVersion;
-  private String currentDeployment;
-  private List<DeploymentConfiguration> deploymentConfigurations = new ArrayList<>();
+public class DeploymentConfiguration {
+  public String name;
+  public String version;
+  public Providers providers;
+  public List<Map> webhooks = new ArrayList<>();
 }
--- a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/Provider.java
+++ b/halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/Provider.java
@@ -14,15 +14,16 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.halyard.controllers.v1;
+package com.netflix.spinnaker.halyard.model.v1.providers;
 
-import org.springframework.http.HttpStatus;
-import org.springframework.web.bind.annotation.ResponseStatus;
-import org.yaml.snakeyaml.error.Mark;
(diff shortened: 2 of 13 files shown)


halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/DeploymentConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * A DeploymentConfiguration is an installation of Spinnaker
+ */
+public class DeploymentConfiguration {
+  public String name;
+  public String version;
+  public Providers providers;
+  public List<Map> webhooks;
+
+  public DeploymentConfiguration() {
```
ttomsu: Constructor isn't necessary, no?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/Provider.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.model.v1.providers;
+
+import java.util.List;
+
+public class Provider<T extends Account> {
+  public boolean enabled;
```
ttomsu: Is there any particular reason you're using public values instead of private + getters/setters?

If not, I strongly recommend private vals + Lombok's `@Data`.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/DockerRegistryReference.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.model.v1.providers.kubernetes;
+
+import java.util.List;
+
+public class DockerRegistryReference {
+  public String accountName;
+  public List<String> namespaces;
```
ttomsu: I also recommend initializing all collections & maps with default empty values. Not doing this literally bit me a few hours ago

halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/Provider.java
lwander: As a beginner - what's the detriment of public values?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/DeploymentConfiguration.java
lwander: Good catch

halyard-web/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/Provider.java
ttomsu: I'm going to augment your question to be "what's the detriment to public _fields within data classes_"

A few reasons that by themselves aren't convincing, but altogether seem compelling to me:

1. External classes coupling directly to a field mean you don't know who needs access. Put another way, you can't set a debug breakpoint on a getter and track back to where the data is giong.
2. External classes can change the value. [This post](http://stackoverflow.com/questions/7959129/whats-the-deal-with-javas-public-fields/7959871#7959871) advocates making public fields final, which I think isn't a bad option.
3. The inability to modify the value on the fly (via overridden getter) should your future needs change.
4. If you ever decide to build up a hierarchy interfaces for related objects, your interfaces can only specify getters/setters. If that happens, you now have two ways to access your field - directly and through the getter, and that will cause confusion later on.

lwander: Thanks, this is helpful. I'll look into Lombok's `@Data`.

lwander:
@ttomsu all comments addressed.

PR #11, opened 2016-10-13
Added file & null value validation

lwander:
I also changed the validator error message to omit the field name explicitly, and rather format the error message to contain the field name when applying all validators.

@duftler or @ttomsu or @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.halyard.model.v1.providers.kubernetes;
 
 import com.netflix.spinnaker.halyard.model.v1.providers.Account;
 import com.netflix.spinnaker.halyard.validate.v1.ValidateField;
+import com.netflix.spinnaker.halyard.validate.v1.ValidateNotNull;
+import com.netflix.spinnaker.halyard.validate.v1.providers.ValidateFileExists;
 import com.netflix.spinnaker.halyard.validate.v1.providers.kubernetes.ValidateKubernetesContext;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
@@ -32,6 +34,9 @@ public class KubernetesAccount extends Account implements Cloneable {
   String context;
   String cluster;
   String user;
+  @ValidateField(validators = {ValidateFileExists.class})
+  String kubeconfigFile;
   List<String> namespaces = new ArrayList<>();
+  @ValidateField(validators = {ValidateNotNull.class})
   List<DockerRegistryReference> dockerRegistries = new ArrayList<>();
 }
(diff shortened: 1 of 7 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -32,6 +34,9 @@
   String context;
   String cluster;
   String user;
+  @ValidateField(validators = {ValidateFileExists.class})
+  String kubeconfigFile;
   List<String> namespaces = new ArrayList<>();
+  @ValidateField(validators = {ValidateNotNull.class})
```
ttomsu: Looks to me like you're including an empty list anyway, so is there a case where this validator would ever fire? Perhaps a `ValidatorNotEmpty` is needed also/instead?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/validate/v1/ValidateFieldRegex.java
```diff
@@ -33,7 +33,7 @@ protected ValidateFieldRegex(String subject, String pattern, String patternDescr
   public Stream<String> validate() {
     boolean matches = Pattern.matches(pattern, subject);
     if (!matches) {
-      return Stream.of(String.format("%s (\"%s\" failed to match %s)", patternDescription, subject, pattern));
+      return Stream.of(String.format("%s (\"%s\" failed to match \"%s\")", patternDescription, subject, pattern));
```
ttomsu: nit: Are the double quotes inside truly necessary? It would make the string easier to read if you used single, unescaped quotes:

`String.format("%s ('%s' failed to match '%s')"`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/validate/v1/providers/ValidateFileExists.java
```diff
@@ -0,0 +1,47 @@
...
+    super(subject);
+  }
+
+  @Override
+  public Stream<String> validate() {
+    try {
+      BufferedReader reader = new BufferedReader(new FileReader(new File(subject)));
+      reader.read();
+    } catch (FileNotFoundException e) {
+      return Stream.of(String.format("Unable to open file \"%s\".", subject));
```
ttomsu: nit: "Unable to open" doesn't directly map to "File not found" in my head. I'd keep the "File not found" language

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
lwander: It would fire if someone tries to delete delete the list of `dockerRegistries` explicitly.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/validate/v1/providers/ValidateFileExists.java
lwander: Good point

halyard-config/src/main/java/com/netflix/spinnaker/halyard/validate/v1/ValidateFieldRegex.java
ttomsu: Meh... I use it all over Fiat and have never had a problem. No biggie though.

PR #13, opened 2016-10-13
Add doxygen style comments

lwander:
`@` annotations aren't punctuated when they refer to a class/method/exception/field.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
@@ -30,18 +30,37 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 
-/*
+/**
+ * A parser for all Config read by Halyard at runtime.
+ *
+ * @see Halconfig
+ *
  * Since we aren't relying on SpringBoot to configure Halyard's ~/.hal/config, we instead use this class as a utility
  * method to read ~/.hal/config's contents.
  */
 @Component
 public class HalconfigParser {
+  /**
+   * Path to where the halconfig file is located.
+   *
+   * @param path Defaults to ~/.hal/config.
+   * @return The path with home (~) expanded.
+   */
   @Bean
   String halconfigPath(@Value("${halconfig.filesystem.halconfig:~/.hal/config}") String path) {
     path = path.replaceFirst("^~", System.getProperty("user.home"));
     return path;
   }
 
+  /**
+   * Version of halyard.
+   *
+   * This is useful for implementing breaking version changes in Spinnaker that need to be migrated by some tool
+   * (in this case Halyard).
+   *
+   * @param version The version (seems like the i.d. function for Spring Boot).
+   * @return The version.
+   */
   @Bean
   String halyardVersion(@Value("${Implementation-Version:unknown}") String version) {
     return version;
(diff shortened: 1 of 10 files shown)


lwander:
@duftler PTAL

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
```diff
@@ -68,6 +95,14 @@ Halconfig parseConfig(InputStream is) throws UnrecognizedPropertyException {
     }
   }
 
+  /**
+   * Returns the current halconfig stored at the halconfigPath
```
duftler: Missing trailing period.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -26,8 +26,25 @@
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+/**
+ * This allows fields making use of validators to be validated.
+ *
+ * This is mean to be implemented by all halconfig contents meant to be updated. It reads the Validate* anotations, and
```
duftler: s/mean/meant

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -26,8 +26,25 @@
...
+ */
 public interface Updateable {
-  /* Attempts to update field to value, but will be rejected if the validation based on `context fails */
+  /**
+   * Attempts to update field to value, but will be rejected if the validation fails.
+   *
+   * @param fieldName Name of the field to be updated.
+   * @param value The value to update the named field to.
+   * @param valueType The type of value being updated (can't be inferred when value is null).
+   * @return The list of errors when validation fails (or null/[] when there are no failures).
```
duftler: Is there some condition that causes `null` to be returned in some cases, and `[]` in others?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -52,6 +69,15 @@
     return errors;
   }
 
+  /**
+   * Validate a given field without updating it.
+   *
+   * @param fieldName Name of the field to be validated.
+   * @param valueType The type of value being validated (can't be inferred when value is null).
+   * @return The list of errors when validation fails (or null/[] when there are no failures).
```
duftler: Same question here.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/validate/v1/providers/ValidateFileExists.java
```diff
@@ -21,6 +21,9 @@
 import java.io.*;
 import java.util.stream.Stream;
 
+/**
+ * Ensure that the given file exists, and is readable.
```
duftler: I noticed several 'types' of validator descriptions:
"Generic validator that accepts/rejects..."
"Simple regex to see if..."
"Ensure that the given file exists..."

Might want to pick one style and use it throughout (I prefer the last style).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
lwander: No, it's `null` throughout, but the stream processing behaves the same for `null` and `[]`. It seemed like wasted effort to validate that it's one and not the other, but I'll add the checks if you think it's valuable.

lwander: Talked offline and made the return explicitly `[]`.

PR #17, opened 2016-10-17
Add class-wide validation for accounts

lwander:
@duftler or @ttomsu or @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
@@ -16,10 +16,13 @@
 
 package com.netflix.spinnaker.halyard.model.v1;
 
+import com.netflix.spinnaker.halyard.model.v1.providers.Account;
+import com.netflix.spinnaker.halyard.validate.v1.ValidateAccount;
 import com.netflix.spinnaker.halyard.validate.v1.ValidateField;
 import com.netflix.spinnaker.halyard.validate.v1.Validator;
 
 import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -100,4 +105,39 @@
...
+   * Validate this entire class
+   *
+   * @return The list of errors when validation fails ([] when there are no failures and validation has passed).
+   * @throws IllegalAccessException
+   * @throws NoSuchFieldException
+   */
+  default public List<String> validate() throws IllegalAccessException, NoSuchFieldException {
+    List<String> errors = new ArrayList<>();
+
+    if (this instanceof Account) {
```
jtk54: Will this only apply to `Account`s? Why is this in the interface? Seems like this ought to be in `Account`, particularly if you are planning on using `Updateable` for other things.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -100,4 +105,39 @@
...
+          .map(v -> (ValidateAccount) v)
+          .map(ValidateAccount::validators)                                      // Pick of the validators
+          .flatMap(Stream::of)                                                   // Flatten the stream of lists
+          .map(v -> {
+            try {
+              return v.getConstructor(this.getClass()).newInstance(this);        // Construct the validators
+            } catch (Exception e) {
+              throw new RuntimeException(e);
+            }
+          }).filter(v -> !v.skip())                                              // Ignore skippable validators
```
jtk54: Maybe put the `.filter` on a new line for consistency. Just a nit though.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
lwander: Good thinking

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
lwander: This is how the intellij settings shipping with gradle are formatting it, and I'd rather rely on that w/o too much thought about formatting by hand.

jtk54: That's ok with me.

PR #19, opened 2016-10-19
Adds endpoints for showing configured spinnaker deployments

lwander:
@ttomsu or @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
@@ -39,4 +39,12 @@ public class KubernetesAccount extends Account implements Cloneable {
   List<String> namespaces = new ArrayList<>();
   @ValidateField(validators = {ValidateNotNull.class})
   List<DockerRegistryReference> dockerRegistries = new ArrayList<>();
+
+  public String getKubeconfigFile() {
+    if (kubeconfigFile == null || kubeconfigFile.isEmpty()) {
+      return System.getProperty("user.home") + "/.kube/config";
+    } else {
+      return kubeconfigFile;
+    }
+  }
 }
--- a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ConfigController.java
+++ b/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ConfigController.java
@@ -37,4 +37,14 @@ public class ConfigController {
   Halconfig config() throws UnrecognizedPropertyException {
     return halyardConfig.getConfig();
   }
+
+  @RequestMapping(value = "/currentDeployment", method = RequestMethod.GET)
+  String currentDeployment() throws UnrecognizedPropertyException {
+    Halconfig halconfig = halyardConfig.getConfig();
+    if (halconfig != null) {
+      return halconfig.getCurrentDeployment();
+    } else {
+      return null;
+    }
+  }
 }
--- a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
+++ b/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.controllers.v1;
+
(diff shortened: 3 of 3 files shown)


lwander:
deployment

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -39,4 +39,12 @@
   List<String> namespaces = new ArrayList<>();
   @ValidateField(validators = {ValidateNotNull.class})
   List<DockerRegistryReference> dockerRegistries = new ArrayList<>();
+
+  public String getKubeconfigFile() {
+    if (kubeconfigFile == null || kubeconfigFile.isEmpty()) {
+      return System.getProperty("user.home") + "/.kube/config";
+    } else {
```
ttomsu: No need for `else`

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ConfigController.java
```diff
@@ -37,4 +37,14 @@
   Halconfig config() throws UnrecognizedPropertyException {
     return halyardConfig.getConfig();
   }
+
+  @RequestMapping(value = "/currentDeployment", method = RequestMethod.GET)
+  String currentDeployment() throws UnrecognizedPropertyException {
+    Halconfig halconfig = halyardConfig.getConfig();
```
ttomsu:
```
Halconfig halconfig = halyardConfig.getConfig();
return halconfig == null ? null : halconfig.getCurrentDeployment();
```

Makes you miss groovy, doesn't it?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -0,0 +1,90 @@
...
+@RestController
+@RequestMapping("/v1/config/deployments")
+public class DeploymentController {
+  @Autowired
+  HalconfigParser halyardConfig;
+
+  @RequestMapping(value = "/{deployment:.+}", method = RequestMethod.GET)
+  DeploymentConfiguration deploymentConfigurations(@PathVariable String deployment) throws UnrecognizedPropertyException {
+    Halconfig halconfig = halyardConfig.getConfig();
+    if (halconfig != null) {
```
ttomsu: Prefer early exits to deeply nested if blocks:

```
Halconfig halconfig = halyardConfig.getConfig();
if (halconfig == null) {
  return null
}
```

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -0,0 +1,90 @@
...
+      List<DeploymentConfiguration> matching = halconfig.getDeploymentConfigurations()
+          .stream()
+          .filter(d -> d.getName().equals(deployment))
+          .collect(Collectors.toList());
+
+      if (matching.size() == 0) {
+        throw new DeploymentNotFoundExecption(deployment);
+      } else if (matching.size() > 1) {
+        throw new DuplicateDeploymentException(deployment);
+      } else {
```
ttomsu: Drop `else` block

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -0,0 +1,90 @@
...
+    if (halconfig != null) {
+      List<DeploymentConfiguration> matching = halconfig.getDeploymentConfigurations()
+          .stream()
+          .filter(d -> d.getName().equals(deployment))
+          .collect(Collectors.toList());
+
+      if (matching.size() == 0) {
+        throw new DeploymentNotFoundExecption(deployment);
+      } else if (matching.size() > 1) {
+        throw new DuplicateDeploymentException(deployment);
```
ttomsu: Seems odd to me to throw an exception on a simple GET command, even if there are multiple matches. Are there other ways to enforce uniqueness of the `deployment` id?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -0,0 +1,90 @@
...
+      }
+    } else {
+      return null;
+    }
+  }
+
+  @RequestMapping(value = "/", method = RequestMethod.GET)
+  List<DeploymentConfiguration> deploymentConfigurations() throws UnrecognizedPropertyException {
+    Halconfig halconfig = halyardConfig.getConfig();
+    if (halconfig != null) {
```
ttomsu: Simplify like above

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -0,0 +1,90 @@
...
+      return halconfig.getDeploymentConfigurations();
+    } else {
+      return null;
+    }
+  }
+
+  @Data
```
ttomsu: These aren't @Data classes, and probably don't need the equals and hash code either. You probably just want to make deployment name `private final` and use `@RequiredArgsConstructor`

lwander: `deploymentName` needs to be accessible for a user to make sense of what's wrong, so they are `@Data`. `EqualsAndHashCode` comes from the fact that I don't use the super, so it shouldn't be taken into account when performing equality checks.

PR #20, opened 2016-10-20
Allow validation based on halconfig context

lwander:
This is necessary for validations that occur across accounts, i.e. Kubernetes depends on Docker Registries.

@duftler @ttomsu @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/FieldReference.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/FieldReference.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.model.v1;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A reference to a field in halconfig.
+ * @param <T> The type of the update
+ */
+@Data
+@NoArgsConstructor
+public class FieldReference<T> extends Reference<T> {
+
+  /**
+   * The name of the field being updated.
+   */
+  private String fieldName;
+}
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Reference.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Reference.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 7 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Reference.java
```diff
@@ -0,0 +1,37 @@
...
+ * A reference to a class in halconfig.
+ *
+ * @param <T> Type of the reference
+ */
+@Data
+public class Reference<T> {
+  /**
+   * The optional value of the reference. May be null.
+   */
+  private T value;
```
jtk54: Why not use `Optional`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Updateable.java
```diff
@@ -75,45 +71,45 @@
   /**
    * Validate a given field without updating it.
    *
-   * @param fieldName Name of the field to be validated.
-   * @param valueType The type of value being validated (can't be inferred when value is null).
+   * @param field The Field being valiated.
```
jtk54: s/valiated/validated.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Reference.java
jtk54: I believe that would remove the need for `valueType`.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/FieldReference.java
```diff
@@ -0,0 +1,33 @@
...
+
+import lombok.Data;
+
+/**
+ * A reference to a field in halconfig.
+ * @param <T> The type of the update
+ */
+@Data
+public class FieldReference<T> extends Reference<T> {
+  public FieldReference() {}
```
ttomsu: Alternatively, lombok's `@NoArgsConstructor`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/Reference.java
lwander: It's a good point, but this value just gets passed transparently to one of halconfig fields. Right now I never have to check if it's null, so wrapping it in `Optional` means extra lines of code.

jtk54: `Optional` also doesn't get rid of the extra field either -_-.

lwander: Unfortunately :(

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/model/v1/UpdateableSpec.groovy
```diff
@@ -168,3 +180,23 @@ public class AccountValidator extends Validator<MyAccount> {
...
+  Stream<String> validate() {
+    if (context.halyardVersion == null) {
+      return Stream.of(UpdateableSpec.ERROR_MESSAGE)
+    } else {
+      return null
+    }
+  }
+
+  @Override
+  boolean skip() {
```
ttomsu: Unrelated to this PR, seems like this should be `enabled`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/model/v1/FieldReference.java
lwander: Ooh, very nice.

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/model/v1/UpdateableSpec.groovy
lwander: Another reason was that (at least for me) I think of "enable" as something determined by the user, external to what they are enabling/disabling. In this case, the validator itself determines if it should skip a validation, so it feels weird to have it "enable/disable" itself.

PR #21, opened 2016-11-07
Propagate errors & warnings to client

lwander:
@duftler or @jtk54 or @ttomsu PTAL

Spring by default only shows the `message` an exception was instantiated with and thrown when handling a response. I added a `HalconfigExceptionHandler` to catch all `HalconfigException`s and return the `errors` & `warnings` they contain.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
@@ -18,12 +18,16 @@ package com.netflix.spinnaker.halyard.config.v1;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import com.netflix.spinnaker.halyard.errors.v1.config.NoConfigException;
+import com.netflix.spinnaker.halyard.errors.v1.config.ParseConfigException;
 import com.netflix.spinnaker.halyard.model.v1.Halconfig;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.parser.ParserException;
+import org.yaml.snakeyaml.scanner.ScannerException;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -103,14 +107,21 @@ public class HalconfigParser {
    * @return The fully parsed halconfig.
    * @throws UnrecognizedPropertyException
    */
-  public Halconfig getConfig() throws UnrecognizedPropertyException {
+  public Halconfig getConfig() {
     Halconfig res = null;
     try {
       InputStream is = new FileInputStream(new File(halconfigPath));
       res = parseConfig(is);
     } catch (FileNotFoundException e) {
-      res = new Halconfig().setHalyardVersion(halyardVersion);
+      throw new NoConfigException(halconfigPath);
+    } catch (UnrecognizedPropertyException e) {
+      throw new ParseConfigException(e);
+    } catch (ParserException e) {
+      throw new ParseConfigException(e);
+    } catch (ScannerException e) {
+      throw new ParseConfigException(e);
     }
+
     return res;
   }
 }
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigException.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigException.java
@@ -14,23 +14,26 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.halyard.model.v1;
+package com.netflix.spinnaker.halyard.errors.v1;
 
-import lombok.Data;
-import lombok.EqualsAndHashCode;
-import lombok.NoArgsConstructor;
+import lombok.Getter;
 
 /**
- * A reference to a field in halconfig.
- * @param <T> The type of the update
+ * This is the base exception class that needs to be thrown by all validators.
  */
-@Data
-@NoArgsConstructor
-@EqualsAndHashCode(callSuper = false)
-public class FieldReference<T> extends Reference<T> {
+public class HalconfigException extends RuntimeException {
+  @Getter
+  protected HalconfigFixableIssue issue = new HalconfigFixableIssue();
 
-  /**
-   * The name of the field being updated.
-   */
-  private String fieldName;
+  public void addError(String error) {
+    issue.getErrors().add(error);
+  }
+
+  public void addWarning(String warning) {
+    issue.getWarnings().add(warning);
+  }
+
+  public HalconfigException() {
+    super();
+  }
 }
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/config/ParseConfigException.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/config/ParseConfigException.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.errors.v1.config;
+
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import com.netflix.spinnaker.halyard.errors.v1.HalconfigException;
+import org.yaml.snakeyaml.parser.ParserException;
+import org.yaml.snakeyaml.scanner.ScannerException;
+
+public class ParseConfigException extends HalconfigException {
+  public ParseConfigException(UnrecognizedPropertyException e) {
+    addError("Unrecognized property in your halconfig: " + e.getMessage());
+  }
+
+  public ParseConfigException(ParserException e) {
+    addError("Could not parse your halconfig: " + e.getMessage());
+  }
+
+  public ParseConfigException(ScannerException e) {
+    addError("Could not parse your halconfig: " + e.getMessage());
+  }
+}
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.services.v1;
+
+import com.netflix.spinnaker.halyard.config.v1.HalconfigParser;
+import com.netflix.spinnaker.halyard.model.v1.Halconfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * This class is meant to be autowired into any controller or service that needs to read the current halconfig.
+ */
+@Component
+public class ConfigService {
+  @Autowired
+  HalconfigParser halconfigParser;
+
+  public Halconfig getConfig() {
+    return halconfigParser.getConfig();
+  }
+
+  public String getCurrentDeployment() {
+    return getConfig().getCurrentDeployment();
+  }
+}
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/DeploymentService.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/DeploymentService.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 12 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
```diff
@@ -103,14 +107,21 @@ Halconfig parseConfig(InputStream is) throws UnrecognizedPropertyException {
...
       res = parseConfig(is);
     } catch (FileNotFoundException e) {
-      res = new Halconfig().setHalyardVersion(halyardVersion);
+      throw new NoConfigException(halconfigPath);
+    } catch (UnrecognizedPropertyException e) {
+      throw new ParseConfigException(e);
+    } catch (ParserException e) {
+      throw new ParseConfigException(e);
+    } catch (ScannerException e) {
+      throw new ParseConfigException(e);
```
ttomsu: Java 7 introduced multiple exception type handling:

http://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigAlert.java
```diff
@@ -0,0 +1,42 @@
...
+   * lists of regions/namespaces not supported by the cloud provider, unresolved references to other configuration, etc...
+   */
+  protected List<String> errors = new ArrayList<>();
+  /**
+   * Warnings are things that Halyard thinks might be a problem. This includes credentials with leading/trailing spaces,
+   * an apparent but uncheckable lack of IAM permissions, etc...
+   */
+  protected List<String> warnings = new ArrayList<>();
+
+  public HalconfigAlert() { }
```
ttomsu: Empty constructor is implied if no other constructor is available

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigException.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.netflix.spinnaker.halyard.errors.v1;
+
+import lombok.Data;
+
+/**
+ * This is the base exception class that needs to be thrown by all validators.
+ */
+@Data
```
ttomsu: I still think @Data here is extraneous - a @Getter on HalconfigAlert seems more appropriate

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -22,69 +22,76 @@
...
     Halconfig halconfig = halyardConfig.getConfig();
-    if (halconfig != null) {
-      List<DeploymentConfiguration> matching = halconfig.getDeploymentConfigurations()
-          .stream()
-          .filter(d -> d.getName().equals(deployment))
-          .collect(Collectors.toList());
+    List<DeploymentConfiguration> matching = halconfig.getDeploymentConfigurations()
+        .stream()
+        .filter(d -> d.getName().equals(deployment))
+        .collect(Collectors.toList());
```
ttomsu: https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Iterables.html#getOnlyElement-java.lang.Iterable- has a neat example of using `MoreCollectors.onlyElement()`

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
```diff
@@ -22,69 +22,76 @@
...
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.*;
 
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
 import java.util.List;
 import java.util.stream.Collectors;
 
 @RestController
+@NoArgsConstructor
```
ttomsu: O_o

Any use for this?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ProviderController.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/v1/config/deployments/{deployment:.+}")
+public class ProviderController {
+  @Autowired
+  HalconfigParser halyardConfig;
+
+  @Autowired
+  DeploymentController deploymentController;
```
ttomsu: Generally you want to abstract the core, shareable logic into Services that are autowired/injected rather than the controllers themselves. See Gate for an example of this pattern.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ProviderController.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @RequestMapping(value = "/providers", method = RequestMethod.GET)
+  Providers providers(@PathVariable String deployment) throws UnrecognizedPropertyException {
+    DeploymentConfiguration deploymentConfiguration = deploymentController.getDeploymentConfiguration(deployment);
+    return null;
```
ttomsu: Should this be `return deploymentController.getDeploymentConfiguration(deployment);` ?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+@ControllerAdvice
+public class HalconfigExceptionHandler extends ResponseEntityExceptionHandler {
```
ttomsu: I don't think you need to poke this far into the Spring guts to do what you need. Checkout https://github.com/ttomsu/gate/blob/76ccad0da6aafc713d2e7fba28dc4c48de9588f9/gate-web/src/main/groovy/com/netflix/spinnaker/gate/config/GateWebConfig.groovy#L64-L64 for a less framework-coupled example.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
lwander: If I use that, `ParseConfigException` is handed an `Exception`, and I can't delineate between errors in my Constructor without inspecting the class type, which feels messy.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DeploymentController.java
lwander: Ah whoops, the DeploymentController was meant for a later PR and is not ready. Let me fix this up.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigExceptionHandler.java
lwander: Hmm, that other example feels just as coupled (accepts a bunch of framework specific arguments, needs special annotations)

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ProviderController.java
lwander: Ah, wait this is the unfinished one, the DeploymentController above should have been OK

jtk54: Was this meant to be a stub, or just unfinished?

lwander:
@ttomsu, I pulled all the retrieval logic for configs & deployments into services (looks much nice, nice suggestion), tidied up the access rights to the `HalconfigException` and removed unnecessary constructors & annotations.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigAlert.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.netflix.spinnaker.halyard.errors.v1;
+
+import lombok.Data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This is what Halyard will return to a client in case of an error fixable by the client. This includes validation
```
duftler: Would it make more sense to call this something like "HalconfigFixableError"? "Alert" is an overloaded term anyway.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/config/ParseConfigException.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.netflix.spinnaker.halyard.errors.v1.config;
+
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import com.netflix.spinnaker.halyard.errors.v1.HalconfigException;
+import org.yaml.snakeyaml.parser.ParserException;
+import org.yaml.snakeyaml.scanner.ScannerException;
+
+public class ParseConfigException extends HalconfigException {
```
duftler: If you're not planning on doing more work in these constructors, or calling them from multiple locations, maybe you can just pass the string error messages from the call-sites and have one constructor?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
```diff
@@ -0,0 +1,37 @@
...
+  @Autowired
+  HalconfigParser halconfigParser;
+
+  public Halconfig config() {
+    return halconfigParser.getConfig();
+  }
+
+  public String currentDeployment() {
+    Halconfig halconfig = halconfigParser.getConfig();
+    return halconfig.getCurrentDeployment();
```
duftler: Maybe `return config().getCurrentDeployment()`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
```diff
@@ -0,0 +1,37 @@
...
+import com.netflix.spinnaker.halyard.model.v1.Halconfig;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConfigService {
+  @Autowired
+  HalconfigParser halconfigParser;
+
+  public Halconfig config() {
```
duftler: Did you avoid naming this `getConfig` so it would be more apparent that there might be some side-effect from calling it?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/DeploymentService.java
```diff
@@ -0,0 +1,54 @@
...
+  HalconfigParser halconfigParser;
+
+  public DeploymentConfiguration deploymentConfiguration(String deployment) {
+    Halconfig halconfig = halconfigParser.getConfig();
+    List<DeploymentConfiguration> matching = halconfig.getDeploymentConfigurations()
+        .stream()
+        .filter(d -> d.getName().equals(deployment))
+        .collect(Collectors.toList());
+
+    if (matching.size() == 0) {
```
duftler: Could use a `switch` here instead ('0', '1', 'default'). Would also let you avoid potentially calling `matching.size()` twice.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigAlert.java
duftler: They sound ok, except they leave out the `Fixable` part. I don't feel too strongly about it though.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/config/ParseConfigException.java
duftler: Seems like a wash, as opposed to being easier. And the content of the error messages would be close to the source, which might make it easier later on to troubleshoot. But also don't feel too strongly about it.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
duftler: In that case, this class could probably use some javadocs.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigAlert.java
duftler: `HalconfigFixableIssue` sounds ok I think.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
ttomsu: You can still combine the exceptions that have the same handling:

```
 } catch (FileNotFoundException e) {
  throw new NoConfigException(halconfigPath);
 } catch (UnrecognizedPropertyException|ParserException|ScannerException e) {
   throw new ParseConfigException(e);
 }
```

halyard-config/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigAlert.java
ttomsu: I still think "Alert" better conveys the semantics of warnings+errors. "FixableIssue" sounds weird to me

halyard-web/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigExceptionHandler.java
ttomsu: `HttpServletResponse` is a Java standard servlet object. The only additional annotation would be the `@ResponseBody`, which would more clearly convey what you're trying to return, which I think is the HalconfigAlert, but I'm not certain. So in short, I think your method could be simply:

```
@ExceptionHandler(UpstreamBadRequest)
public HalconfigAlert handleHalconfigException(HttpServletResponse response, HalconfigException exception) {
  response.setStatus(HttpStatus.CONFLICT)
  return exception.getAlert()
}


@ResponseBody
public class HalconfigAlert {
...
}
```

halyard-config/src/main/java/com/netflix/spinnaker/halyard/services/v1/ConfigService.java
ttomsu: I'd lean toward using the `get` prefix, myself.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/v1/HalconfigParser.java
lwander: Resolved offline - it squashes `e` into the superclass of all those exceptions making the overloaded constructors useless.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/errors/v1/HalconfigExceptionHandler.java
lwander: For future reference `@ResponseBody` is needed on the method handling the exception

ttomsu: Are you using Spring 4+?

From the javadocs of `@ResponseBody`:

>  As of version 4.0 this annotation can also be added on the type level in which case it is inherited and does not need to be added on the method level.

PR #22, opened 2016-11-08
Added CLI package & refactored config

lwander:
I added a proof of concept CLI to as a `halyard-cli` package, and moved the `halyard-config` package up on directory into `config`, since I realized all `$project-$package` projects have their code in `$package` (to avoid having subproject's packages collide).


lwander:
@duftler PTAL, this should be a quickie

PR #24, opened 2016-11-09
(cli) added command parsing framework

lwander:
I'm breaking the actual communication with the daemon into the next PR, but this enables commands like

```
hal config show --provider=kubernetes

hal -v --color=false config

hal --help
```

etc, etc...

@duftler @ttomsu @jtk54 PTAL

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1;
+
+import com.beust.jcommander.Parameters;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This is a top-level command for dealing with your halconfig.
+ *
+ * Usage is `$ hal config`
+ */
+@Parameters(commandDescription = "Configure and view your halconfig")
+public class ConfigCommand extends NestableCommand {
+  @Getter(AccessLevel.PROTECTED)
+  private Map<String, NestableCommand> subcommands = new HashMap<>();
+
+  @Getter(AccessLevel.PROTECTED)
+  private String commandName = "config";
+
+  public ConfigCommand(GlobalOptions globalOptions) {
+    super(globalOptions);
+
+    ShowCommand show = new ShowCommand(globalOptions);
+    this.subcommands.put(show.getCommandName(), show);
+  }
+
+  @Override
+  protected void executeThis() {
+
+  }
+}
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GlobalOptions.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GlobalOptions.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import lombok.Data;
+
+/**
+ * This is the collection of general, top-level flags to be interpreted by halyard.
+ */
+@Data
+@Parameters(separators = "=")
+public class GlobalOptions {
+  @Parameter(names = { "-v", "--verbose" }, description = "Enable verbose output")
+  private boolean verbose = false;
+
+  @Parameter(names = { "-c", "--color" }, description = "Enable terminal color output.", arity = 1)
+  private boolean color = true;
+}
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This command displays its parent's resource.
+ *
+ * Usage is `$ hal config show`
+ */
+@Parameters(commandDescription = "Show the resource in question", separators = "=")
+public class ShowCommand extends NestableCommand {
+  @Getter(AccessLevel.PROTECTED)
+  private Map<String, NestableCommand> subcommands = new HashMap<>();
+
+  @Getter(AccessLevel.PROTECTED)
+  private String commandName = "show";
+
+  @Parameter(names = "--provider", description = "Select the provider's config to display", arity = 1)
+  protected String provider;
+
+  public ShowCommand(GlobalOptions globalOptions) {
+    super(globalOptions);
+  }
+
+  @Override
+  protected void executeThis() {
+
+  }
+}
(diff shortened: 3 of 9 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
```diff
@@ -0,0 +1,56 @@
...
+public class ShowCommand extends NestableCommand {
+  Map<String, NestableCommand> subcommands;
+
+  public ShowCommand(GlobalOptions globalOptions) {
+    super(globalOptions);
+    this.subcommands = new HashMap<>();
+  }
+
+  @Override
+  protected void executeThis() {
```
jtk54: This is stubbed until the next PR, correct? Just checking : D

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+@Parameters(commandDescription = "Configure and view your halconfig")
+public class ConfigCommand extends NestableCommand {
+  private Map<String, NestableCommand> subcommands;
+
+  public ConfigCommand(GlobalOptions globalOptions) {
+    super(globalOptions);
+
+    ShowCommand show = new ShowCommand(globalOptions);
+    this.subcommands = new HashMap<>();
```
lwander: Java infers the type of for the constructor so what you suggested is exactly what will be instantiated. Are you making the suggestion for reading clarity?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GlobalOptions.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import lombok.Data;
+
+/**
+ * This is the collection of general, top-level flags to be interpreted by halyard.
+ */
+@Data
+@Parameters(separators = "=")
```
lwander: No `=` is optional in flags, but without it, things like `--color=false` will fail to parse which feels weird.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
lwander: Yup!

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GlobalOptions.java
lwander: Personally, I'm not sure where to comment this in the code. On one hand it's in the library documentation, but on the other hand I 100% agree it looks like `=` is now required in your CLI flags which is misleading.

jtk54: Yep agreed this is a little confusing, so both `--flag value` and `--flag=value` are good now? That seems correct to me, but we should document this either around the annotations or the flags (or block of flags) themselves.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
ttomsu: Is there ever a case where `subcommands` should be/is expected to be null instead of an empty list? If not, I'd suggest putting the `new HashMap` on the field declaration.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    this.subcommands.put(show.getCommandName(), show);
+  }
+
+  @Override
+  protected void executeThis() {
+
+  }
+
+  @Override
+  protected Map<String, NestableCommand> getSubcommands() {
```
ttomsu: `@Getter(AccessLevel.PROTECTED)` on field?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
```diff
@@ -0,0 +1,55 @@
...
+    return subcommands;
+  }
+
+  @Override
+  protected String getCommandName() {
+    return "config";
```
ttomsu: I hate getters/setters

```
@Getter(AccessLevel.PROTECTED)
private final String commandName = "config"
```

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
```diff
@@ -0,0 +1,56 @@
...
+    return subcommands;
+  }
+
+  @Override
+  protected String getCommandName() {
+    return "show";
+  }
+
+  @Parameter(names = "--provider", description = "Select the provider's config to display", arity = 1)
+  protected String provider;
```
ttomsu: I'd move all fields to the top of the class.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/Ui.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  public void raw(String message) {
+    reset(message);
+  }
+
+  public void info(String message) {
+    if (verbose) {
+      reset(color.BOLD + color.BLUE + ". " + color.RESET + message);
```
ttomsu: I see how your approach works, and I like the simplicity of it, but I think it's unconventional in terms of using non-static, non-final, ALLCAPS field access.

I don't have another suggestion, though.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
lwander: Ah that's neat, I'll do that

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
lwander: Nope, good idea

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GlobalOptions.java
lwander: I'm hesitant to comment this library feature, since it's likely to pop up in several places around the code base and I don't want to worry about copy/pasting comments. It's easy enough to drill down into the annotation and see what it's really doing.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ConfigCommand.java
lwander: Much neater :+1:

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/Ui.java
lwander: I agree it feels weird. It's a tiny class & should be easy to refactor if we get a better idea.

ewiseblatt: At least in this case BOLD is a weight not a color.
Does RED + BLUE make sense? I'm assuming not.
RESET doesnt sound like a Color at all.

If that is the case, then Color and Weight are specializations of some abstract class. I'm calling it Flavor because I cant think of anything here.

Now you have a type "PrintablePhrase" which is a {Flavor, String}
Your output message is a sequence of PrintablePhrase to go back and forth.
To get BOLD + BLUE you have an operator on Flavor that is Flavor = Color + Weight.

This is more API overhead than you have now, but I think clarifies what's happening and offers more opportunities to compose messages since you can have helper functions that produce PrintablePhrases and chain them together. You could even have an implicit reset across phrases and issue a Reset if two adjacent phrases have different Flavors. That probably simplifies use as well by decoupling the phrases and encouraging more composition.

And with all the above, you have a NONE color indicating you are not specifying one. Probably a BLACK too. I'm not sure if NONE means the shell default or continue that which came before. You might need some kind of ANY to indicate leave as (dont reset when appending phrases) as opposed to resetting back to the shell default.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/Ui.java
```diff
@@ -0,0 +1,62 @@
...
+  public Ui(boolean colorEnabled, boolean verbose) {
+    this.color = new Color(colorEnabled);
+    this.verbose = verbose;
+  }
+
+  private void reset(String message) {
+    try {
+      System.out.print(message);
+    } catch (Exception e) {
+      throw e;
```
ewiseblatt: What is the point of catch and throw here? Why not just try/finally?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/Ui.java
```diff
@@ -0,0 +1,62 @@
...
+    reset(message);
+  }
+
+  public void info(String message) {
+    if (verbose) {
+      reset(color.BOLD + color.BLUE + ". " + color.RESET + message);
+    }
+  }
+
+  public void warning(String message) {
```
ewiseblatt: This becomes output(
    [new PrintablePhrase(Flavor.combine(color.BOLD, color.YELLOW), "-"),
     new PrintablePhrase(Color.NONE, message)]
)
This is ugly since java doesnt have operator overloading (which I dont normally like anyway).  Since the first phrase is constant, you can precompute that into a constant PrintablePhrase.

Here it looks tedious, but it means that you could have multicolored text within the message itself since the output takes a printable phrase.

reset() becomes something like outputAtLevel(Level, Iterable<PrintablePhrase>)
warning, etc are these types of shorthand as you have here, but you could have a warning(PrintablePhrase) which would permit multicolored text which could be useful for more complicated output scenarios.

 The Level could be defined like your colors
WARNING = new PrintablePhrase(Flavor.combine(Weight.BOLD, Color.YELLOW), "- ")

Given a constructor PrintablePhrase(String message) : PrintablePhrase(Flavor.NONE, message)

Then warning(message) becomes
outputAtLevel(WARNING, PrintablePhrase(message))

outputAtLevel(Level level, PrintablePhrase phrase)  becomes
output([level, phrase])
where output adds the RESET when sequencing across phrases.

Still tedious at the lowest levels but looks concise and expressive for the use sites.

lwander: Yeah this expresses what I'm trying to do much nicer - I'll rewrite the UI stuff soon

ewiseblatt: Last remark. This is specific for CLI (ok since it is in the CLI package). For the above suggestion you want an API where the "flavor" is injected based on the output media you are rendering for (i.e. BOLD, YELLOW are abstract interface, not static constants), so a Web UI would define the color and weight differently to give you the right CSS. Then the call sites could have multi-color text that renders properly in browser or shell. Maybe there are existing ways to do this.

PR #25, opened 2016-11-10
Contributing doc defined

lwander:
First pass at the contributing doc.

@duftler @ttomsu @jtk54 @ewiseblatt @skim1420 PTAL (this is not high-priority until I get closer to completing kubernetes & gce validators).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,61 @@
+# Contributing
+
+Pull requests are welcome!
+
+## Getting Started
+
+First [fork this repo](https://help.github.com/articles/fork-a-repo/), and
+then clone it & set the upstream repo:
+
+```bash
+git clone git@github.com:your-github-username/halyard.git
+
+   # In my case (username is lwander), the above command looks like:
+   # git clone git@github.com:lwander/halyard.git
+
+cd halyard
+
+git remote add upstream git@github.com:spinnaker/halyard.git
+
+  # The above command allows you to push patches to the upstream repo for 
+  # approval.
+```
+
+## Configuring IntelliJ
+
+In the project's root directory run:
+
+```bash
+./gradlew idea
+```
+
+and then point IntelliJ at the generated project. You shouldn't need to import
+this project, simply navigating to __File__ > __Open__ > _the project's root
+directory_ should be enough.
+
+## Running the Daemon
+
+In the project's root directory run:
+
+```bash
+./gradlew
+```
+
+and the daemon will build & run. The first time you do this it will take a few
+minutes as it fetches the dependencies.
+
+## Running the Command-Line Interface (CLI)
+
+In the project's root directory run:
+
+```bash
+cd halyard-cli/
+
+make
+
+./hal --help
+```
+
+> WARNING! Since this project uses one gradle project for two executables, you
+> currently can't run `make` and `./gradlew` at the same time (one of the two
+> will likely hang indefinitely).
--- a/docs/project-structure.md
+++ b/docs/project-structure.md
@@ -1,0 +1,45 @@
+# Project structure
+
+This is a brief overview of how the project is structured.
+
+## halyard-cli/
+
+All CLI related code is here. 
+
+ * `command.v1` defines the command structure and entrypoints into command 
+   processing.  For example, the behavior of `hal config` is defined in the
+   `com.netflix.spinnaker.halyard.cli.command.v1.ConfigCommand.java` class.
+ 
+ * `ui.v1` defines UI output behavior & formatting.
+
+## halyard-config/
+
+The config validation & structure is defined here.
+
+ * `config.v1` defines utilities for loading the halconfig.
+
+ * `errors.v1` defines error & exception classes to be thrown by the
+   validation.
+
+ * `model.v1` defines the structure of the halconfig. The `providers`
+   sub-package models the different 
+   [clouddriver](https://github.com/spinnaker/clouddriver) provider's config 
+   sections, and is important if you want to add support for a specific
+   provider's configuration options
+ 
+ * `services.v1` defines classes for retrieving and updating different parts 
+   of your halconfig. These are ultimately consumed by the web controllers.
+
+ * `validate.v1` defines the validation behavior for our config. The
+   classes described in the `providers` sub-package are used in annotations for
+   specific fields/classes to allow only the necessary validators to run when
+   required.
+
+## halyard-web/
+
+The web controllers & endpoints are defined here.
+
+ * `controllers.v1` defines the set of endpoints the daemon is reachable from.
+
+ * `errors.v1` catches validation exceptions and sends them to the client as
+   JSON.


CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+```bash
+git clone git@github.com:your-github-username/halyard.git
+
+   # in my case (username is lwander), the above command looks like:
+   # git clone git@github.com:lwander/halyard.git
+
+cd halyard
+
+git remote add-url upstream git@github.com:spinnaker/halyard.git
```
duftler: Should `add-url` be `add`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Getting started
+
+First [fork this repo](https://help.github.com/articles/fork-a-repo/), and
+then clone it & set the upstream repo:
+
+```bash
+git clone git@github.com:your-github-username/halyard.git
+
+   # in my case (username is lwander), the above command looks like:
```
duftler: Could start this comment with a capital letter.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+   # git clone git@github.com:lwander/halyard.git
+
+cd halyard
+
+git remote add-url upstream git@github.com:spinnaker/halyard.git
+
+  # the above command allows you to push patches to the upstream repo for
```
duftler: Same here, and could end with a period.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
+# Contributing
+
+Pull requests are welcome!
+
+## Getting started
+
+First [fork this repo](https://help.github.com/articles/fork-a-repo/), and
+then clone it & set the upstream repo:
```
duftler: You terminate this instruction with a ":", but you don't have this with the later instructions. I prefer the ":", but whichever you choose, we should be consistent.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+cd halyard-cli/
+
+make
+
+./hal --help
+```
+
+> WARNING! Since this project uses one gradle project for two executables, you
+> currently can't run `make` and `./gradlew` at the same time (one of the two
+> will likley hang indefinitely).
```
duftler: s/likley/likely

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+This is a brief overview of how the project is structured.
+
+### halyard-cli/
+
+All CLI related code is here. 
+
+ * `command.v1` defines the command structure and entrypoints into command 
+ processing.  For example, the behavior of `hal config` is defined in the
+ `ConfigCommand.java` class.
```
duftler: I think a fully-qualified class name would be better here than `ConfigCommand.java`. `ConfigCommand.java` is a file not a class. So something like "... is defined in the `com.netflix.spinnaker.halyard.cli.command.v1.ConfigCommand` class.".

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+   specific fields/classes to allow only the necessary validators to run when
+   required.
+
+### halyard-web/
+
+The web controllers & endpoints are defined here.
+
+ * `controllers.v1` defines the set of endpoints the daemon is reachable from.
+
+ * `errors.v1` catches validation exceptions and forwards them to the client as
```
ewiseblatt: what is a client? The CLI (talks to the web server, indicating this is probably embedded in the daemon) and/or a browser? If it is a browser then "forwarding" sounds odd as compared to "responds". But why isnt the controller doing that? This feels like a "goto".

Ok, I cheated and looked at the implementation. I guess exception handlers makes sense. They look like they are spring magic stuff so have no idea how control works but presumably the controller is implicitly delegating to these (via an unarticulated assumption of someone else introducing them to spring, which is one of my naive annoyances with spring).

If the above is right, I'd be more explicit and state that these are exception handlers for the different types of exceptions that can be thrown within a controller flow, and handle the exception by terminating the flow and responding to the request as appropriate.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+cd halyard
+
+git remote add-url upstream git@github.com:spinnaker/halyard.git
+
+  # the above command allows you to push patches to the upstream repo for 
+  # approval
+```
+
+## Configuring IntelliJ
```
lwander: I read this more as
1. How do I get the code?
2. How to read open it in my editor?
3. How do I run it?
4. Where do I make changes?

which is what I'd like to get out of a contributing guide.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+In the project's root directory run
+
+```bash
+./gradlew
+```
+
+and the daemon will build & run. The first time you do this it will take a few
+minutes as it fetches the dependencies.
+
+## Running the CLI
```
lwander: Good idea, that cli/daemon and gui/daemon interface isn't well documented outside of the rest endpoints

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+ * `validate.v1` defines the validation behavior for our config. The
+   classes described in the `providers` sub-package are used in annotations for
+   specific fields/classes to allow only the necessary validators to run when
+   required.
+
+### halyard-web/
+
+The web controllers & endpoints are defined here.
+
+ * `controllers.v1` defines the set of endpoints the daemon is reachable from.
```
lwander: "endpoints the daemon is reachable from" means the daemon is defining its endpoints. I'm not sure where you are getting the client bit from, but I'm happy to reword it.

I appreciate your feedback especially since you haven't seen how the rest of the code is structured.

CONTRIBUTING.md
ewiseblatt: This is for your benefit mostly so you can do as you please.

FWIW github seems to support my claim in their "standard" recommendation
https://help.github.com/articles/setting-guidelines-for-repository-contributors/

CONTRIBUTING.md
lwander: Yup you're right, but this is too verbose: 
"these are exception handlers for the different types of exceptions that can be thrown within a controller flow, and handle the exception by terminating the flow and responding to the request as appropriate"

I also agree that the weird arbitrary spring magic bugs me too.

I guess what's missing in my doc is a clear view of what constitutes the client, and what constitutes the server, because I think with the knowledge that this code is running on the server the idea that "we catch validation exceptions, and send them to the client" is clear enough for someone to open up the `validate.v1` package and understand what's going on.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+ * `model.v1` defines the structure of the halconfig. The `providers`
+   sub-package is important if you want to add support for a specific
+   provider's configuration options
+ 
+ * `services.v1` defines services for retrieving different parts of your
+   halconfig. These are ultimately consumed by the web controllers.
+
+ * `validate.v1` defines the validation behavior for our config. The
+   classes described in the `providers` sub-package are used in annotations for
```
lwander: Yup you're right in your interpretation, but Halyard is only one service. Those different `services` above weren't explained well, and just encapsulate logic for reading/writing the halconfig .

CONTRIBUTING.md
ewiseblatt: I think client and server are intuitive. What threw me is "forwarding" because that suggests a peer-to-peer or pub/sub type of communication rather than a request/response type that I assume you have.

CONTRIBUTING.md
lwander: Ok agreed, I'll move the last section on code structure into a separate doc

CONTRIBUTING.md
lwander: That's fair, I'll change forwarding in favor of something like "responds to the client" or something

PR #30, opened 2016-11-11
(cli) show command

lwander:
@ewiseblatt @ttomsu @jtk54 PTAL

I haven't refactored the UI class yet, but that's incoming. 

Also I tried to record a gif of what the flow look like, but the pacing is strange since it seems to allot 1 frame per character drawn or so:
tty

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
@@ -18,44 +18,77 @@ package com.netflix.spinnaker.halyard.cli.command.v1;
 
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
+import com.netflix.spinnaker.halyard.cli.ui.v1.Ui;
+import com.netflix.spinnaker.halyard.config.errors.v1.HalconfigFixableIssue;
+import lombok.AccessLevel;
+import lombok.Getter;
 import lombok.Setter;
+import retrofit.RetrofitError;
 
+import java.net.ConnectException;
 import java.util.Map;
 
 abstract class NestableCommand {
   @Setter
-  protected JCommander commander;
-  private GlobalOptions globalOptions;
+  @Getter(AccessLevel.PROTECTED)
+  private JCommander commander;
+
+  @Getter(AccessLevel.PROTECTED)
+  private Ui ui;
 
   @Parameter(names = { "-h", "--help" }, help = true)
   private boolean help;
 
-  public NestableCommand(GlobalOptions globalOptions) {
-    this.globalOptions = globalOptions;
+  NestableCommand() {
+    boolean color = GlobalOptions.getGlobalOptions().isColor();
+    ui = new Ui(color);
   }
 
   /**
-   * This recusively walks the chain of subcommands, until it finds the last in the chain, and runs executeThis.
+   * This recursively walks the chain of subcommands, until it finds the last in the chain, and runs executeThis.
    *
    * @see NestableCommand#executeThis()
    */
   public void execute() {
     String subCommand = commander.getParsedCommand();
     if (subCommand == null) {
-      executeThis();
+      safeExecuteThis();
     } else {
       getSubcommands().get(subCommand).execute();
     }
   }
 
-  abstract protected void executeThis();
+  /**
+   * Used to consistently format exceptions thrown by connecting to the halyard daemon.
+   */
+  private void safeExecuteThis() {
+    try {
+      executeThis();
+    } catch (RetrofitError e) {
+      if (e.getCause() instanceof ConnectException) {
+        ui.failure(e.getCause().getMessage());
+        ui.remediation("Is your daemon running?");
+      } else {
+        HalconfigFixableIssue issue = (HalconfigFixableIssue) e.getBodyAs(HalconfigFixableIssue.class);
+        for (String warning : issue.getWarnings()) {
+          ui.warning(warning);
+        }
+
+        for (String error : issue.getErrors()) {
+          ui.failure(error);
+        }
+      }
+    }
+  }
+
   abstract protected Map<String, NestableCommand> getSubcommands();
   abstract protected String getCommandName();
+  abstract protected void executeThis();
 
   /**
    * Register all subcommands with this class's commander, and then recursively set the subcommands.
    */
-  protected void configureSubcommands() {
+  void configureSubcommands() {
     for (NestableCommand subCommand: getSubcommands().values()) {
       commander.addCommand(subCommand.getCommandName(), subCommand);
       // We need to provide the subcommand with its own commander before recursively populating its subcommands, since
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
@@ -18,19 +18,24 @@ package com.netflix.spinnaker.halyard.cli.command.v1;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.config.model.v1.Halconfig;
 import lombok.AccessLevel;
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
 
 import java.util.HashMap;
 import java.util.Map;
+
+import static com.netflix.spinnaker.halyard.cli.services.v1.Daemon.getService;
 
 /**
  * This command displays its parent's resource.
  *
  * Usage is `$ hal config show`
  */
+@Slf4j
 @Parameters(commandDescription = "Show the resource in question", separators = "=")
-public class ShowCommand extends NestableCommand {
+class ShowCommand extends NestableCommand {
   @Getter(AccessLevel.PROTECTED)
   private Map<String, NestableCommand> subcommands = new HashMap<>();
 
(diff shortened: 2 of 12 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
```diff
@@ -18,8 +18,11 @@
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.cli.execute.v1.ExecuteShow;
```
ewiseblatt: I'm not sure I understand the modularization here.
You have a command in one class and executor in another. Not exactly sure the distinction, but that's ok. However they are in different packages. Why is the primary organization "horizontal" rather than "vertical"? I'd expect better encapsulation keeping the knowledge of execute localized together. Even with ExecuteShow being an inner class. I dont appreciate any implementation reasons why they might be in different packages.

The implementation is even cross-referencing packages. I'm not sure if that is common in Java or not, but I acyclic dependencies typically raise concerns for me.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/execute/v1/ExecuteShow.java
```diff
@@ -0,0 +1,35 @@
...
+package com.netflix.spinnaker.halyard.cli.execute.v1;
+
+import com.netflix.spinnaker.halyard.cli.command.v1.ShowCommand;
+import com.netflix.spinnaker.halyard.config.model.v1.Halconfig;
+
+public class ExecuteShow extends Executor<ShowCommand> {
+  public ExecuteShow(ShowCommand command) {
+    super(command);
+  }
+
```
ewiseblatt: This seems complicated structurally.
You have a command object that is acting as an Executor factory. The instances it creates are templated back into the command class and all its really adding here is a method that essentially delegates to a lambda expression.

Why not just have it provide the lambda instead of the executor and cut all the intermediary boilerplate classes and code?

Given the diversity of halyard, I wont be surprised if in fact you do need different types of executors. But then at least perhaps have a LambdaExecutor where ShowCommand would instantiate one of those (no need for subclassing) and give it the lambda.

If you take the encapsulation approach I mentioned earlier, then nobody else is using the executor directly anyway so doesnt know the difference. The individual commands become simpler and more concise to write and maintain. And if you do need to change the executor, you can still do that with a local change and consumers are unaffected (plus, you can share those other executors in a similar way if other commands have the same refined needs)

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
```diff
@@ -0,0 +1,52 @@
...
+import retrofit.client.OkClient;
+
+import java.net.ConnectException;
+
+public class Daemon {
+  public static DaemonService getService() {
+    return getService(false);
+  }
+
+  public static DaemonService getService(boolean log) {
```
ewiseblatt: I do this kind of thing too, but want to point out that log only matters if it needs to create the service. Since you also have a getService() without log, it looks suspicious. This is a singleton and log is global, so perhaps you should set the log preference as global state and get rid of this method. That way the individual call sites dont need to replicate responsibility over figuring out whether they should log or not and a new site wont introduce a bug if it is using different rules.

Have one place determine the logging policy and set it here. It isnt the consumer's problem to worry about unless you have a logging and non-logging variant and they get what they ask for.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/Ui.java
```diff
@@ -56,6 +52,10 @@ public void failure(String message) {
     reset(color.BOLD + color.RED + "! " + color.RESET + message);
   }
 
+  public void remediation(String message) {
```
ewiseblatt: Just a thought. If you have a non-interactive mode where you inspect a configuration (e.g. post deploy problems when perhaps asynchronous changes were made that invalidated the configuration that was once good) then it would be useful to accumulate these so you could summarize all the recommendations at the end. That could be true of problems and other concerns as well. Usually that's done with local state but having these specialized reporting methods suggests there might be a way to bake that into the model [later].

lwander: You read my mind - while writing this I realized I should change up my error/warning reporting to also supply possible remediations wherever possible.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
lwander: You're right, I've been kind of cheating myself by calling the `GlobalOptions` global, but then always passing them around. I'll fix this up.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShowCommand.java
lwander: My goal was to separate the structure of the commands (i.e. the `update` and `show` commands follows `config`) from how they are executed, because to me those are logically separate. You're right I didn't do a good job of keeping these structurally simple, let me revisit this Monday and show you a new implementation.

ewiseblatt: It's more than structural simplicity. It's encapsulating the appropriate details.

Giving the classes limited roles is a reasonable thing to do. And having an internal model that defines those classes and can provides an interpreter that mediates among them symbolically so that control flow is standardized and you're just plugging in specialized classes to implement the different activities is a common pattern. What's missing here is that usually that pattern treats those classes as a clique defined by what's being operated on, not the type of operation.

Rather than Executions {Insert, Delete }, Command {Insert, Delete}
You want  Insert {Execution, Command}  Delete {Execution, Command}

That gives encapsulation around the resource rather than the activity.

It is far more likely that you'd want cohesion around the different activities on a resource than around different resources for a given activity. The base classes are already defining the common parts of the activity anyway. Now you can add new resources without impacting other ones. Just drop a new module in. Otherwise you need to touch and modify all the other existing modules, not to say you've lose the encapsulation about the resource so are spreading that knowledge across all your modules as well.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/Halconfig.java
```diff
@@ -42,4 +42,28 @@
    * List of available deployments.
    */
   private List<DeploymentConfiguration> deploymentConfigurations = new ArrayList<>();
+
+  @Override
+  public String toString() {
+    StringBuilder result = new StringBuilder();
```
ttomsu: Stylistically, these could all be linked:

```
result.append("version: ")
      .append(halyardVersion)
      .append("\ncurrentDeployment: ")
      .append(currentDeployment)
...
```

lwander:
Did a refactor, removing the `Executor`. PTAL @ewiseblatt

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -18,44 +18,77 @@
...
+      if (e.getCause() instanceof ConnectException) {
+        ui.failure(e.getCause().getMessage());
+        ui.remediation("Is your daemon running?");
+      } else {
+        HalconfigFixableIssue issue = (HalconfigFixableIssue) e.getBodyAs(HalconfigFixableIssue.class);
+        for (String error : issue.getErrors()) {
+          ui.failure(error);
+        }
+
+        for (String warning : issue.getWarnings()) {
```
ewiseblatt: FWIW, I like errors after warnings (increasing severity). That way when you finish scrolling and just look at the bottom, you see the worst case. I guess it could also keep the suspense reading linearly.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
```diff
@@ -0,0 +1,50 @@
...
+  // TODO(lwander): setup config file for this
+  static final private String endpoint = "http://localhost:8064";
+
+  static private DaemonService service;
+
+  private static DaemonService createService(boolean log) {
+    return new RestAdapter.Builder()
+        .setEndpoint(endpoint)
+        .setClient(new OkClient())
+        .setLogLevel(log ? RestAdapter.LogLevel.FULL : RestAdapter.LogLevel.NONE)
```
ewiseblatt: Why the extremes? Wouldnt you normally want something more "normal"?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/Halconfig.java
```diff
@@ -42,4 +42,28 @@
...
+        .append('\n');
+
+    result.append("currentDeployment: ")
+        .append(currentDeployment)
+        .append('\n');
+
+    result.append("deployments: ")
+        .append('\n');
+
+    for (DeploymentConfiguration deployment : deploymentConfigurations) {
```
ewiseblatt: Explicitly indicate when there are none.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
lwander: I figure this will only be used if a customer complains that their halyard is broken, and I need to see what the daemon & cli are saying to one another in full. This only logs the HTTP traffic between the two, which really isn't useful to know unless you need to see exactly what's going wrong.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
lwander: Heh, I'll flip it around.

lwander:
One more set of updates completed @ewiseblatt

ewiseblatt:
LGTM

PR #31, opened 2016-11-14
(cli) refactor UI

lwander:
@ewiseblatt @ttomsu PTAL


lwander:
Build is failing because java home is set incorrectly :confused:... looking now

ttomsu:
LGTM

lwander:
@ewiseblatt I added the AnsiMessageBuilder (although a little differently from what you recommended). I prefer to keep the printing logic outside of the messages because it feels strange and locks us into always printing to stdout, when instead we could write many printers conforming to our AnsiPrinterInterface if we just pass the messages to them.

ewiseblatt:
The "Style" change over Weight was still an improvement. I'd keep that.

lwander:
Agreed

lwander:
Good to go?

ewiseblatt:
sure

PR #32, opened 2016-11-16
Flushed out provider & account endpoints

lwander:
@duftler or @ttomsu or @jtk54 or @ewiseblatt PTAL

Don't pay too much attention to the `NamingService`, halfway through I realized it would be better to add fields to the issues & warnings describing exactly which account/provider/field is affected, but that would bleed to far into my refactor of the `FixableIssues`.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ProviderService.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ProviderService.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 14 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ProviderService.java
```diff
@@ -0,0 +1,64 @@
...
+
+    return provider;
+  }
+
+  public Providers getProviders(String deploymentName) {
+    DeploymentConfiguration deploymentConfiguration = deploymentService.getDeploymentConfiguration(deploymentName);
+
+    Providers providers = deploymentConfiguration.getProviders();
+    if (providers == null) {
+      throw new NoProviderException("The " + NamingService.deployment(deploymentName) + "has no providers configured");
```
duftler: Missing a space just before 'has'.

Issue #27, opened 2016-11-10
Validate Kubernetes Provider


lwander:

lwander:
Closing with #33

PR #33, opened 2016-11-17
Refactor of error handling

lwander:
In the case of an error, I switched from returning two lists of strings
representing errors and warnings, to a single list of "Problems" (I
guess the name could be something different, but I don't think it's worth
debating too intensely). Each "problem" is the combination of an error severity
(ranging from the unfixable, to a simple suggestion), a message, an optional
remediation, and an optional location/coordinate to the place in the config
where this error was encountered. Right now it's mostly validating simple,
generic properties of the config/request (like was this requested account
defined exactly once), but once I plug in the rest of my validators it will
return the full list of problems they encounter when validating your config and
any changes you've made to it.

@ewiseblatt @duftler @ttomsu @jtk54 PTAL


lwander:
I've added on a second commit that significantly simplifies the validators. I'm dropping the annotation based approach in favor of simply having each class define how to validate itself.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -62,16 +64,44 @@ private void safeExecuteThis() {
...
+          String message = problem.getMessage();
+          String remediation = problem.getRemediation();
+
+          if (coordinates != null) {
+            problemLocation = "In " + coordinates + ":";
+          } else {
+            problemLocation = "Global:";
+          }
+
+          if (severity == HalconfigProblem.Severity.FATAL) {
```
jtk54: Use a `switch` on `severity`.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/AnsiUi.java
```diff
@@ -49,7 +49,7 @@ public static void failure(String message) {
   }
 
   public static void remediation(String message) {
-    AnsiMessage prefix = new AnsiMessage("? ")
+    AnsiMessage prefix = new AnsiMessage("\t? ")
```
jtk54: Why `\t`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * This serves to point to a location in your Halconfig. This is intended to help users of hal to troubleshoot where
+ * a certain error is coming from.
+ */
+@Data
+public class HalconfigCoordinates {
+  String deployment;
+  String provider;
+  String webhook;
```
jtk54: Which webhook is this?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
```diff
@@ -0,0 +1,62 @@
...
+          .append(".");
+    }
+
+    if (account != null) {
+      res.append(account)
+          .append(".");
+    }
+
+    String output = res.toString();
+    // cut of trailing period
```
jtk54: s/of/off/

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
```diff
@@ -0,0 +1,62 @@
...
+@Data
+public class HalconfigCoordinates {
+  String deployment;
+  String provider;
+  String webhook;
+  String account;
+
+  @Override
+  public String toString() {
+    StringBuilder res = new StringBuilder();
```
jtk54: Why not build a list of these ("deployment.${deployment}", etc) and then do `list.join(".")` to handle the period interspersing. This way seems fragile.

jtk54: Also, is this just for rendering or will this actually be consumed by something? If the latter, what's wrong with json?

lwander:
I added one more commit to take the kubernetes validation logic I tore out and reinstate it (with some extra checks) using the refactored validators, as well as expose this functionality in an endpoint.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/errors/v1/config/IllegalConfigException.java
```diff
@@ -16,15 +16,30 @@
...
+import com.netflix.spinnaker.halyard.config.errors.v1.HalconfigProblem;
+import com.netflix.spinnaker.halyard.config.errors.v1.HalconfigProblemSet;
+
+import java.util.List;
 
 /**
  * This is reserved for Halyard configs that fall between unparseable (not valid yaml), and incorrectly configured
  * (provider-specific error). Essentially, when a config has problems that prevent halyard from validating it, although
- * it is readable by our yaml parser into the halfconfig Object, this is thrown
+ * it is readable by our yaml parser into the hafconfig Object, this is thrown
```
jtk54: s/hafconfig/halconfig

lwander:
Closes https://github.com/spinnaker/halyard/issues/27

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -47,4 +55,112 @@ public String getKubeconfigFile() {
       return kubeconfigFile;
     }
   }
+
+  public List<HalconfigProblem> validateKubeconfig(DeploymentConfiguration deployment, HalconfigCoordinates coordinates) {
```
jtk54: Does this need to be public?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -47,4 +55,112 @@ public String getKubeconfigFile() {
...
+        result.add(new HalconfigProblem(HalconfigProblem.Severity.ERROR, coordinates,
+            "Unable to communicate with your Kubernetes cluster: " + e.getMessage(),
+            "Verify that these credentials work manually using \"kubectl\""));
+      }
+    }
+
+    return result;
+  }
+
+  public List<HalconfigProblem> validateDockerRegistries(DeploymentConfiguration deployment, HalconfigCoordinates coordinates) {
```
jtk54: Same here with public.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
lwander: You can't do that in java world

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/AnsiUi.java
lwander: To make clear that this recommendation is for the above error/warning

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
lwander: Any webhook

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
lwander: It's for rendering only, I can switch to a list.join, but don't really see the need for 4 elements

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/errors/v1/config/IllegalConfigException.java
lwander: ha, I miscorrected my typo

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
lwander: No, good catch

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
lwander: Yes. changes to the docker registries outside this account mean the kubernetes account registries need to be validated without triggering live calls to the kubernetes cluster.

jtk54: SGTM

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
jtk54: Wat: http://stackoverflow.com/questions/6391777/switch-on-enum-in-java

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
jtk54: If you ever add anything though...

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
lwander: Ah nice, intellij was complaining about non-constant values in a case statement which from a compiler standpoint didn't make sense (enum values should be determined statically). Let me switch this around

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigCoordinates.java
jtk54: Ok

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/Validatable.java
```diff
@@ -0,0 +1,33 @@
...
+import com.netflix.spinnaker.halyard.config.errors.v1.HalconfigProblem;
+
+import java.util.List;
+
+/**
+ * Anything implementing this interface is validatable, meaning after we validate it, we get a non-null list of HalconfigProblems
+ *
+ * Warning: If any of the fields in a class implement Validatable, do not call validate on them!! This will mean even the simplest
+ * config changes could cause the full halconfig for every deployment to be validated (very costly, many network calls). It is up
+ * to whatever is performing a field update to decide what to validate.
```
lwander: Ah, I like the idea of the problem constructor, I'll add that in. 

As far as restricting calling "validate" recursively, I'll look into how I can make that happen.

ewiseblatt: If the caller into the factory identifies themself ("this") when creating a problem then you could detect that someone you didnt call had called back into you. Or if you do want some kind of delegation, you can detect repeated problems. It could be helpful to have a call into the factory that confirms no problems are found as a means to audit coverage over all the configuration. That call too would act as a control point to detect duplicate calls if in fact that is something you want to protect against or debug.

ewiseblatt: I'm only mentioning the detection thing since you seem concerned about that situation. I do think the auditability of things that are ok to ensure coverage would be desirable regardless.

lwander:
@ewiseblatt I followed your advice and created a `HalconfigProblemSetBuilder`, which is passed into the validators and modified by them as they see fit. I don't see a nice way preventing Validators from recursively calling themselves just yet, so I think I'll save that for a later PR. Any objections to merging?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/HalconfigProblemSetBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.AccessLevel;
+import lombok.Setter;
+
+public class HalconfigProblemSetBuilder {
+  private HalconfigProblemSet problemSet;
+
+  @Setter(AccessLevel.PUBLIC)
+  private HalconfigCoordinates coordinates;
+
+  public void addProblem(Severity severity, String message, String remediation) {
```
lwander: I'll switch it over - but I'm not sure what this buys me

lwander:
Addressed

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ValidateService.java
```diff
@@ -0,0 +1,39 @@
...
+@Component
+public class ValidateService {
+  @Autowired
+  DeploymentService deploymentService;
+
+  public void validate(Validatable validatable, HalconfigCoordinates coordinates) {
+    DeploymentConfiguration deployment = deploymentService.getDeploymentConfiguration(coordinates);
+    HalconfigProblemSetBuilder builder = new HalconfigProblemSetBuilder().setCoordinates(coordinates);
+
+    validatable.validate(builder, deployment);
```
ewiseblatt: This is awkward because one validate takes a builder and does not throw where this other one takes a validatable and throws exceptions.
I think you want different terminology at the minimum.


I'm not sure about this builder passed in as being a builder as opposed to a builder factory. When you set the coordinates on it above, are you expecting that the caller did not change them? Or are these mutated as the caller propagates down the tree?

Should the deployment be an attribute of the builder?
It isnt clear from the service how this is used. It seems terminal but I'd expect it to flow down in some kind of traversal. Maybe your comment earlier about calling validate was saying not to do that. I still dont understand how this is used for composite configs unless the parts take the identity of their container.

ewiseblatt:
I cannot find the "not sure what it buys me" comment to respond to it.
If you dont know, then dont listen. Maybe I'm wrong or I can explain to
clarify.
From our side converstation it sounds like I have different expectations of
how this is going to be consumed.

lwander:
For posterity about the above comment: I was hoping you could explain the motivation/benefit of your proposed changes, which you described offline + some other ideas on how to make this easier to extend. I'll work on implementing those this weekend/next week.

lwander:
@ewiseblatt, I pulled the validation into a vistor that each yaml node accepts. Nodes to visit are decided by an iterator that returns nodes yaml nodes matching a given filter. The core validation logic is now very small: https://github.com/spinnaker/halyard/pull/33/commits/b5822ccc65af3b9d8c71836ae9c4746f85d35778#diff-af46e1bd0bad8e97c8e74a98cb583073R35

As far what new nodes will have to implement goes: It'll be a `validateClassName(ClassName c)` for each node added, as well as some information about the type of the node for the sake of filtering. 

This isn't done yet (still need to to plug the last bits in), and I'll have it finished tomorrow.

lwander:
Ok, I think the refactor is done (pending feedback of course)

The idea is now that every node in the yaml tree extends `Node`, which provides a few bits of utility that need to be implemented for each new type `Node` (i.e. when we add support for Fiat config, but _not_ when target/veritas add support for configuring opentstack). `Node` requires us to describe how to: get children, get the parent, accept a validator, and say if it matches a `NodeFilter` or not. Most of the logic for these things have already been implemented in either the abstract `Node` class, or in related utility classes (see `NodeIteratorFactory`), so extending this isn't much work for new types `Node`s. If instead you are adding a new account/provider/webhook/etc... you just need to extend the right superclass (`Account`/`Provider`/`Webhook`/etc...) and provide whatever fields should show up in the yaml.

The core mechanism to keep in mind in the `Node` package is the `NodeIterator`, which allows us to walk over each node in the yaml tree, or each node that matches an input `NodeFilter`. This is important during validation.

Next we have the `Validator`, which is simply a visitor that takes a `ProblemSetFactory` (to keep track of all the errors/warnings encountered during validation), and a `Node`. Inside the validate package, we have the custom validators (see `AccountValidator` for a small example), and the `ValidatorCollector`, which through a Spring annotation collects every class that implements `Validator`. Using reflection, the `ValidatorCollector` takes a node, and scans each `Validator` to see if it can validate the given `Node` or not.

Lastly, in the services package, the `ValidatorService` takes a `NodeFilter`, and then iterates over every node in the current halconfig that matches the filter, and runs the `ValidatorCollector` on it. Once complete. Incoming requests just format their `NodeFilter` to match what they way, and then have the `ValidatorService` do the rest.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
```diff
@@ -37,8 +39,13 @@
   @Autowired
   ProviderService providerService;
 
-  public Account getAccount(String deploymentName, String providerName, String accountName) {
-    Provider provider = providerService.getProvider(deploymentName, providerName);
+  @Autowired
+  ValidateService validateService;
+
+  public Account getAccount(NodeReference reference) {
+    Provider provider = providerService.getProvider(reference);
```
ewiseblatt: I'm still finding this API weird. I'd expect reference.getProvider().
I guess this ok because either way you are getting the provider. But since you only autowire the provider service to give it the reference to get the provider, it makes this that much more awkward. A reference sounds like a data object so this getProvider() wouldnt be appropriate there. But if this is a common pattern for resolving objects in the reference, then NodeReference could have a companion class that you use instead and that would autowire all the things instead. I dont know if that's a good idea or not. Just a reaction reading this.
The difference again is one of encapsulating the implementation details. The interface is the clique "NodeReference" and "NodeReferenceResolver" (bad name) rather than NodeReference and *Service.

On the other hand, at least in this case, AccountService and ProviderService are already coupled since Accounts are part of Providers. So in this case it isnt as big a deal, though still feels like something is going on here and, at least for now, you dont actually use the service for anything other than resolving the reference.

Leave it as is for now and see how it bakes out. I am just sharing thoughts and gut reaction for now.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentService.java
```diff
@@ -34,20 +38,27 @@
...
+        .filter(d -> d.getName().equals(deploymentName))
         .collect(Collectors.toList());
 
     switch (matching.size()) {
       case 0:
-        throw new IllegalConfigException("The " + NamingService.deployment(deployment) + " not found");
+        throw new IllegalRequestException(new ProblemBuilder(Problem.Severity.FATAL,
+            "No matching deployment could be found")
+            .setReference(reference)
+            .setRemediation("Create a new deployment").build());
```
ewiseblatt: This isnt just any deployment but one whose name is [deploymentName] right?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentService.java
```diff
@@ -34,20 +38,27 @@
...
       case 1:
         return matching.get(0);
       default:
-        throw new IllegalConfigException("The " + NamingService.deployment(deployment) + " appears more than once");
+        throw new IllegalConfigException(new ProblemBuilder(Problem.Severity.FATAL,
+            "More than one matching deployment found")
```
ewiseblatt: Is this multiple matching deployments of the given name? Could you clarify here where they came from to help track them down?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ProviderService.java
```diff
@@ -42,21 +48,32 @@ public Provider getProvider(String deploymentName, String providerName) {
       provider = providers.getDockerRegistry();
     } else if (providerName.toLowerCase().equals("google")) {
       provider = providers.getGoogle();
+    } else {
```
ewiseblatt: I dont like this enumeration here. I let it go in node/Providers.java. Could you pick one or the other so there is only one place to modify when adding a new provider?

For example here you could
provider = providers.find(providerName)
And add the above into the node/Providers. Then this class doesnt care who the providers are and doesnt need to be modified when adding a new provider.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/ValidatorCollection.java
```diff
@@ -0,0 +1,91 @@
...
+      Method m = validator.getClass().getMethod("validate", ProblemSetBuilder.class, c);
+      m.invoke(validator, psBuilder, node);
+      result = 1;
+    } catch (NoSuchMethodException e) {
+      // Do nothing, odds are most validators don't validate every class.
+    } catch (InvocationTargetException | IllegalAccessException e) {
+      log.warn("Failed to invoke validate() on " + validator.getClass() + " for node " + c, e);
+    }
+
+    return runMatchingValidators(psBuilder, validator, node, c.getSuperclass()) + result;
```
ewiseblatt: This feels weird because you are calling validate on each of the classes in the node's hieararchy. Why wouldnt the specialized validate already accomodate this?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
```diff
@@ -0,0 +1,156 @@
...
+
+    // This indicates if a first pass at the config looks OK. If we don't see any serious problems, we'll do one last check
+    // against the requested kubernetes cluster to ensure that we can run spinnaker.
+    boolean smoketest = true;
+
+    try {
+      File kubeconfigFileOpen = new File(kubeconfigFile);
+      kubeconfig = KubeConfigUtils.parseConfig(kubeconfigFileOpen);
+    } catch (IOException e) {
+      psBuilder.addProblem(ERROR, e.getMessage());
```
ewiseblatt: If the problem were there is no kubeconfig file, would you want to instruct how to create one (or point to instructions)?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
```diff
@@ -0,0 +1,156 @@
...
+
+    if (smoketest) {
+      Config config = KubernetesConfigParser.parse(kubeconfigFile, context, cluster, user, namespaces);
+      KubernetesClient client = new DefaultKubernetesClient(config);
+
+      try {
+        client.namespaces().list();
+      } catch (Exception e) {
+        psBuilder.addProblem(ERROR, "Unable to communicate with your Kubernetes cluster: " + e.getMessage())
+            .setRemediation("Verify that these credentials work manually using \"kubectl\"");
```
ewiseblatt: "these credentials" which are "these" or is that clear from the reference reported with the remediation?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
```diff
@@ -0,0 +1,156 @@
...
+
+    if (dockerRegistries == null || dockerRegistries.isEmpty()) {
+      psBuilder.addProblem(ERROR, "You have not specified any docker registries to deploy to")
+          .setRemediation("Add a docker registry that can be found in this deployment's dockerRegistries provider");
+    }
+
+    DockerRegistryProvider dockerRegistryProvider = deployment.getProviders().getDockerRegistry();
+    if (dockerRegistryProvider == null || dockerRegistryProvider.getAccounts() == null || dockerRegistryProvider.getAccounts().isEmpty()) {
+      psBuilder.addProblem(ERROR, "The docker registry provider has not yet been configured for this deployment")
+          .setRemediation("Kubernetes needs a Docker Registry as an image source to run");
```
ewiseblatt: Not a useful remediation. Perhaps point to the spinnaker.io documentation for adding docker registries.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -62,16 +66,45 @@ private void safeExecuteThis() {
...
+
+        problemSet.sortIncreasingSeverity();
+        for (Problem problem : problemSet.getProblems()) {
+          Severity severity = problem.getSeverity();
+          NodeReference reference = problem.getReference();
+          String problemLocation;
+          String message = problem.getMessage();
+          String remediation = problem.getRemediation();
+
+          if (reference != null) {
```
lwander: Good thinking

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @JsonIgnore
+  public abstract NodeReference getReference();
+
+  @Getter
+  @JsonIgnore
+  protected Node parent = null;
+
+  @JsonIgnore
+  public void parentify() {
```
lwander: Because the snakeyaml parser alone doesn't add parents, so I have my "special" yaml parser call this before returning your halconfig object.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (a.equals(ANY) || b.equals(ANY)) {
+      return true;
+    }
+
+    return a.equals(b);
+  }
+
+  public NodeFilter anyHalconfigFile() {
```
lwander: Good point. I'm not married to the NodeFilter setup just yet, so this may change in the near future.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeIteratorFactory.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * @see Node
+ * @see NodeIterator
+ */
+@Slf4j
+public class NodeIteratorFactory {
+  /**
+   * Transforms a Node into an iterator that allows us to iterate over all sub-fields with type node.
```
lwander: Ah right, good point. The node remains unchanged - I'll make that clearer.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeReference.java
```diff
@@ -14,23 +14,19 @@
...
 @Data
-@NoArgsConstructor
-@EqualsAndHashCode(callSuper = false)
-public class FieldReference<T> extends Reference<T> {
-
-  /**
-   * The name of the field being updated.
-   */
-  private String fieldName;
+public class NodeReference implements Cloneable {
```
lwander: Chatted offline, the idea is that since we dictate the structure this is OK for now, but worth revisiting. (The nice thing about the current abstraction is that new providers can come in and make changes without touching this model, so we are free to refactor/extend it).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Validatable.java
```diff
@@ -14,18 +14,13 @@
...
 /**
- * An annotation to validate a specific field.
+ * Anything implementing this interface is validatable, meaning after we validate it, we get a non-null list of HalconfigProblems
  */
-@Retention(RetentionPolicy.RUNTIME)
-@Target(ElementType.FIELD)
-public @interface ValidateField {
-  Class<? extends Validator<?>>[] validators();
+public interface Validatable {
+  void accept(ProblemSetBuilder psBuilder, Validator v);
```
lwander: There is only one way - this is standard in the visitor pattern because as we iterate over `Node`s, we want to to ensure that the Validator is called on the right `Node` subclass, so we have the `Node` pass `this` into the `Validator` which will be the concrete subclass we want to validate.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/problem/ProblemSet.java
```diff
@@ -0,0 +1,88 @@
...
+ * This represents all problems collected when validating the currently loaded/modified halconfig.
+ */
+public class ProblemSet {
+  @Getter
+  private List<Problem> problems = new ArrayList<>();
+
+  /**
+   * Sort the listed problems in increasing order of severity.
+   */
+  public void sortIncreasingSeverity() {
```
lwander: Mm good thinking - I'll see how useful the ./hal command output is

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/google/GoogleProvider.java
```diff
@@ -16,12 +16,22 @@
...
+import com.netflix.spinnaker.halyard.config.model.v1.node.Provider;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
 @Data
 @EqualsAndHashCode(callSuper = false)
 public class GoogleProvider extends Provider<GoogleAccount> implements Cloneable {
+  @Override
+  public String getNodeName() {
+    return "dockerRegistry";
```
lwander: whoopsie

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentService.java
lwander: Good point

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ProviderService.java
lwander: This is leaving once the `NodeFilter` is used to do the lookup here - I agree it's messy especially when I have the more powerful filters to rely on.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/ValidatorCollection.java
lwander: This way we avoid duplicating validation between subclasses - like with the given example for `Account` vs. `Kubernetes Account`. The first validator checks only the name, and the second checks all the k8s specific fields.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
lwander: Good point, I'll make it clearer

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
lwander: Actually, I think it's best to hold off on that, since ideally the registry should be added with halyard, but there is no documentation for that. I'll add a todo.

PR #34, opened 2016-11-21
Lookup provider for more generic node retrieval

lwander:
@ewiseblatt PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
@@ -34,34 +34,38 @@ public class NodeFilter extends NodeReference {
     return a.equals(b);
   }
 
-  public NodeFilter anyHalconfigFile() {
+  public NodeFilter withAnyHalconfigFile() {
     halconfigFile = ANY;
     return this;
   }
 
-  public NodeFilter anyDeployment() {
+  public NodeFilter withAnyDeployment() {
     deployment = ANY;
     return this;
   }
 
-  public NodeFilter anyWebhook() {
+  public NodeFilter withAnyWebhook() {
     webhook = ANY;
     return this;
   }
 
-  public NodeFilter anyProvider() {
+  public NodeFilter withAnyProvider() {
     provider = ANY;
     return this;
   }
 
-  public NodeFilter anyAccount() {
+  public NodeFilter withAnyAccount() {
     account = ANY;
     return this;
   }
 
-  public NodeFilter anyMaster() {
+  public NodeFilter withAnyMaster() {
     master = ANY;
     return this;
+  }
+
+  public static NodeFilter makeRejectAllFilter() {
+    return new NodeFilter();
   }
 
   public NodeFilter(NodeReference reference) {
@@ -72,4 +76,6 @@ public class NodeFilter extends NodeReference {
     this.account = reference.account;
     this.master = reference.master;
   }
+
+  private NodeFilter() { }
 }
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
@@ -17,13 +17,12 @@
 package com.netflix.spinnaker.halyard.config.services.v1;
 
 import com.netflix.spinnaker.halyard.config.errors.v1.config.IllegalConfigException;
-import com.netflix.spinnaker.halyard.config.errors.v1.config.IllegalRequestException;
+import com.netflix.spinnaker.halyard.config.errors.v1.config.ConfigNotFoundException;
 import com.netflix.spinnaker.halyard.config.model.v1.node.NodeReference;
 import com.netflix.spinnaker.halyard.config.model.v1.node.NodeFilter;
 import com.netflix.spinnaker.halyard.config.model.v1.problem.Problem;
 import com.netflix.spinnaker.halyard.config.model.v1.problem.ProblemBuilder;
 import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
-import com.netflix.spinnaker.halyard.config.model.v1.node.Provider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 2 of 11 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/LookupService.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Slf4j
+@Component
+public class LookupService {
```
lwander: I would have preferred to make this `LookupService<T extends Node>`, but java erases the runtime type information, leaving me unable to correctly cast the returned list of nodes.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
```diff
@@ -64,6 +64,8 @@ public NodeFilter anyMaster() {
     return this;
   }
 
+  public NodeFilter() { }
```
ewiseblatt: Do you mean to filter everything or did you really mean nothing?
Does this mean you need to explicitly populate everything to use [at least everything above some level]?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
```diff
@@ -37,36 +38,42 @@
...
   @Autowired
-  ProviderService providerService;
+  LookupService lookupService;
 
   @Autowired
   ValidateService validateService;
 
-  public Account getAccount(NodeReference reference) {
-    Provider provider = providerService.getProvider(reference);
+  public List<Account> getAccounts(NodeReference reference) {
```
ewiseblatt: getAllAccounts.  getAccounts is too similar to getAccount below.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
```diff
@@ -37,36 +38,42 @@
...
 
-    String accountName = reference.getAccount();
+    return lookupService.getMatchingNodesOfType(filter, Account.class)
+            .stream()
+            .map(n -> (Account) n)
+            .collect(Collectors.toList());
+  }
 
-    List<Account> accounts = provider.getAccounts();
+  public Account getAccount(NodeReference reference) {
```
ewiseblatt: Why does getAccount throw an exception (for not found) but getAccounts does not? It sounds like the caller assumes it exists here but not above.

Though I'm inconsistent, I like to believe that I would use find* to not throw an exception and get* to, at least when I have control over the API naming conventions. So above would be findAllAccounts() and here would be getAccount() or if it were findAccount() it could return null. But youd still throw if there were multiple since it is ambiguous and that isnt expected given the API is assuming a single value.

Had you not had getAccounts() above, I might not thought anything of this as is. The two makes me wonder what the assumed usage pattern is.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
```diff
@@ -37,36 +38,42 @@
...
         .stream()
-        .filter(a -> a.getName().equals(accountName))
+        .map(n -> (Account) n)
         .collect(Collectors.toList());
 
     switch (matchingAccounts.size()) {
       case 0:
         throw new IllegalRequestException(new ProblemBuilder(
-            Problem.Severity.FATAL, "No matching account found account")
+            Problem.Severity.FATAL, "No matching account with name \"" + accountName + "\" found")
```
ewiseblatt: What makes this a fatal error? Wouldnt you need more context to know that it is fatal or not?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/LookupService.java
ewiseblatt: I'll take your word for it. I'm not a Java expert to recommend other ways (e.g. not casting). You might ask someone else if you havent already.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/LookupService.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * If the filter represents a pruned tree, then this
+   * @param node is the node whos children we want to find.
+   * @param filter is the filter to lookup by.
+   * @return
+   */
+  private List<Node> getMatchingLeafNodes(Node node, NodeFilter filter) {
+    log.info("Checking for leaf nodes of node " + node.getNodeName());
```
ewiseblatt: The logging here and especially below sounds like it is going to be very verbose. Are you sure you want this @info?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
lwander: Filter nothing, I think it'll be clearer if I wrap this into a static method called `makeRejectAllFilter()` or something similar.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
lwander: I'm thinking of changing this to `getAccountNames` for now, since from an API perspective that's more useful (at least for the CLI).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
lwander: Right now, the only use-cases this endpoint enables are listing the accounts & validating them. Even when I add the ability to add/remove/edit accounts, specifiying a non-existing account for any of those operations is impossible to fulfill.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/LookupService.java
lwander: You're right, I'll drop it to `trace`

lwander:
Addressed @ewiseblatt

lwander:
I'm not 100% sure how to format some of the remediations just yet BTW, since a few will best be served by using Halyard. Once I play more with the CLI I will have a better feel for what should be provided for each error, and will write a guide at that point.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/LookupService.java
lwander: Talked with @duftler, he agreed with the type erasure and brought up a second reason why this wouldn't work - we can't autowire generic type

lwander:
I have some CLI changes incoming that will touch these endpoints, so I'm going to merge these now

PR #37, opened 2016-11-30
Parse BOM from GCS to get service config

lwander:
@jtk54 @duftler @ewiseblatt PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/spinnaker/v1/BillOfMaterials.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/spinnaker/v1/BillOfMaterials.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2016 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/spinnaker/v1/BillOfMaterials.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Optional;
+
+@Data
+public class BillOfMaterials {
+  String version;
+  SpinnakerComponents services;
+
+  @Data
+  static class SpinnakerComponents {
+    String clouddriver;
```
lwander: Described here: https://github.com/spinnaker/halyard/blob/master/docs/design.md#11-versioning-spinnaker

lwander:
@ewiseblatt : https://github.com/spinnaker/halyard/blob/master/docs/design.md#1-configuring-spinnaker

lwander:
I would treat this as a proof-of-concept, putting the pieces together we need from the release process with halyard to generate config.

Essentially, it checks a known endpoint for a bill of materials matching the version of Spinnaker to deploy. Once it has that bill of materials, it looks up whatever subcomponent's version it needs, and uses that version to get its config.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ConfigController.java
```diff
@@ -32,6 +34,9 @@
   @Autowired
   ConfigService configService;
 
+  @Autowired
+  ComponentProfile componentProfile;
```
jtk54: Remove or use : )

lwander:
I'm going to merge this for now - it's in line with out we set out to treat the BOM, but realistically only a small bit of code that's easy to replace if we come up with a different versioning policy.

PR #38, opened 2016-11-30
enable/disable google & docker

lwander:
Pretty straightforward, good example of how to add commands. @ewiseblatt @duftler PTAL


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/google/GetGoogleAccountCommand.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+
+/**
+ * Describe a specific kubernetes account
+ */
+@Parameters()
+public class GetGoogleAccountCommand extends AbstractGetAccountCommand {
+  @Getter(AccessLevel.PROTECTED)
+  private String providerName = "kubernetes";
```
ewiseblatt: "google"

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/google/GoogleCommand.java
```diff
@@ -0,0 +1,48 @@
...
+import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.cli.command.v1.NestableCommand;
+import com.netflix.spinnaker.halyard.cli.command.v1.providers.AbstractProviderCommand;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
```
ewiseblatt: I understand providers are going to be different from one another, but so far the kubernetes, docker, and google all feel very similar and all require a lot of boilerplate code. That's a double whammy.

If you could factor out common patterns such that the classes only distinguish how they are different or non-standard, it would likely cut down on a lot of redundancy and make it much clearer what [if any] is different about a given provider.

lwander: Not sure I understand, the classes (enable/disable) that have the most code have 3 fields. This top-level command class  has two lines per subcommand (one to instantiate, one to register). I could use multi-inheritance to provide each provider's commands with their provider name, but that only cuts out 1 line.

ewiseblatt: Conceptually you have a provider that you can enable disable. To do that you have a base class shared by enable and disable, then concrete classes for each. You have a factory type class that then instantiates these. That's 4 classes with a lot of code, most not interesting. Now I have 3 (eventually at least 7) providers all of which doing that. Sure the providers will be doing more than enable/disable but with so many providers I'm sure there will be much more overlap and some things surely unique.

That's a lot of code to maintain, especially that is conceptually duplicative. In the end when you have more than a handful of providers, it wont be obvious what is unique and what is common. Where there are things that are similar, they will each maintain that concept independently so only be the same by coincidence as opposed to definition or policy.

It's unfortunate that you cannot say that [all] providers can be enabled and disabled. Then this PR would just be saying, "by the way, 'dockerRegistry' and 'google' are also providers". That would be a 2 lines in 1 file, not > 200 lines in 9 files. Sure most of those lines are copyright header, import statements, and other boilerplate. But it's still there, you have to parse through it, and it isnt obvious if the provider is doing something different or left something out, etc.

I do like that you have a primitive concept (AbstractEnableDisableCommand) that you draw on and that is implementing the policy I refer to. However subclassing it takes away from that. Especially since the subclass is because of data values. That's how the annotations and libraries you are using guide you so maybe that's the hot thing these days, but it seems an unfortunate direction to me. The library is cutting down code you have to write like Microsoft's Wizards in the 90's. But that misses the forest for the trees. It's the model that is bringing in the baggage. Reducing code doesnt change the cognitive complexity or clarity of what's going on. Personally I'd rather see more lines of code but a more concise abstraction (so I dont need to read the code) than less code but I need to read it all to understand the subtlelties. I dont know the subclasses are just replacing data fields unless I read each one.

Again, it might not be doable, but this would read much more clearly if you can say "I am defining the provider 'google'. Add the ChangeGoogleProjectCommand" where ChangeGoogleProject is some command that is specific to that provider (hence Google in the name). I chose this to make a point. All providers have enable disable, so because it is a provider it will have Enable/Disable. Because I declared the provider's name, I do not need to reiterate it for each command (assuming the command still needs it). So the above still adds enable/disbale but it is implicit since it is defined by the global policy. When you add another universal command, there is only one place to do it. Not 7. And you wont have the typos where you copy/pasted the wrong provider name. And when I look at the provider it is clear what is unique about it which makes more clear what the capabilities are and arent.

lwander: I see what you mean, I'm going to think a little about how I can simplify this. But if it turns into too much work, we'll likely have to merge this and revisit a refactor to keep up with time demands.

ewiseblatt:
sure. as I said i dont know how practical it is, but halyard will have a
lot of configuration stuff in it and a lot of providers so there is a big
cross product in there making it worth considering.

lwander:
@ewiseblatt PTAL at the refactor

lwander:
^ build seems to be failing due to a bug in travis (can no longer find java)

lwander:
@ewiseblatt, I added another commit to remove the SubcommandBuilder and abstract access to the subcommands

PR #41, opened 2016-12-02
Refactor abstract provider naming

lwander:
I didn't like how the subcommands of the concrete provider commands would extend the concrete provider command, so I separated AbstractProviderCommand into AbstractNamedProviderCommand (for things like DockerRegistry, Kubernetes, etc..), and AbstractProviderCommand (for any command that depends on a providerName).


lwander:
Oh man I got git into trouble this time, closing and adding a PR that contains this work + some more.

Issue #5, opened 2016-09-28
Generate full `~/.spinnaker` contents


lwander:
Given a valid halconfig and version, pull the correct authoritative config and generate the user's `~/.spinnaker` directory.

PR #42, opened 2016-12-02
Generate all config

lwander:
@ewiseblatt @duftler @jtk54 PTAL

see https://github.com/spinnaker/halyard/pull/41 for some explanation for the first commit.

This PR allows halyard to walk your halconfig, and collect any nodes that are relevant to specific spinnaker components, and inject those into your spinnaker config.


lwander:
Closes https://github.com/spinnaker/halyard/issues/5

lwander:
tty

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractProviderCommand.java
```diff
@@ -17,65 +17,11 @@
...
-        service.getProvider(currentDeployment, providerName, !noValidate),
-        Providers.translateProviderType(providerName)
-    );
-  }
-
-  @Override
-  protected void executeThis() {
-    AnsiUi.success(getProvider().toString());
-  }
+  abstract protected String getProviderName();
```
jtk54: Seems like a better fit to move this into `AbstractNamedProviderCommand` and get rid of this class. Are you expecting to extend this class without the other?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/GetAccountCommandBuilder.java
```diff
@@ -39,6 +39,6 @@ private GetAccountCommand(String providerName) {
     }
 
     @Getter(AccessLevel.PROTECTED)
-    String providerName;
+    private String providerName;
```
jtk54: I'm not familiar with Lombok, but wouldn't `AccessLevel.PROTECTED` overwrite `private`? Or is that the access modifier for the getter function?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -74,10 +72,15 @@ String halyardVersion(@Value("${Implementation-Version:unknown}") String version
   }
 
   @Bean
-  String spinconfigBucket(@Value("${spinnaker.config.bucket:halconfig}") String spinconfigBucket) {
+  String spinconfigBucket(@Value("${spinnaker.input.config.bucket:halconfig}") String spinconfigBucket) {
     return spinconfigBucket;
   }
 
+  @Bean
```
jtk54: This is a bit strange to read, put the fields first.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -74,10 +72,15 @@ String halyardVersion(@Value("${Implementation-Version:unknown}") String version
   }
 
   @Bean
-  String spinconfigBucket(@Value("${spinnaker.config.bucket:halconfig}") String spinconfigBucket) {
+  String spinconfigBucket(@Value("${spinnaker.input.config.bucket:halconfig}") String spinconfigBucket) {
```
jtk54: Why not just use `@Value` on the fields? Are these consumed by something else? If they are, I'd suggest refactoring this into a `_Config` file and just autowiring that here (and the other places it's necessary) to keep organized.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
```diff
@@ -64,17 +64,37 @@ public NodeFilter withAnyMaster() {
     return this;
   }
 
-  public static NodeFilter makeRejectAllFilter() {
-    return new NodeFilter();
+  public static NodeFilter makeAcceptAllFilter() {
```
jtk54: Seems like you need a `NodeFilterFactory` or `NodeFilterBuilder` alongside the actual `NodeFilter`.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Component
+@Slf4j
+public class GenerateService {
```
jtk54: I'd go with `GenerationService`. Seems weird to have a verb first.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
```diff
@@ -39,6 +37,12 @@ DeploymentConfiguration getDeployment(
       @Path("deployment") String deployment,
       @Query("validate") boolean validate);
 
+  @POST("/v1/config/deployments/{deployment}/generate/")
+  Void generateDeployment(
+      @Path("deployment") String deployment,
+      @Body String _ignore,
```
jtk54: Why are we catching the `@Body` if we're ignoring it? I think you can just exclude it.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/GetAccountCommandBuilder.java
lwander: Yup, so the getter is protected, and the field is private. That's the intention

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractProviderCommand.java
lwander: No, then the other Abstract provider based commands need to extend this, which is odd. Read #41

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
lwander: Oh I didn't know you could just `@Value` fields alone, let me try that.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
lwander: All the other services are named this way, I like having the verb first since the service is effectively the verb in these cases.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
lwander: Because of a retrofit constraint.

jtk54: Ok ('-_-)

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
```diff
@@ -1,4 +1,37 @@
...
+import com.netflix.spinnaker.halyard.cli.services.v1.DaemonService;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+public class GenerateCommand extends NestableCommand {
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "generate";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Generate the full Spinnaker config for you current deployment.";
```
ewiseblatt: "for your current deployment." or "for you, Current Deployment."

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
```diff
@@ -1,4 +1,37 @@
...
+
+  @Override
+  protected void executeThis() {
+    DaemonService service = Daemon.getService();
+    String currentDeployment = service.getCurrentDeployment();
+    service.generateDeployment(currentDeployment, "", false);
```
ewiseblatt: I dont understand how this works. I dont think you discuss it in your design doc yet (the section "deploying and configuring" is TBD, but I didnt look over the rest to see if it is there).

I assume this is generating a deployment configuration and not actually deploying. In section 1.2 of your design doc you say the configuration is never stored on the machine running halyard, only staged there during "distribution of the configuration". I assume that is what we are doing here.

But it seems weird that you ask for the deployment and then generate it.  Why is current-deployment in the halconfig at all? Actually I have a few more questions about halconfig like why is the version information and configuration tied together? I understand that the configuration you write needs to tie together, but this is where the trying would happen. Otherwise it seems that changing a version requires a complete re-specification of the configuration rather than being able to reference a baseline and add some deltas to it (before eventually re-baselining it).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/FileWriter.java
```diff
@@ -16,31 +16,21 @@
...
-
-  public void write(Object node, String path) throws IOException {
-    write(node, path, false);
+public class FileWriter {
+  public void write(String contents, String path) throws IOException {
+    write(contents, path, false);
   }
 
-  public void write(Object node, String path, boolean append) throws IOException {
+  public void write(String contents, String path, boolean append) throws IOException {
```
ewiseblatt: Would you ever be appending? That seems like it would be odd. And would suggest your file is in an inconsistent state (assuming it is structured).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -166,30 +136,15 @@ public void saveConfig() {
       );
     }
 
-    Writer writer = null;
     try {
-      writer = new BufferedWriter(
-          new OutputStreamWriter(new FileOutputStream(halconfigPath), "utf-8"));
-      writer.write(yamlParser.dump(objectMapper.convertValue(halconfig, Map.class)));
-    } catch (Exception e) {
+      fileWriter.write(yamlParser.dump(objectMapper.convertValue(halconfig, Map.class)), halconfigPath);
```
ewiseblatt: Should you be doing something more atomic so as not to trash the old file? Like write to a temp file and move that over the original?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -17,13 +17,49 @@
...
 public class ResourceConfig {
+  /**
+   * Path to where the halconfig file is located.
+   *
+   * @param path Defaults to ~/.hal/config.
+   * @return The path with home (~) expanded.
+   */
+  @Bean
+  String halconfigPath(@Value("${halconfig.filesystem.halconfig:~/.hal/config}") String path) {
+    return replaceBashHome(path);
```
ewiseblatt: Is this supporting Windows? What's special about "Bash"?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -17,13 +17,49 @@
...
+   * Version of halyard.
+   *
+   * This is useful for implementing breaking version changes in Spinnaker that need to be migrated by some tool
+   * (in this case Halyard).
+   *
+   * @param version is the version (seems like the i.d. function for Spring Boot).
+   * @return the version of halyard.
+   */
+  @Bean
+  String halyardVersion(@Value("${Implementation-Version:unknown}") String version) {
```
ewiseblatt: Implementation-Version is not currently universal. I think only Clouddriver and Front50 set it.

For stackdriver I used getClass().getPackage().getImplementationVersion(). That will give you the version of the thing that this is packaged in (e.g. the halyard version) as apposed to the thing that this is deployed in (e.g. clouddriver version if this were embedded at runtime, for example).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -17,13 +17,49 @@
...
+    return version;
+  }
+
+  @Bean
+  String spinconfigBucket(@Value("${spinnaker.input.config.bucket:halconfig}") String spinconfigBucket) {
```
ewiseblatt: why spinnaker.input/output.config as opposed to spinnaker.config.input/output?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -17,13 +17,49 @@
...
+    return spinconfigBucket;
+  }
+
+  @Bean
+  String spinnakerOutputPath(@Value("${spinnaker.output.config.filsystem:~/.halyard/}") String path) {
```
ewiseblatt: filesystem is an odd name. The default is a directory. The trailing '/' is also odd. I would not count on it having that.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -35,4 +71,8 @@ Yaml yamlParser() {
     options.setDefaultFlowStyle(DumperOptions.FlowStyle.BLOCK);
     return new Yaml(options);
   }
+
+  private String replaceBashHome(String path) {
```
ewiseblatt: Bash isnt necessary. I'd make this more generic like "normalizePath".

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -32,6 +36,21 @@
   @JsonIgnore
   public abstract NodeIterator getChildren();
 
+  @JsonIgnore
+  private Set<ComponentName> spinnakerComponents = new HashSet<>();
+
+  protected void registerWithSpinnakerComponent(ComponentName component) {
+    if (spinnakerComponents.contains(component)) {
+      throw new RuntimeException("Node " + getNodeName() + " cannot be registered to spinnaker componenet " + component + " more than once");
```
ewiseblatt: "component"

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
```diff
@@ -64,17 +64,37 @@ public NodeFilter withAnyMaster() {
...
-  public static NodeFilter makeRejectAllFilter() {
-    return new NodeFilter();
+  /**
+   * Modifies the current filter to accept anything the reference does.
+   *
+   * @param reference is the reference to accept.
+   */
+  public NodeFilter extendReference(NodeReference reference) {
+    halconfigFile = !reference.halconfigFile.isEmpty() ? reference.halconfigFile : halconfigFile;
+    deployment = !reference.deployment.isEmpty() ? reference.deployment : deployment;
```
ewiseblatt: Is there a function equivalent to the python x(a,b): b if a.isEmpty() else a

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Providers.java
```diff
@@ -34,6 +35,11 @@
   DockerRegistryProvider dockerRegistry = new DockerRegistryProvider();
   GoogleProvider google = new GoogleProvider();
 
+  protected Providers() {
+    registerWithSpinnakerComponent(ComponentName.CLOUDDRIVER);
```
ewiseblatt: I dont understand what this is doing.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
```diff
@@ -42,7 +42,10 @@
   ValidateService validateService;
 
   public List<Account> getAllAccounts(NodeReference reference) {
-    NodeFilter filter = new NodeFilter(reference).withAnyHalconfigFile().withAnyAccount();
+    NodeFilter filter = NodeFilter.makeRejectAllFilter()
```
ewiseblatt: I get it, but reads odd to say reject everything then extend that with other stuff as opposed to creating an empty filter and building it up. The implementation is exactly the same, but the name of that empty filter is what makes this read odd.

Why not makeEmptyFilter() instead of makeRejectAllFilter()?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -1,4 +1,103 @@
...
+  FileWriter yamlWriter;
+
+  @Autowired
+  Yaml yamlParser;
+
+  @Autowired
+  ObjectMapper objectMapper;
+
+  public void generateConfig(NodeReference nodeReference) {
+    NodeFilter filter = NodeFilter.makeAcceptAllFilter()
```
ewiseblatt: This is odd having just read "makeRejectAllFilter().extendReference()".

At first I thought it was wrong, but I think this is right.
One is narrowing the other is widening. The filter is not relevant to all the data which is why the two are different. But what is odd is that they are both extending.

Dont change anything, i'm just thinking "out loud". Rather than all or none filters you are creating a narrowing (initially all) or widening filter (initially none) then adding into that. When you add into a narrowing filter, it shrinks in scope. When you add into a widening filter, it grows in scope. The implementation is exactly what you are doing but the model is a bit different so the actions have different names and is hopefully clearer.

It's not so bad the way it is now, but AcceptAll.extend seems odd because how can I extend all? And RejectAll.extend seems odd because I think the emphasis is on Reject, not on All so "extend" sounds like shrink (extend the scope of rejection) in which case how can I shrink None.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -1,4 +1,103 @@
...
+      log.info("Writing profile to  " + path);
+      try {
+        boolean append = false;
+
+        for (Node config : configInjectMap.getOrDefault(componentName, new ArrayList<>())) {
+          yamlWriter.write(yamlParser.dump(objectMapper.convertValue(config, Map.class)), path, append);
+          append = true;
+        }
+
+        yamlWriter.write("\n### Fetched config begins here\n", path, append);
```
ewiseblatt: what does this mean and who is this comment for?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -1,4 +1,103 @@
...
+
+        String componentYaml = componentProfile.getProfile(componentName);
+
+        yamlWriter.write(componentYaml, path, append);
+      }
+      catch (IOException ioe) {
+        throw new HalconfigException(
+            new ProblemBuilder(Problem.Severity.FATAL,
+                "Failed to write config for component " + componentName + " to path \"" + path + "\"").build()
```
ewiseblatt: what happens to the file backing yamlWriter? It is corrupted right? Does that matter?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
lwander: The current-deployment is in Halyard to allow multiple sets of Spinnaker configurations to be managed at once.

Version information is tied to the config, because the config in spinnaker/spinnaker/config is liable to change as each subcomponent changes.

What do you mean?
> the configuration you write needs to tie together, but this is where the trying would happen

Odds are, changing a version won't require rewriting your config. Some properties in the config we are extending might change, but that happens without the user knowing about it.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/FileWriter.java
lwander: I am appending, and for a brief time it is possible that it's in an inconsistent state.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
lwander: I think I'll replace the append operation with a string builder, and then write that to disk atomically instead.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: I'm pretty sure this is a bash (or maybe even unix) convention to replace a leading `~` with the user's home directory. If a windows user writes `~` in their config, it'll work identically.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
lwander: I don't think x(a,b) is a python function

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/AccountService.java
lwander: Sounds good

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: Don't I still have to set the version somewhere  for getClass().getPacker().getImplementationVersion()?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
lwander: What if instead of "extend" we called it "refine"?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
lwander: I don't think it is, it's not storing any state.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
ewiseblatt: "where the trying" should heave read "where the tying".

What I'm getting at is since config wont typically change substantially between versions I wouldnt expect the config and version to be intrinsically coupled together. Especially since the most likely config change from one version to the next would be incremental addition of config for a feature added to the new version.

I imagine one would have multiple configs because one has (or is preparing for) multiple deployments or is holding on to something older to revert back to either because it is a change in the conceptual configuration or from a version update.

Intuitively I'd expect configurations to have an intent that is version independent (e.g. production vs testing where you're going to use the same fundamental policies and approach as you evolve from version to version). So I'd expect configurations to have names that reflect that (as you do). Specific versions might require specific deviations from the policy because of new features added or old features deprecated. Given that I'd expect you to have a set of policies/intent configurations that I'm calling baseline configurations, then a set of versions you want to support which might have "configuration deltas" that override (or add to) the baseline. These would probably be per policy.

Maybe it doesnt matter because you never look at the raw config, the tooling provides the diffs, and when you create a new version, you start with the last version and modify it. So you get to the same place from a UX standpoint. It might be more work and complex this way, but you only have one type of config so the processing of the config itself is probably simpler.

Just thinking out loud.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/FileWriter.java
ewiseblatt: The problem with being inconsistent is if you never complete (crash, etc) then you have corrupted your file if you were rewriting it or dont know it is corrupted if it is new.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
ewiseblatt: An atomic disk write is a mv(2) where you are just swapping inodes but the content is already there. I dont think you can assume write(2) is atomic. You dont need the builder per say. I think you need a temporary file as an intermediary.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
ewiseblatt: My point was more about syntactic sugar for the pattern here than the specific syntax itself.
Is there a simplifying Java notation you can apply here?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
ewiseblatt: Yes. I was assuming you are reflecting here (Implementation-Version is for halyard runtime, not something you are imposing on it from the outside). If so, then halyard is running from a build where gradle packaged this module with a version. getPacakge().getImplementationVersion() will be that version from Java's perspective. Note it is "Package" not "Packer". I assume you have a typo and did not misread.

At least, that is the empirical evidence I have from trying to look at versions from a kork library. I went that route to get the kork version (which I can argue makes more sense for what I wanted) where the spring property was rarely defined in current microservices.

You can go the above route, just be aware that you'll have to change how they are currently built for this to work. Since you ask for "halyard version" specifically, I think you want the halyard package version and not the microservice runtime version. But that might not be your intent, in which case dont call this the "halyard" version.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
lwander: Well in some way we need the user to express what version of Spinnaker they want to deploy, and I also want to be ready for the scenario where there is a breaking config change that needs to be addressed by halyard or the user. I don't see a way around these two things without writing the version of spinnaker to deploy in the config. 

I think once you start having the user think in config deltas, we are making the upgrade path too complicated - ideally the 90% case will be a config transform where we rename/replace fields on behalf of the user. Other config changes happen in the background and are invisible to the user when the upgrade to a new version.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/FileWriter.java
lwander: Yup agreed, I think I left a comment below saying I'm switching to an atomic update.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
lwander: Sounds good

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/GenerateCommand.java
ewiseblatt: ok.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: No good reason, switching

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/NodeFilter.java
lwander: I think the ternary operator is as simple as it gets, unless we have an Optional type here. I made a helper to clean this up.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: I guess I don't understand, what's the difference between the two?

ewiseblatt: The difference if halyard is a component embedded in another service (e.g. the service is an endpoint in some uber-admin service) as opposed to the top-level service (halyard is standalone). If embedded then the spring parameter is for the top level spring thing and the halyard package is the version of halyard that service is using. If they are the same, then they should be the same, except that the spring parameter may not have been defined.

PR #43, opened 2016-12-19
add-account command implemented for k8s

lwander:
The client builds the account object for a specific provider, and sends it to the daemon for validation. If it passes, the halconfig is updated.

@duftler or @jtk54 or @ewiseblatt or @ttomsu PTAL

### DIFF
--- a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
+++ b/halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
@@ -16,40 +16,53 @@
 
 package com.netflix.spinnaker.halyard.controllers.v1;
 
-import com.netflix.spinnaker.halyard.config.model.v1.node.NodeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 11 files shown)


halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
```diff
@@ -59,4 +72,57 @@ Account account(
...
+        .setDeployment(deploymentName)
+        .setProvider(providerName)
+        .setAccount(accountName);
+
+    Account account = objectMapper.convertValue(
+        rawAccount,
+        Providers.translateAccountType(providerName)
+    );
+
+    Runnable doUpdate = () -> accountService.setAccount(reference, account);
```
jtk54: Seems like the same pattern here and in the function below. Fine for now with just the 2, any more and you should break that out into a private function.

lwander: Good call - for now they call different functions `getAccount` vs `setAccount` but I'll generify it once I see a chance.

PR #44, opened 2016-12-19
Allow user to specify what severity level should abort an operation

lwander:
Previously, any problem (including a warning) would cause an operation to fail. Now, all responses from the daemon are wrapped in a `DaemonResponse` which provides the result of the operation alongside any problems encountered that didn't exceed the user specified severity level. If any problem is encountered that exceeds the severity level an exception is thrown as before, and the operation response is omitted. 

@duftler or @ttomsu or @ewiseblatt or @jtk54 PTAL


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/ProblemSetWriter.java
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.ui.v1;
+
+import com.netflix.spinnaker.halyard.config.model.v1.problem.Problem;
+import com.netflix.spinnaker.halyard.config.model.v1.problem.Problem.Severity;
+import com.netflix.spinnaker.halyard.config.model.v1.problem.ProblemSet;
+
+public class ProblemSetWriter {
```
ttomsu: Classes with suffix `-Writer` generally implement the `Writer` interface. 

https://docs.oracle.com/javase/7/docs/api/java/io/Writer.html

You might consider renaming to stay within the common conventions, or actually make it implement `Writer`.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractProviderEnableDisableCommand.java
```diff
@@ -52,8 +53,8 @@ public String getDescription() {
 
   private void setEnable() {
     DaemonService service = Daemon.getService();
-    String currentDeployment = service.getCurrentDeployment();
-    service.setProviderEnabled(currentDeployment, getProviderName(), !noValidate, isEnable());
+    String currentDeployment = ResponseUnwrapper.get(service.getCurrentDeployment());
```
ttomsu: Instead of having `ResponseUnwrapper` everywhere, have you considered returning a proxy class from `Daemon.getService` that will do the unwrapping for you?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/UpdateService.java
```diff
@@ -26,15 +28,17 @@
   @Autowired
   HalconfigParser halconfigParser;
 
-  public void safeUpdate(Runnable update, Runnable validate) {
+  public ProblemSet safeUpdate(Runnable update, Supplier<ProblemSet> validate) {
```
ttomsu: `validator`, maybe?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/ProblemSetWriter.java
ewiseblatt: I see his point though am not as troubled by it. Your AnsiUi doesnt really have an interface that error, etc use. They are based on AnsiPrinter but this isnt quite that. The method here ultimately is emitting to an AnsiPrinter, though doing so via the AnsiUI so it isnt obvious to me what to name this.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/problem/ProblemSet.java
```diff
@@ -50,10 +50,11 @@ public ProblemSet() { }
   /**
    * This will throw an exception if this problem set stores any problems.
    */
-  public void throwIfProblem() {
+  public ProblemSet throwIfProblem() {
```
ewiseblatt: Is this a common pattern? Otherwise why does this return a ProblemSet when add() and sort() dont? If anything I'd expect the other way around.
It seems odd to be returning a problem set since you would not expect these methods to be mutable.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/problem/ProblemSetBuilder.java
```diff
@@ -46,6 +50,6 @@ public ProblemSet build() {
         .map(ProblemBuilder::build)
         .collect(Collectors.toList());
 
-    return new ProblemSet(problems);
+    return new ProblemSet(problems).throwifSeverityExceeds(severity);
```
ewiseblatt: saving a pair of statements here does not justify to me the non-standard change to the public interface.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DefaultControllerValues.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.controllers.v1;
+
+public class DefaultControllerValues {
```
ewiseblatt: Why have this?
Do you really want all the methods to have the same defaults as a matter of policy? Why would you change it from the default?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/problem/ProblemSetBuilder.java
lwander: Very fair, I'll change back the public interface

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DefaultControllerValues.java
lwander: For development purposes I have the default validate behavior set to "false", which won't be the case when this is released. This makes switching this back easier.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractProviderEnableDisableCommand.java
lwander: How would I do that with retrofit?

lwander: I could cook something up with reflection, but that feels ugly

ttomsu: It's another layer of indirection - the CLI portion has a class that is composed of the DaemonService.

lwander: Ah I see what you mean, yeah I agree that's nicer, I'll switch it over

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/DefaultControllerValues.java
ewiseblatt: SGTM

lwander:
All comments addressed

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/ui/v1/ProblemSetWriter.java
lwander: This was only being used by the ResponseUnwrapper, so I called the write method `formatProblemSet` and folded it in there.

PR #50, opened 2016-12-21
Google add account command

lwander:
@duftler


lwander:
addaccount

duftler:
Looks like the use of terminating periods for the parameter descriptions is a bit inconsistent.

lwander:
Whoops, will fix that

PR #58, opened 2017-01-04
Edit kubernetes account command

lwander:
@duftler

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1.providers;
+
+import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.cli.command.v1.NestableCommand;
+import com.netflix.spinnaker.halyard.cli.services.v1.Daemon;
+import com.netflix.spinnaker.halyard.cli.ui.v1.AnsiUi;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import lombok.AccessLevel;
+import lombok.Getter;
+
+@Parameters()
+public abstract class AbstractEditAccountCommand<T extends Account> extends AbstractHasAccountCommand {
+  @Getter(AccessLevel.PROTECTED)
+  private Map<String, NestableCommand> subcommands = new HashMap<>();
+
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "edit-account";
+
+  protected abstract Account editAccount(T account);
+
+  public String getDescription() {
+    return "Edit a " + getProviderName() + " account.";
+  }
+
+  @Override
+  protected void executeThis() {
+    String accountName = getAccountName();
+    String providerName = getProviderName();
+    String currentDeployment = Daemon.getCurrentDeployment();
+    Account account = Daemon.getAccount(currentDeployment, providerName, accountName, !noValidate);
+
+    Daemon.setAccount(currentDeployment, providerName, accountName, !noValidate, editAccount((T) account));
+    AnsiUi.success("Edited " + providerName + " account \"" + accountName + "\"");
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Provides consistent update semantics for updating a list of entries given a new list, entries to remove, and entries to add.
+   *
+   * @param old is the prior set of entries - these are modified to addTo and removeFrom.
+   * @param setTo is a new set of entries. If provided, the old ones are discarded.
+   * @param addTo is a single entry to add to old.
+   * @param removeFrom is a single entry to remove from old.
+   * @return the updates set of entries.
```
duftler: s/updates/updated

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
```diff
@@ -0,0 +1,92 @@
...
+   * @throws IllegalArgumentException when setTo && (addTo || removeFrom) are provided.
+   */
+  protected static List<String> updateStringList(List<String> old, List<String> setTo, String addTo, String removeFrom) {
+    if (!(old != null && !old.isEmpty())) {
```
duftler: I think it would read easier if you apply the negation to the expression, but can live with it if you really prefer it this way.

duftler: If {old is null or empty}...

duftler: Why do you have to initialize it if it is already non-null and empty? Is it to ensure it is mutable?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    boolean set = setTo != null && !setTo.isEmpty();
+    boolean add = addTo != null && !addTo.isEmpty();
+    boolean remove = removeFrom != null && !removeFrom.isEmpty();
+
+    if (set && !(add || remove)) {
+      return setTo;
+    } else if (set && (add || remove)) {
+      throw new IllegalArgumentException("Both set and addTo or removeFrom cannot be set");
```
duftler: Maybe: "If set is specified, neither addTo nor removeFrom may be specified." Or "Specify either set or one of (addTo, removeFrom)."

duftler: Do you think it would flow better if this `else if` was moved out above the `if/else-if/else` block entirely. Then the block is just left with the `if/else`. Feels tough to read with the exception potentially thrown from the middle.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/kubernetes/KubernetesEditAccountCommand.java
```diff
@@ -0,0 +1,129 @@
...
+  @Parameter(
+      names = "--context",
+      description = "The kubernetes context to be managed by Spinnaker. "
+          + "See http://kubernetes.io/docs/user-guide/kubeconfig-file/#context for more information."
+  )
+  private String context;
+
+  @Parameter(
+      names = "--clear-context",
+      description = "Removes the currently specified context, defaulting to 'current-context' in your kubeconfig."
```
duftler: Is 'specified context' here referring to the same thing as the 'managed context' from the description of the `--context` parameter just above?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/kubernetes/KubernetesEditAccountCommand.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Override
+  protected Account editAccount(KubernetesAccount account) {
+    boolean contextSet = context != null && !context.isEmpty();
+    if (contextSet && !clearContext) {
+      account.setContext(context);
+    } else if (!contextSet && clearContext) {
+      account.setContext(null);
+    } else if (contextSet && clearContext) {
+      throw new IllegalArgumentException("Both --context and --clear-context cannot be set");
```
duftler: Feel free to ignore this comment entirely, but I think it is sometimes clearer to say exactly what to do instead of what not to do. So something like "Specify either --context or --clear-context." Eh, maybe not. I read it too many times and can't tell anymore...

lwander:
Addressed all comments, thanks @duftler

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/providers/AbstractEditAccountCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    boolean set = setTo != null && !setTo.isEmpty();
+    boolean add = addTo != null && !addTo.isEmpty();
+    boolean remove = removeFrom != null && !removeFrom.isEmpty();
+
+    if (set && !(add || remove)) {
+      return setTo;
+    } else if (set && (add || remove)) {
+      throw new IllegalArgumentException("If set is specified, neither addTo nor removeFrom cannot be specified");
```
duftler: s/cannot/can

PR #62, opened 2017-01-11
Refactor config generation

lwander:
The old approach was to label nodes in the halconfig as belonging to certain
spinnaker components, and then appending that node to that component's config
when it was written. The benefit of that approach was it made it easy to
connect the halconfig to the spinnaker component's config, but it made it
impossible to generate more complicated config, and forced the halconfig model
to depend on the config generation service.

I've rewritten it so each service decides how to generate its own config.

@duftler @ewiseblatt @jtk54 PTAL


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -36,21 +32,6 @@
   @JsonIgnore
   public abstract NodeIterator getChildren();
 
-  @JsonIgnore
-  private Set<ComponentName> spinnakerComponents = new HashSet<>();
```
lwander: This is where the halconfig model depended on the config generation by labeling itself with the which compenent's config to write.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -16,90 +16,49 @@
...
+  @Autowired(required = false)
+  List<SpinnakerComponent> spinnakerComponents = new ArrayList<>();
 
   public void generateConfig(NodeReference nodeReference) {
-    NodeFilter filter = NodeFilter.makeAcceptAllFilter()
-        .refineWithReference(nodeReference);
-
-    List<Node> matching = lookupService.getMatchingNodesOfType(filter, Node.class);
-
-    Map<ComponentName, List<Node>> configInjectMap = new HashMap<>();
```
lwander: This is where the halconfig was traversed to see which node to append to which component.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/spinnaker/v1/component/ClouddriverComponent.java
```diff
@@ -0,0 +1,33 @@
...
+@Component
+public class ClouddriverComponent extends SpringComponent {
+  @Override
+  public String getComponentName() {
+    return "clouddriver";
+  }
+
+  @Override
+  public String generateFullConfig(String baseConfig, DeploymentConfiguration deploymentConfiguration) {
+    return yamlToString(deploymentConfiguration.getProviders()) + "\n" + baseConfig;
```
lwander: This is an example of attaching the halconfig providers section to the base config found here: https://github.com/spinnaker/clouddriver/tree/master/halconfig

jtk54: The set of key/value pairs in `deploymentConfiguration.getProviders()` doesn't have any intersection with what's in `baseConfig`, right? Otherwise I think you could overwrite keys in the latter overwrite the former.

lwander: The idea is that if they do, halyard's key/values should prevail.

PR #64, opened 2017-01-18
Assorted quality of life improvements

lwander:
1. Clean up yaml dumping style for [scalars](http://yaml.org/spec/current.html#id2503232).
2. Set per-deployment versions, defaulting to "latest" when not specified.
3. Prefix generated config with a warning that this file was auto-generated (seems like the easiest short-term solution to our discussion today @duftler)


lwander:
Strengthed warning messages

PR #65, opened 2017-01-18
Add support for logging options.

lwander:
Any field that has a corresponding "fieldName"Options() method will supply alternative entries during validation.

I also updated the formatting of errors to group them by location.

validate-options

@duftler @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
@@ -17,7 +17,13 @@
 package com.netflix.spinnaker.halyard.config.model.v1.node;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.halyard.config.model.v1.problem.ProblemSetBuilder;
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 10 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -61,6 +68,36 @@ public boolean matchesToRoot(NodeFilter filter) {
...
+    try {
+      optionsMethod = this.getClass().getDeclaredMethod(fieldOptions, ProblemSetBuilder.class);
+      optionsMethod.setAccessible(true);
+
+      return (List<String>) optionsMethod.invoke(this, problemSetBuilder);
+    } catch (IllegalAccessException | InvocationTargetException e) {
+      log.warn("Failed to call " + fieldOptions + "() on " + this.getClass());
+      e.getCause().printStackTrace();
+
+      throw new RuntimeException(e.getCause());
```
jtk54: Won't printing the cause and then re-throwing a `RuntimeException` with that cause print the stack trace twice?

lwander: Whoops - that was there from debugging.

PR #68, opened 2017-01-18
Add more k8s option, punctuate properly, fix location formatting

lwander:
@duftler or @jtk54 PTAL


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -70,4 +74,19 @@ public void accept(ProblemSetBuilder psBuilder, Validator v) {
         .map(NamedContext::getName)
         .collect(Collectors.toList());
   }
+
+  protected List<String> dockerRegistriesOptions(ProblemSetBuilder psBuilder) {
```
lwander: I realize this is named funny (`dockerRegistryOptions` sounds better), but the reflection for option creation requires the prefix of the variable name to match the options method.

PR #72, opened 2017-01-26
Bootstrap clouddriver in the kubernetes deployment

lwander:
This bootstraps clouddriver and opens a connection for other microservices to be deployed. I need to write secrets to the cluster now, and that feels like a significant piece of work for a separate PR.

@duftler @jtk54 PTAL


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
```diff
@@ -0,0 +1,69 @@
...
+    return true;
+  }
+
+  @Override
+  public NodeReference getReference() {
+    return parent.getReference();
+  }
+
+  public enum DeploymentType {
+    ClusteredSimple("Deploy Spinnaker with one server group and load balancer "
```
jtk54: This is still pulling packages with `apt-get`, correct? Just the package of the microservice deployed on the server group? Or does this use Spinnaker for deployment?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Getter
+    final String description;
+
+    DeploymentType(String description) {
+      this.description = description;
+    }
+  }
+
+  DeploymentType type;
```
jtk54: Are these meant to be strictly `private`? I think the default scope for fields in Java is `protected`.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/component/v1/ServiceFactory.java
```diff
@@ -17,15 +17,17 @@
...
 
-public class ServiceFactory<T extends ComponentService> {
+@Component
+public class ServiceFactory {
   @Autowired
   OkClient okClient;
 
-  public T createService(String endpoint, ComponentType componentType) {
-    Class<T> clazz = (Class<T>) componentType.getServiceClass();
+  public Object createService(String endpoint, ComponentType componentType) {
```
jtk54: Seems weird that a `ServiceFactory` returns an `Object`, why not `ComponentService` similar to before, or `RestAdapter`?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DeploymentFactory.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  private Deployment createClusteredSimpleDeployment(DeploymentConfiguration deploymentConfiguration) {
+    DeploymentEnvironment deploymentEnvironment = deploymentConfiguration.getDeploymentEnvironment();
+    String accountName = deploymentEnvironment.getAccountName();
+
+    if (accountName == null || accountName.isEmpty()) {
+      throw new HalconfigException(new ProblemBuilder(Severity.FATAL, "An account name must be "
+          + "specified the desired place to run your simple clustered deployment.").build());
```
duftler: Is there a word missing somewhere in here?

lwander: Whoops, yup - fixing

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/component/v1/ServiceFactory.java
jtk54: Does this return a retrofit `interface` which is an `Object` only? I'm curious now, I haven't worked with these.

lwander:
I'll replace the templating if this gets too complicated - pebble seems like a good alternative.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
lwander: No - this'll use baked images for the various cloud providers.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/component/v1/ServiceFactory.java
lwander: Retrofit doesn't allow you to generate stubs for interfaces that extend other interfaces: https://github.com/square/retrofit/issues/504

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
lwander: Oops, good call

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
jtk54: Makes sense, where will the baked images come from?

lwander: Good question, either

1. a locally running rosco
2. the packer-manipulating bits of rosco pulled out of the published jars
3. or simply a few calls to packer

PR #76, opened 2017-01-28
Allow user to specify per-deployment services

lwander:
This is a migration path to halyard, as well as a tool for specifying more complicated config.

@duftler or @jtk54 PTAL

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
@@ -22,22 +22,26 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.stereotype.Component;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
-import org.yaml.snakeyaml.introspector.Property;
-import org.yaml.snakeyaml.nodes.NodeTuple;
-import org.yaml.snakeyaml.nodes.Tag;
-import org.yaml.snakeyaml.representer.Representer;
+
+import java.io.File;
+import java.nio.file.Paths;
 
 @Component
 public class ResourceConfig {
   /**
-   * Path to where the halconfig file is located.
+   * Directory containing the halconfig.
    *
-   * @param path Defaults to ~/.hal/config.
+   * @param path Defaults to "~/.hal".
    * @return The path with home (~) expanded.
    */
   @Bean
-  String halconfigPath(@Value("${halconfig.directory.halconfig:~/.hal/config}") String path) {
+  String halconfigDirectory(@Value("${halconfig.directory.halconfig:~/.hal}") String path) {
     return normalizePath(path);
+  }
+
+  @Bean
+  String halconfigPath(@Value("${halconfig.directory.halconfig:~/.hal}") String path) {
+    return normalizePath(Paths.get(path, "config").toString());
   }
 
   /**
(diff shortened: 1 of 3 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -78,6 +81,12 @@ Yaml yamlParser() {
   }
 
   private String normalizePath(String path) {
-    return path.replaceFirst("^~", System.getProperty("user.home"));
+    String result = path.replaceFirst("^~", System.getProperty("user.home"));
+    // Strip trailing path separator
+    if (result.matches(File.separator + "$")) {
```
duftler: Using `String.endsWith()` would be more efficient and easier to read than the regex.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
```diff
@@ -22,24 +22,27 @@
...
    */
   @Bean
-  String halconfigPath(@Value("${halconfig.directory.halconfig:~/.hal/config}") String path) {
+  String halconfigDirectory(@Value("${halconfig.directory.halconfig:~/.hal}") String path) {
     return normalizePath(path);
   }
 
+  @Bean
+  String halconfigPath(@Value("${halconfig.directory.halconfig:~/.hal}") String path) {
+    return normalizePath(path + File.separator + "config");
```
duftler: Can live with this, but you may want to use a combination of the `File` constructors and `File.getCanonicalPath()` to avoid doing the string manipulation yourself.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
```diff
@@ -18,30 +18,62 @@
...
+    File spinnakerOutput = new File(spinnakerOutputPath);
+
+    try {
+      FileUtils.deleteDirectory(spinnakerOutput);
+    } catch (IOException e) {
+      throw new HalconfigException(
+          new ProblemBuilder(Severity.FATAL, "Unable to clear old spinnaker config: " + e.getMessage() + ".").build());
+    }
+
+    if (!spinnakerOutput.mkdir()) {
```
duftler: If you want to also create missing intermediate dirs, you can use `mkdirs()` here.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
duftler: Or `Paths/Path` like you are doing down below.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/GenerateService.java
lwander: Ah perfect, thanks

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: Good idea

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/ResourceConfig.java
lwander: Yeah that sounds better

PR #78, opened 2017-01-28
Display version options using hal

lwander:
@danielpeach PTAL

versions-hal


lwander:
@skim1420 please disregard

skim1420:
Please make them alpha ascending. Please name the first one Albatross.

lwander:
Will do - 1.0 will be Spinnaker Albatross

PR #82, opened 2017-01-30
Fix retrofit error suppression

lwander:
Errors without a problem set causing halyard to fail without an explanation.

The object mapper was also being instantiated incorrectly, using a bean definition from another imported project causing it to ignore unknown properties. @duftler or @danielpeach or @jtk54 PTAL


lwander:
ping

PR #83, opened 2017-01-31
Track required files during config generation

lwander:

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Consumer;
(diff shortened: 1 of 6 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -137,4 +138,15 @@ public void parentify() {
         .filter(f -> f.getDeclaredAnnotation(LocalFile.class) != null)
         .collect(Collectors.toList());
   }
+
+  public void recursiveConsume(Consumer<Node> consumer) {
+    consumer.accept(this);
+
+    NodeIterator children = getChildren();
+    Node child = children.getNext();
+    while (child != null) {
```
danielpeach: Could this loop be replaced with a `children.forEachRemaining`? Or is children not an iterator?

lwander: It doesn't implement the iterator interface because it has special features for filtering nodes.

PR #100, opened 2017-02-03
Refactor node filters

lwander:
Now the filters decide which nodes to match against, instead of having the nodes match against the filters using hacky string/glob comparison. 

@jtk54 PTAL


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -148,4 +157,46 @@ public void recursiveConsume(Consumer<Node> consumer) {
       child = children.getNext();
     }
   }
+
+  /**
```
jtk54: Duplicate `/**`

PR #101, opened 2017-02-03
Sort help & subcommand flags

lwander:
This makes reading the command's help a little easier, as well as pulls `--help` into a global parameters (which it is).


lwander:
@danielpeach

PR #104, opened 2017-02-03
Allow halyard to stream task results

lwander:
halstream


lwander:
@duftler @jtk54 @danielpeach PTAL

lwander:
The motivation was that deploy & generate can take some time to complete, and the user is left without feedback while the CLI has a long-running connection open against the Daemon (that can timeout). Changing these long running events to a streaming model makes it easier for the user to process.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/ResponseUnwrapper.java
```diff
@@ -21,17 +21,47 @@
...
+
+      try {
+        Thread.sleep(WAIT_MILLIS);
+      } catch (InterruptedException ignored) {
+      }
+
+      task = Daemon.getTask(task.getUuid());
+    }
+    
+    events = task.getEvents();
```
duftler: Maybe pull this common routine out?

Issue #106, opened 2017-02-04
Configure Front50 Storage Options


lwander:
This will probably look something like

```yaml
persistentStorage:
  accountName: my-[aws/google]-account
  bucketName: # either user specified, or randomly generated by halyard
  rootFolder: spinnaker
```

PR #110, opened 2017-02-04
Configure (without validating) front50

lwander:


lwander:
https://github.com/spinnaker/halyard/issues/106

PR #113, opened 2017-02-05
Adds basic tab-completion support to halyard

lwander:

This doesn't (yet) do fancy things like ask the daemon for options, but generates completion help for all statically known subcommands & flags. If you're interested in how it works, read [this](https://debian-administration.org/article/317/An_introduction_to_bash_completion_part_2).


lwander:
completion

PR #121, opened 2017-02-06
Adds dockerfile for build triggers & GCR

lwander:
@ttomsu PTAL (pretty much copied from fiat)


lwander:
Please use the "squash and merge" option - it keeps the commit history much cleaner.

PR #134, opened 2017-02-12
Improve hal's presentation of event messages

lwander:
See: 
tty


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/tasks/v1/DaemonTask.java
```diff
@@ -8,29 +8,49 @@
 
 /**
  * This represents a long-running task managed by the Daemon that can be polled for status information.
+ * It is made of multiple stages, each of which have multiple events.
```
duftler: Seems like a word is missing between 'made' and 'of'. Maybe "made up of" or "composed of"?

lwander: Whoops, yup

Issue #108, opened 2017-02-04
Configure Spinnaker Access/Security


lwander:
The gist of the features is to make [this](http://www.spinnaker.io/docs/securing-spinnaker) easier. Since I don't have much experience securing Spinnaker: @ttomsu, could you list a few common sticking points during this config process?

lwander:
So my thoughts on preliminary config are

```yaml
security:
  apiDomain:    # fully qualified domain name, `null` when none exists
  apiAddress:   # defaults to localhost
  uiDomain:     # fully qualified domain name, `null` when none exists
  uiAddress:    # defaults to localhost
  ssl:
    enabled:
    serverKeyPath:
    caKeyPath:
    keystorePassword:   # one password for all keys due to tomcat limitation
    keystoreAlias:
    keystorePath:
  authn:
    enabled:    # when false, assume/create ssh tunnels
    oauth2:
      # all gate - spring.oauth settings go here (client, resource userInfoMapping)
    saml2:
    x509:
  authz:
    # gafw config
```

lwander:
@ttomsu PTAL ^

lwander:
Closing in favor of https://github.com/spinnaker/halyard/issues/221 & https://github.com/spinnaker/halyard/issues/231

PR #138, opened 2017-02-13
Start of halyard config security

lwander:
https://github.com/spinnaker/halyard/issues/108

This is the smallest amount of work needed to get Deck to read Gates new public URL by adding support for SSL configuration & public UI + API addresses. @ttomsu and I figure it's best to do TLS across all the Spinnaker services (when it's enabled) so it's an all-or-nothing approach there.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Authn.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Authn.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.security;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Node;
+import com.netflix.spinnaker.halyard.config.model.v1.node.NodeIterator;
+import com.netflix.spinnaker.halyard.config.model.v1.node.NodeIteratorFactory;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+
+@Data
+public class Authn extends Node {
+  @Override
+  public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
+    v.validate(psBuilder, this);
+  }
+
+  @Getter
+  private String nodeName = "authn";
+
+  @Override
+  public NodeIterator getChildren() {
+    return NodeIteratorFactory.makeEmptyIterator();
+  }
+}
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ssl.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ssl.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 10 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Authn.java
```diff
@@ -0,0 +1,45 @@
...
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class Authn extends Node {
+  @Override
+  public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
+    v.validate(psBuilder, this);
+  }
+
+  @Override
+  public String getNodeName() {
```
ttomsu: This looks like it can be reduced to
```
@Getter
private String nodeName = "authn"
```

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Authn.java
```diff
@@ -0,0 +1,45 @@
...
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class Authn extends Node {
+  @Override
+  public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
```
ttomsu: The `accept` and `getChildren` methods look like a lot of duplicated boilerplate - can you pull these default implementations into the `Node` class and make them protected methods than can be overwritten by subclasses that need to?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/SpinnakerEndpoints.java
```diff
@@ -54,5 +75,29 @@
     public String getBaseUrl() {
       return protocol + "://" + address + ":" + port;
     }
+
+    Service(Security security) {
+      if (security.getSsl().isEnabled()) {
+        protocol = "https";
```
ttomsu: Does this affect even single machine deployments? Intra-machine SSL on a single instance is not an efficient use of CPU

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/SpinnakerEndpoints.java
```diff
@@ -54,5 +75,29 @@
...
+      }
+    }
+  }
+
+  /**
+   * Like a Service, but this has a publicly accessible endpoint (that is likely different from the one
+   * assigned within the Spinnaker cluster).
+   */
+  @Data
+  @EqualsAndHashCode(callSuper = true)
```
ttomsu: Why is this `callSuper` true and all of the others are false?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ssl.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.netflix.spinnaker.halyard.config.model.v1.security;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.*;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
```
ttomsu: callSuper is false by default

ttomsu: I would also strongly consider adding `lombok.equalsAndHashCode.callSuper` to your `lombok.config` file with your preferred value, instead of having `EqualsAndHashCode` overrides everywhere.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Authn.java
lwander: `accept`: AFAIK the visitor pattern only works like this (to make sure the right class type is passed to the validator).

`getChildren` The thing is, the common implementation (empty iterator) is very wrong in many cases. I want to draw attention to the fact that this needs to be implemented for any new contributors so they don't get bitten by an incorrect (although typically functional) default.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ssl.java
lwander: Ah didn't know about that lombok config line, that seems useful

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/SpinnakerEndpoints.java
lwander: For now, yes, but I'll focus on that once we get further along.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/SpinnakerEndpoints.java
lwander: Because here I do care about fields in the super-classes implementation for equality & hashing. In the other cases (subclasses of `Node`) the superclass doesn't carry useful info.

Issue #116, opened 2017-02-05
Support AWS configuration


lwander:

lwander:
The linked PRs & details in https://github.com/spinnaker/halyard/issues/71 will be helpful.

There is an additional requirement to support front50 config - see https://github.com/spinnaker/halyard/issues/106

lwander:
The clouddriver config & cli commands are added by #146

lwander:
Here's the remaining work:

1. Validate your AWS account by extending [this](https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/aws/AwsAccountValidator.java).
2. Configure your rosco images. This https://github.com/spinnaker/halyard/pull/147 does a good job of showing how that can be done. Note, I don't have the story figured out yet for dealiing with template files, and overriding base images - that'll come later.
3. [Configure settings.js](https://github.com/spinnaker/halyard/blob/master/docs/adding-a-provider.md#5-update-settingsjs).
4. Validate Front50 S3 config. This https://github.com/spinnaker/halyard/pull/189 gets the job started, but I'm not sure how Front50 deals with credentials.

lwander:
@gardleopard

lwander:
@gardleopard following our convo this morning:

```bash
hal config provider aws account add --<some parameters>
hal config provider aws enable
hal config storage aws edit --<some other parameters>
hal config storage edit --type aws

hal deploy edit --type localdebian
sudo hal deploy apply
```

is the minimal set of commands that should be needed to configure & deploy to aws.

gardleopard:
@lwander `hal config storage aws edit --<some other parameters>` should be `s/aws/s3/g` right?

lwander:
ah yes, that's right, sorry about that. I'll edit the above message

lwander:
Closing this in favor of a more descriptive bug.

lwander:
https://github.com/spinnaker/halyard/issues/537

PR #146, opened 2017-02-15
AWS config

lwander:
https://github.com/spinnaker/halyard/issues/116

@duftler PTAL

@gardleopard PTAL

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/AbstractAddAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/AbstractAddAccountCommand.java
@@ -32,7 +32,7 @@ public abstract class AbstractAddAccountCommand extends AbstractHasAccountComman
   private Map<String, NestableCommand> subcommands = new HashMap<>();
 
   @Getter(AccessLevel.PUBLIC)
-  private String commandName = "add-account";
+  private String commandName = "add";
 
   protected abstract Account buildAccount(String accountName);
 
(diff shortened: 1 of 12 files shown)


lwander:
This is missing any sort of validation (this is where I'll need help from someone with more AWS knowledge than me)

gardleopard:
I tested this a little bit now. Clouddriver would not accept the generated account config.

 
my working clouddriver-local.yml
```
  - accountId: "01337"
    discovery: "http://{{region}}.eureka.url.to.use:8080/eureka-server/v2"
    name: leet-aws-account
    regions:
    - {name: eu-west-1}
```
.halyard/clouddriver.yml
```
  - name: leet-aws-account
    defaultKeyPair: null
    edda: null
    discovery: http://{{region}}.eureka.url.to.use:8080/eureka-server/v2
    accountId: '01337'
    regions:
    - eu-west-1
```

.hal/config
```
      - name: leet-aws-account
        defaultKeyPair: null
        edda: null
        discovery: http://{{region}}.eureka.url.to.use:8080/eureka-server/v2
        accountId: '01337'
        regions:
        - eu-west-1
```

generated with:
```
./halyard-cli/bin/halyard-cli config provider aws account add infra-del-sndbx-test --accountId 01337  --regions eu-west-1 --discovery "http://{{region}}.eureka.eureka.url.to.use:8080/eureka-server/v2"
```

gardleopard:
the doc for discovery should probably have a mention about `http://{{region}}.eureka.url.to.use:8080/eureka-server/v2`, because region is magic in this context.
```
--discovery
    The endpoint your Eureka discovery system is reachable at. See
    https://github.com/Netflix/eureka for more information.
```

lwander:
Please try the add account command again, the `region` field structure was wrong.

lwander:
What do you suggest the help text should be for the discovery param?

gardleopard:
Im not sure what the help text should be. Could add something like:
`Example: http://{{region}}.eureka.url.to.use:8080/eureka-server/v2 {{region}} will make spinnaker use regions in the hostname while accessing discovery so that you can have discovery for multiple regions`

lwander:
Sounds good, updated.

lwander:
awshelp

gardleopard:
The config worked now. I nuked .hal/config, set up the aws account, generated config and copied the `accounts:` list to our clouddriver instance. I tried with two accounts and it works.

PR #147, opened 2017-02-15
Add optional packer params to Azure config

EricJizbaMSFT:


EricJizbaMSFT:
@scotmoor @rguthriemsft @lwander PTAL

Note: when testing I had some weirdness around the structure of the cli. For example, I had to run "./hal config provider **azure azure** edit test-account --packer-resource-group <> --packer-storage-account <>" with two 'azure's. Also, the 'edit-account' command changed to 'edit', but the 'add-account' command is still the same. I'm 99.9% sure my changes are not the cause of this weirdness, but thought I'd mention anyways

lwander:
Sorry about the CLI weirdness, I'm changing the command structure - I'll have that fixed in the next PR or so.

Where in the overall config do these packer parameters go? If it's in rosco, why are these scoped to a clouddriver account in the config?

EricJizbaMSFT:
@lwander Wow yes very good point about Rosco. Do you have an example of scoping something to Rosco in the code? Is that possible at the moment?

lwander:
No, that's on my slate for next week, maybe late this week. Let's keep this PR open for now and amend it once we are ready to export Rosco config?

However, front50 is configurable right now with Halyard and I understand you recently added support for azure in front50?

EricJizbaMSFT:
Sure we can keep it open. Yes we added support for front50 - I can look into that separately

lwander:
Hey @EricJizbaMSFT, take a look at https://github.com/spinnaker/halyard/pull/167 and https://github.com/spinnaker/halyard/pull/160 as a guideline for this.

EricJizbaMSFT:
@scotmoor @rguthriemsft PTAL. I've added full Azure support for Rosco (originally this PR only had the two packer parameters, but now it has base image support and minimal bakeryDefaults support since Azure doesn't use much there)

@lwander when you originally asked me about rosco vs. clouddriver, I thought you were talking about config gen and I didn't realize Rosco hadn't been implemented yet. AFAIK the account defined here can be used for both clouddriver and rosco config gen, right? The packerResourceGroup and packerStorageAccount _are_ supposed to be at the account level, but are only used in Rosco.yml (although if they're also written to clouddriver.yml, I don't think that does much harm)

EricJizbaMSFT:
Also I'm not saying it's _correct_ that the packerResourceGroup and packerStorageAccount are at the account level, I'm just saying that's how it currently is for Azure. If those should be in bakeryDefaults, I can open up an issue against Rosco, but I still want to get this in since it works with the current implementation

lwander:
Ah yeah I see what you mean - yeah those parameters are defined at the account level in Azure's Rosco settings and are in the right spot in Halyard. I thought you had meant they were scoped to Rosco's global bakery or individual image settings.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/azure/AzureBaseImageValidator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.List;
+
+@Data
+public class AzureBaseImageValidator extends Validator<AzureBaseImage> {
+  final private List<AzureCredentials> credentialsList;
+
+  @Override
+  public void validate(ConfigProblemSetBuilder p, AzureBaseImage n) {
+
+    AzureCredentials credentials = credentialsList.get(0); // The first credentials should be fine since we're validating against public images
```
lwander: Quick question, will the images always be public?

EricJizbaMSFT: Not sure @scotmoor @rguthriemsft ?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/azure/AzureAddAccountCommand.java
```diff
@@ -77,17 +77,31 @@ protected String getProviderName() {
...
+
+  @Parameter(
+      names = "--packer-storage-account",
+      description = AzureCommandProperties.PACKER_STORAGE_ACCOUNT_DESCRIPTION
+  )
+  private String packerStorageAccount;
+
   @Override
   protected Account buildAccount(String accountName) {
     AzureAccount account = (AzureAccount) new AzureAccount().setName(accountName);
```
duftler: Any reason to not just chain all the setter calls together? Looks like this can just be one `return` statement as well.

lwander: It's a lombok issue, `setName()` is defined on `Account`, and the other setters are defined on `AzureAccount`.

duftler: So can we then just surround `(AzureAccount) new AzureAccount().setName(accountName)` with parens?

lwander: Sure, up to you @EricJizbaMSFT

EricJizbaMSFT: Easy enough fix. Also updated AzureAddBaseImageCommand

scotmoor:
LGTM

lwander:
Let's merge this & deal with the issue of possible non-public base images when it becomes a problem.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/azure/AzureBaseImageValidator.java
scotmoor: @lwander they should all be public

PR #151, opened 2017-02-15
Dedup bash completion

lwander:
Now that some commands share names, the bash completion logic needed to
walk the list of commands and recursively match on the NestableCommand
object.

Here's the output https://gist.github.com/lwander/61d934135da10715055b7b13ddc37f56


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -261,6 +252,15 @@ private String commandCompletorCase() {
     bindings.put("subcommands", subcommandNames);
     bindings.put("flags", flagNames);
     bindings.put("command", getCommandName());
+    bindings.put("depth", Integer.toString(depth));
```
duftler: Could use `depth + ""` here.

PR #153, opened 2017-02-16
Install script for Halyard (simplifies debian install path)

lwander:
@duftler PTAL

### DIFF
--- a/InstallHalyard.sh
+++ b/InstallHalyard.sh
@@ -1,0 +1,256 @@
+#!/bin/bash
+
+# This script installs Halyard.
+# See http://www.spinnaker.io/docs/creating-a-spinnaker-instance
+
+
+set -e
+set -o pipefail
+
+REPOSITORY_URL="https://dl.bintray.com/spinnaker/debians"
+
+# We can only currently support limited releases
+# First guess what sort of operating system
+
+if [ -f /etc/lsb-release ]; then
+  . /etc/lsb-release
+  DISTRO=$DISTRIB_ID
+elif [ -f /etc/debian_version ]; then
+  DISTRO=Debian
+  # XXX or Ubuntu
+elif [ -f /etc/redhat-release ]; then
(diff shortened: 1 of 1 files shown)


InstallHalyard.sh
```diff
@@ -0,0 +1,255 @@
...
+function add_apt_repositories() {
+  # Spinnaker
+  # DL Repo goes here
+  REPOSITORY_HOST=$(echo $REPOSITORY_URL | cut -d/ -f3)
+  if [[ "$REPOSITORY_HOST" == "dl.bintray.com" ]]; then
+    REPOSITORY_ORG=$(echo $REPOSITORY_URL | cut -d/ -f4)
+    # Personal repositories might not be signed, so conditionally check.
+    gpg=""
+    gpg=$(curl -s -f "https://bintray.com/user/downloadSubjectPublicKey?username=$REPOSITORY_ORG") || true
+    if [[ ! -z "$gpg" ]]; then
```
jtk54: I think `-n` works as 'nonzero string length'.

InstallHalyard.sh
```diff
@@ -0,0 +1,255 @@
...
+    echo "you must manually install java 8 and then rerun this script; exiting"
+    exit 13
+  fi
+}
+
+function install_halyard() {
+  package="spinnaker-halyard"
+  sudo apt-get install -y --force-yes --allow-unauthenticated $package
+  local apt_status=$?
+  if [[ $apt_status -ne 0 ]]; then
```
jtk54: I'd add some `echo`s to let people know you're going the `dpkg` route and that `apt` install failed.

InstallHalyard.sh
lwander: yup whoops, meant to fix this. thanks

InstallHalyard.sh
lwander: good thinking

PR #156, opened 2017-02-16
Count docker tags rather than ping v2 endpoint

lwander:


judy:
Tested against an AWS ECR registry. I couldn't validate my credentials before this change, and can validate after this change. +1

lwander:
Sweet, is it showing any new warnings? I'm worried it won't pick up on tags in your provided repos even though they've been supplied.

lwander:
The command you posted in slack verifies this. Thanks!

judy:
Nope, no new warnings! Just the expected success output.

PR #161, opened 2017-02-17
Halyard can now deploy an OAuth2-authenticating Spinnaker

lwander:


lwander:
https://github.com/spinnaker/halyard/issues/108

lwander:
pinging @ttomsu

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java
```diff
@@ -41,15 +42,96 @@ public NodeIterator getChildren() {
...
+        newClient.setScope("profile");
+        newClient.setClientAuthenticationScheme("query");
+
+        newResource.setUserInfoUri("https://graph.windows.net/me?api-version=1.6");
+
+        newUserInfoMapping.setEmail("userPrincipalName");
+        newUserInfoMapping.setFirstName("givenName");
+        newUserInfoMapping.setLastName("surname");
+        break;
+      default:
```
ttomsu: Not sure how it would be specified, but one should be able to supply their own OAuth2 endpoint. Would that be supported?

lwander: Yup they have two options:

1. (easy) leave `provider == null` and then set each flag by hand using the CLI/halconfig.
2. (not-as-easy) extend halyard with a new oauth2 provider type.

PR #175, opened 2017-02-23
Fix halyard init entry to run as root

lwander:
This is (unfortunately) needed for halyard to install debians on behalf of the user.

### DIFF
--- a/halyard-web/etc/init/halyard.conf
+++ b/halyard-web/etc/init/halyard.conf
@@ -4,6 +4,4 @@ start on filesystem or runlevel [2345]
 
 expect fork
 
-stop on stopping spinnaker
-
-exec sudo -u spinnaker -g spinnaker /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &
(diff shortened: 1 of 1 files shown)


halyard-web/etc/init/halyard.conf
```diff
@@ -4,6 +4,4 @@ start on filesystem or runlevel [2345]
 
 expect fork
 
-stop on stopping spinnaker
-
-exec sudo -u spinnaker -g spinnaker /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &
+exec sudo /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &
```
jtk54: Why can't we `stop on stopping spinnaker` still?

This essentially requires `sudo` to read the logs, right?

lwander: Because if halyard stops spinnaker, we don't want halyard to stop working.

It requires `sudo` to edit the logs, they're still readable for me.

jtk54: Ok SGTM.

PR #183, opened 2017-02-27
Generate install files with halyard

lwander:
This approach is required since it doesn't make sense to run the daemon as root. If we find this is too cumbersome (write the install file, then have the user execute it) we can just have the CLI execute the command on behalf of the user, so the installation path really just becomes a single command.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/LocalhostDebianDeployment.java
```diff
@@ -61,27 +60,53 @@ public SpinnakerEndpoints getEndpoints() {
...
+
+    DaemonTaskHandler.log("Writing installation file");
+    JarResource installScript = new JarResource("/debian/install.sh");
+    bindings = new HashMap<>();
+    bindings.put("pin-files", pinFiles);
+    bindings.put("spinnaker-artifacts", artifacts);
+    bindings.put("install-redis", "true");
+    bindings.put("install-spinnaker", "true");
+    bindings.put("etc-init", etcInit);
+    bindings.put("packer-version", "0.12.2"); // TODO(lwander) get this from the BOM dependencies
```
jtk54: Is this something worth putting into spinnaker-dependencies?

lwander: I think the best place is in the rosco debian @duftler

PR #193, opened 2017-03-03
bugfix (openstack) only grabbing region when openstack account is enabled

emjburns:
@lwander PTAL - missed this check.

@EricJizbaMSFT this should fix your problem, but if it doesn't ping me!


EricJizbaMSFT:
Awesome - thanks for the quick fix!

PR #201, opened 2017-03-06
Set a deployment's version with Halyard

lwander:
This turned into more work than expected because

1. The package containing logic to lookup versions lived with the deployment code (since that's the only place it was needed). This required moving some of the GCS logic into `halyard-core`.
2. There was no way to load beans in the nodes of the YAML because they were constructed by Jackson either from HTTP requests or based on what was supplied in the halconfig. Supplying our `ProblemSets` with the application context to pass to any validation fixed this. This lets us recommend versions like so: 
versions


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/problem/v1/ConfigProblemSetBuilder.java
```diff
@@ -32,9 +34,16 @@
 public class ConfigProblemSetBuilder {
   private List<ConfigProblemBuilder> builders = new ArrayList<>();
 
+  @Getter
```
jtk54: What happens when you leave off the `AccessLevel` here?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditVersionCommand.java
```diff
@@ -0,0 +1,47 @@
...
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "edit";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Set the desired Spinnaker version.";
+
+  @Parameter(
+      names = "--version",
+      required = true,
+      description = "Must be either a version number \"X.Y.Z\" for a specific release of Spinnaker, \"latest\" for the latest "
```
danielpeach: Drop `for the latest` at the end of the line here.

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
```diff
@@ -12,9 +12,10 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
```
jtk54: On purpose?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentConfiguration.java
```diff
@@ -87,4 +92,9 @@ public NodeIterator getChildren() {
   public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
     v.validate(psBuilder, this);
   }
+
+  protected List<String> versionOptions(ConfigProblemSetBuilder psBuilder) {
+    VersionsService service = psBuilder.getContext().getBean(VersionsService.class);
+    return service.getVersions().getVersions().stream().map(Version::getVersion).collect(Collectors.toList());
```
danielpeach: `getVersions().getVersions()` is a little tough to read. Could the first be `loadVersions`?

lwander: I agree, but then you'd have two methods, one returning a `Versions` object via `get`, and another returning a list of `Versions.Version` objects via `load`, which is confusing.

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
lwander: Likely Intellij during the automatic refactor.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/problem/v1/ConfigProblemSetBuilder.java
lwander: It defaults to public. Everything in the core/config code has public getters & setters, it's only the CLI package where I need to be careful to delineate what to generate.

PR #206, opened 2017-03-07
feat(cli): Inspect running services.

lwander:
This allows users to inspect services they have deployed with Halyard.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/deploy/DetailsDeployCommand.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+
+@Parameters()
+public class DetailsDeployCommand extends NestableCommand {
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "details";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Get details about your currently deployed Spinnaker installation.";
```
jtk54: Is this configuration details specifically?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/ConfigParser.java
```diff
@@ -0,0 +1,83 @@
...
+                  .getMessage()).build()
+      );
+    } finally {
+      if (writer != null) {
+        writer.close();
+      }
+    }
+  }
+
+  public <T> T read(Path path, Class<T> tClass) {
```
jtk54: What does the leading `<T>` do here? Would a base class of the configuration classes you are mapping yaml to accomplish the same result and avoid using a generic type?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/LocalhostDebianDeployment.java
```diff
@@ -40,18 +43,36 @@ public DeploymentType deploymentType() {
...
+      details.setService(service);
     }
 
-    return serviceFactory.createService(endpoint, type);
+    String endpoint = service.getBaseUrl();
+    Object serviceInterface = serviceFactory.createService(endpoint, endpointType);
+    boolean healthy = false;
+
+    try {
+      switch (endpointType) {
```
jtk54: Need a `default` case here?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/ConfigParser.java
lwander: It's saying that this method uses a generic type `T`.

Yup, but it would be a ton of boilerplate.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/deploy/DetailsDeployCommand.java
lwander: It'll encompass health states, logs, etc... as we add more features.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/LocalhostDebianDeployment.java
lwander: good call

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/ConfigParser.java
jtk54: Ok, if you had the base class already I'd say to use it. This is ok as is.

PR #207, opened 2017-03-08
feat(install): Have `hal` execute install script

lwander:


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/job/v1/TeeByteArrayOutputStream.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ */
+
+package com.netflix.spinnaker.halyard.core.job.v1;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class TeeByteArrayOutputStream extends ByteArrayOutputStream {
```
duftler: Are you able to use this instead of writing this from scratch?

https://commons.apache.org/proper/commons-io/javadocs/api-1.4/org/apache/commons/io/output/TeeOutputStream.html

lwander: I wanted to, but it has no "getBytes" method to read what's been written to it. Without that, I can't provide that as the output stream to the job executor, since (if we ever want to standardize on one job executor) we need to support reading what the job wrote to stdout & stderr at any point.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LogLevelConverter.java
```diff
@@ -0,0 +1,35 @@
...
+import ch.qos.logback.classic.Level;
+import com.beust.jcommander.IStringConverter;
+
+public class LogLevelConverter implements IStringConverter<Level> {
+  @Override
+  public Level convert(String value) {
+    Level def = Level.OFF;
+    Level res = Level.toLevel(value, Level.OFF);
+
+    if (!def.levelStr.equalsIgnoreCase(value)) {
```
jtk54: Maybe use the `Level` enum's `valueOf` here.

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/job/v1/TeeByteArrayOutputStream.java
jtk54: +1 Kind of a bummer you have to roll your own.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LogLevelConverter.java
lwander: I would if it was an enum, those `OFF`, `INFO` etc.. are fields in a class: https://logback.qos.ch/apidocs/ch/qos/logback/classic/Level.html

Issue #179, opened 2017-02-23
Generate Igor Config


lwander:
The goal is to configure Jenkins, Travis, Docker triggers, etc...

PR #210, opened 2017-03-08
feat(igor): Generate Igor config

lwander:


lwander:
Closes https://github.com/spinnaker/halyard/issues/179

PR #218, opened 2017-03-13
fix(config): Fix error help messages for bad key format

lwander:
Also prevent users for adding same registry twice.


lwander:
certs-fix

PR #226, opened 2017-03-13
feat(deploy): google flotilla

lwander:
Opening to avoid forgetting about this branch.


lwander:
Closing due to drift in implementation. Will open in new PR.

PR #229, opened 2017-03-14
feat(kubernetes): Red/black spinnaker services during deploy

lwander:
Sorry for the PR size, I wanted to get everything working before submitting.

A quick rundown on the changes:
  * There is now an `OperationFactory` which generates pipelines & tasks for deploying Spinnaker on a per-cloud provider basis.
  * Halyard creates 3 services `spin-clouddriver-bootstrap`, `spin-orca-bootstrap`, and `spin-redis-bootstrap` which are used to externally deploy Spinnaker via red/black of each service. These services are loaded with a different set of profiles so that they only communicate with one another.
  * Previously, one unique secret was created for each smallest common directory between common files that were linked to in halyard. This however meant secrets would persist in your cluster if you changed a secret file's path. Now I place them all in `$OUTPUT/dependencies/<hash-of-directory>-<filename>` allowing me to require only 1 unchanging secret for all files linked to in halyard.
  * I broke each `SpinnakerEndpoints.Service` into its own class, since that winds up making it a lot easier to handle special-case updates of services.

What's missing:
  * Better Orca interfacing. This deployment is now very slow, because each pipeline takes a handful of minutes to complete. That means you're left looking at an unchanging interface for a while, when you could be reading the information Orca provides us with.
  * Manual rollback support. If a red/black fails it'll tell the user and leave the (hopefully still working) old server group around. In the case where a failure is undetected the user needs an option to rollback.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/FlotillaDeployment.java
```diff
@@ -42,26 +46,50 @@ public SpinnakerEndpoints getEndpoints() {
...
-    providerInterface.deployService(deploymentDetails, services.getDeck());
-    providerInterface.deployService(deploymentDetails, services.getEcho());
-    providerInterface.deployService(deploymentDetails, services.getFiat());
-    providerInterface.deployService(deploymentDetails, services.getFront50());
-    providerInterface.deployService(deploymentDetails, services.getGate());
-    providerInterface.deployService(deploymentDetails, services.getIgor());
-    providerInterface.deployService(deploymentDetails, services.getOrca());
-    providerInterface.deployService(deploymentDetails, services.getRosco());
+    providerInterface.deployService(deploymentDetails, orca, services.getEcho());
+    // providerInterface.deployService(deploymentDetails, orca, services.getFiat());
```
duftler: What changed in this PR that this needs to be commented-out now?

lwander: Nothing as far as I can tell. I pinged Travis earlier tonight and we'll look at it tomorrow

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpinnakerProfile.java
```diff
@@ -121,15 +131,37 @@ private ProfileConfig getBaseConfig(DeploymentConfiguration deploymentConfigurat
    * @param node is the node to find dependent files in.
    * @return the list of files required by the node to function.
    */
-  List<String> dependentFiles(Node node) {
+  List<String> processDependentFiles(Node node) {
```
duftler: According to the method comment, this is processing files that are dependencies of this node (that is, files this node is dependent on); not files dependent on this node.

lwander: Good point

PR #238, opened 2017-03-16
feat(deploy): Mount monitoring secrets for kubectl

lwander:
This was going to be part of a larger commit setting up the sidecars, but we need the BOM change to continue publishing releases to actually build the sidecars.


jtk54:
This is dependent on https://github.com/spinnaker/spinnaker/pull/1477, correct?

lwander:
Yes

PR #239, opened 2017-03-16
feat(config): Fiat config generation + CLI commands

lwander:
Only supports google, and doesn't validate yet. I also need to prevent fiat from being deployed if it's not enabled.


lwander:
fiat

ttomsu:
Unrelated to this PR, but mentioning so I don't forget - Halyard should support "Bring Your Own" OAuth. We provided some common ones for convenience, but that should be a fully supported option.

lwander:
It does already :) , you just need to provide a lot of flags.

Issue #224, opened 2017-03-13
Port-forward Kubernetes Deck & Gate Automatically


lwander:
After deploying Spinnaker to Kubernetes, ports to Deck & Gate should be opened if no public load balancer was created.

PR #253, opened 2017-03-17
feat(deploy) auto-connect to k8s

lwander:
I also wound up adding rudimentary instance sizing support.


lwander:
https://github.com/spinnaker/halyard/issues/224

PR #256, opened 2017-03-18
chore(docs): Auto-generated docs & output

lwander:
I'm putting the docs in the `/docs` directory for now, since I don't think we can link into gists using spinnakerio.


lwander:
Don't be afraid of the PR size, it's mainly auto-generated documentation.

PR #258, opened 2017-03-19
feat(halyard) Debian-only install

lwander:
@ttomsu This will let you install all the spinnaker components without configuring them, for the purpose of baking images. You just need to run

`sudo hal deploy run --no-validate --install-only`

after picking a valid version.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/services/v1/DeployService.java
```diff
@@ -75,7 +76,24 @@ public NodeDiff configDiff(String deploymentName) {
     }
   }
 
-  public Deployment.DeployResult deploySpinnaker(String deploymentName) {
+  public RemoteAction installSpinnaker(String deploymentName) {
+    DeploymentConfiguration deploymentConfiguration = deploymentService.getDeploymentConfiguration(deploymentName);
+
```
duftler: Few extraneous blank lines in here.

lwander:
Shoot I merged this before pushing the extraneous line fix; my laptop prompted me with a ssh password and I tabbed away too quickly. I'll put that into the next merged PR

PR #259, opened 2017-03-20
feat(deploy): Configure vault secret storage

lwander:
This does the following:

1. Install the vault binary alongside Halyard.
2. Provides instances a startupscript to pull config into their filesystem.
3. Provides a sample vault + policy setup.

### DIFF
--- a/InstallHalyard.sh
+++ b/InstallHalyard.sh
@@ -3,11 +3,13 @@
 # This script installs Halyard.
 # See http://www.spinnaker.io/docs/creating-a-spinnaker-instance
 
-
 set -e
 set -o pipefail
 
 REPOSITORY_URL="https://dl.bintray.com/spinnaker-team/spinnakerbuild"
+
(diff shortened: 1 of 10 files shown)


InstallHalyard.sh
```diff
@@ -258,12 +260,21 @@ if [ -n "$DOWNLOAD"]; then
   add_apt_repositories
 fi
 
-TEMPDIR=$(mktemp -d installhalyard.XXXX)
-
 echo "$(tput bold)Installing Java 8...$(tput sgr0)"
 
 install_java
 
+apt-get install unzip
```
duftler: Do you want a `-y` here?

duftler: If not, what happens if the user responds with `n`?

lwander: Oops, you really want `-y`. If the user responds with `n` the next call to unzip will fail, and since `-e` is set global to the script, the installation will return 1.

PR #265, opened 2017-03-21
feat(deploy): Deploy Kubernetes services w/ public IPs

lwander:
If you provide a correct ui & api + domain & address alongside configured authn methods this'll deploy spinnaker on k8s as a protected, publicly facing service.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerPublicService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerPublicService.java
@@ -25,18 +25,25 @@ import lombok.EqualsAndHashCode;
 @Data
 abstract public class SpinnakerPublicService<T> extends SpinnakerService<T> {
   String protocol = "http";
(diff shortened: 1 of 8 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerPublicService.java
```diff
@@ -25,18 +25,25 @@
 @Data
 abstract public class SpinnakerPublicService<T> extends SpinnakerService<T> {
   String protocol = "http";
+  String domain;
 
   public abstract String getPublicAddress();
 
   public String getPublicEndpoint() {
-    return getProtocol() + "://" + getPublicAddress() + ":" + getPort();
+    return getProtocol() + "://" + domain + ":" + getPort();
```
duftler: At some point, would like to get away from doing all of this string-work manually.

lwander: https://github.com/spinnaker/halyard/issues/267

PR #266, opened 2017-03-21
chore(docs): Add doc describing Halyard deploys

lwander:
This points out the overall strategy and some of the important things to watch out for during the deployment process.


docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+* **LocalhostGit**: This is every Spinnaker service installed from Git on the 
+ node Halyard is running on. This is intended for use by Developers.
+* **Flotilla**: This is every Spinnaker service installed remotely on a
+ specified Cloud Provider as separate, independently scalable instance. There
+ is no customization or breakout of specialized profiles for this deployment.
+
+Every deployment type must address a few issues:
+
+1. **Configuration**: How Spinnaker recieves the config it needs to run.
+2. **Service Dependency Setup**: What services Spinnaker must install to get
```
skim1420: s/Spinnaker/Halyard

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+2. **Service Dependency Setup**: The only service dependency is Redis, and it
+ is installed locally.
+3. **Deployment & Updates**: All spinnaker packages are pinned by writing
+ `apt` preferences in `/etc/apt/preferences.d/pin-spin-{artifact}` for each
+ Spinnaker artifact. During the initial install and subsequent updates, this
+ file is always updated before attempting to grab new packages via `apt`.
+
+### LocalhostGit Deployments
+
+> Timeline: Mid-to-late April
```
skim1420: We should talk through what problems are we trying to solve, etc. before any work is put into this. For design considerations today that may be affected by what we decide to do in support of this deployment type, let's talk through those first.

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+ /secret/spinnaker/clouddriver-{id}/config1: {"path": "/a/b", "contents" "xyz"}
+ /secret/spinnaker/clouddriver-{id}/config2: {"path": "/z/y", "contents" "cba"}
+ ```
+ The instance will be provided with top-level secret id `clouddriver-{id}` and
+ use that to download the secrets `config1` & `config1`, and place their
+ `contents` into `path`.  Service endpoints will be configured using typical Consul DNS names, which
+ will be written during config generation. Both Vault & Consul endpoints will
+ be distributed as instance metadata.
+2. **Service Dependency Setup**: Redis, Consul, & Vault will all be deployed if
+ not supplied. A first-pass will configure a small Consul cluster and
```
skim1420: Any concerns of conflict or incompatibility in working with existing Consul?

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+ of the instance will download and place in their intended directories. The
+ secret format looks roughly like this:
+ ```
+ /secret/spinnaker/clouddriver-{id}: {"configs": ["config1, "config2"] }
+ /secret/spinnaker/clouddriver-{id}/config1: {"path": "/a/b", "contents" "xyz"}
+ /secret/spinnaker/clouddriver-{id}/config2: {"path": "/z/y", "contents" "cba"}
+ ```
+ The instance will be provided with top-level secret id `clouddriver-{id}` and
+ use that to download the secrets `config1` & `config1`, and place their
+ `contents` into `path`.  Service endpoints will be configured using typical Consul DNS names, which
```
skim1420: What about foregoing Consul and working with GCP ILBs directly? 

Relatedly, what are your plans on exposing gate & deck to external IPs?

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+* **Deployment Environment**: The infrastructure Spinnaker is running on,
+ Spinnaker's Deployment Type, and Spinnaker's service dependencies.
+
+## Deployment Types
+
+* **LocalhostDebian**: This is every Spinnaker service installed as a Debian
+ package on the node Halyard is running on.
+* **LocalhostGit**: This is every Spinnaker service installed from Git on the 
+ node Halyard is running on. This is intended for use by Developers.
+* **Flotilla**: This is every Spinnaker service installed remotely on a
```
ewiseblatt: Is this a new term or one in use? It's cute, but we should agree on the term.

Flotilla is cute, but more standard would be something like "standalone" and "distributed" (StandaloneDebian, StandaloneGit) and Distributed.

If a flotilla adds a spinnaker twist to distributed (e.g. flotillas are deployments managed by halyard) then ok. Otherwise be wary about introducing new concepts.

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+ this deployment to run.
+3. **Deployment & Updates**: How the Spinnaker services are deployed and
+ updated in this environment.
+
+### LocalhostDebian Deployments
+
+> Timeline: Already included in Halyard.
+
+Halyard handles this deployment type by generating an install script very
+similar to `InstallSpinnaker.sh` in spinnaker/spinnaker repository. The CLI
```
ewiseblatt: Does InstallSpinnaker.sh still exist in the future? If so, ok. If not then dont reference it since it will be undefined.

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+receieves this script, and attempts to run it locally. 
+
+1. **Configuration**: The script generated by Halyard has commands to copy all
+ configuration from the necesssary Spinnaker config already generated by 
+ Halyard into `/opt/spinnaker/config`, `/opt/spinnaker-monitoring` and
```
ewiseblatt: what about upstart in /etc/config?
It seems unfortunate that /opt/spinnaker-monitoring is an outlier here.
Should /opt/spinnaker-monitoring install its config to /opt/spinnaker/config?
Is clouddriver, etc still in /opt/clouddriver? Those are referencing into /opt/spinnaker/config. Can we rethink this to simplify (at least consistent)?

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+
+Halyard handles this deployment type by generating an install script very
+similar to `InstallSpinnaker.sh` in spinnaker/spinnaker repository. The CLI
+receieves this script, and attempts to run it locally. 
+
+1. **Configuration**: The script generated by Halyard has commands to copy all
+ configuration from the necesssary Spinnaker config already generated by 
+ Halyard into `/opt/spinnaker/config`, `/opt/spinnaker-monitoring` and 
+ `/etc/apache2`. Any required files are already in the correct location.
+2. **Service Dependency Setup**: The only service dependency is Redis, and it
```
ewiseblatt: If you dont need to be specific about Redis being the only one than dont. Think in terms of policy here. If someone added a new dependency, how should it behave? Provide the guidance here by speaking in terms of policies so that the person introducing something new in the future doesnt need to think about it since it is already defined.

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+
+Halyard handles this deployment by installing all Spinnaker service git repos
+in a user-configurable directory, with a user-configurable `upstream` and
+`origin`.  There will be a special version called `HEAD` that supersedes any 
+configuration in our profile registry that will take all services at git `HEAD` 
+to use for config generation.
+
+### Flotilla Deployments
+
+> Timeline: Varies by cloud provider. See below.
```
ewiseblatt: The tone of this section has changed from your original purpose of explaining how it is done to a discussion about the issues

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+ secrets, and describes to the individual services where to mount them.
+2. **Service Dependency Setup**: The only external dependency is redis, and if
+ it is not provided, Spinnaker will deploy a Redis instance.
+3. **Deployent & Updates**: See above.
+
+#### GCE
+
+> Timeline: Middle April.
+
+1. **Configuration**: Halyard uses the configured Vault storage and access
```
ewiseblatt: Perhaps link to Value

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+> detailed in https://github.com/spinnaker/halyard/projects/1
+
+1. **Configuration**: Halyard publishes all configuration as Kubernetes
+ secrets, and describes to the individual services where to mount them.
+2. **Service Dependency Setup**: The only external dependency is redis, and if
+ it is not provided, Spinnaker will deploy a Redis instance.
+3. **Deployent & Updates**: See above.
+
+#### GCE
+
```
ewiseblatt: what about logging and monitoring on these nodes (not just GCE)?

docs/deploys.md
lwander: The problem with the ILBs is that they don't support internal DNS. Until they do, any changes to your endpoint configuration will require redeploying all services to get new endpoint info.

docs/deploys.md
lwander: No. If the clusters never find out about one-another nothing changes, and if they do, they treat one-another as separate consul "datacenters" https://www.consul.io/docs/guides/datacenters.html

docs/deploys.md
lwander: Ah I forgot to mention `/etc/config`, but yes, those get generated as well for whatever artifacts you're deploying on that node and are bundled into another `spinnaker` upstart entry.

It would be nice if spinnaker-monitoring adopted some of the other Spinnaker conventions. For example, a name that isn't prefixed with "spinnaker-", and using the shared config directory would simplify a lot of special-case handling.

docs/deploys.md
lwander: No reason to assume it will, good point.

docs/deploys.md
lwander: Right, it's not mentioned but I'm assuming the agents are installed on the nodes as a part of our bake process, and it's just a matter of providing them with config.

docs/deploys.md
lwander: You're right, I see that "distributed" makes sense over "flotilla". But I don't think "standalone" correctly conveys the sentiment that Spinnaker is being installed locally. I think just "local" makes more sense, and doesn't conflate with loopback. So we'd have

LocalDebian, LocalGit, and Distributed

lwander: WDYT?

docs/deploys.md
lwander: I'm curious, what kind of design considerations do you have in mind beyond providing a starting-point for developers?

docs/deploys.md
lwander: That's true, let me see how I can reword

docs/deploys.md
ewiseblatt: cameron created the repo as spinnaker-monitoring. I thought monitoring might be too generic for a package name and cause conflict. Within spinnaker I can use "monitoring" but global namespace not so much. Maybe another name?

docs/deploys.md
skim1420: sgtm

docs/deploys.md
skim1420: I'm not saying I have design considerations. I'm saying we should ignore this use case for now, unless there are things we need to consider now that might affect later when we do take on this use case.

lwander: Oh I see what you mean - I was thinking this would be a nice-to-have, since we have someone joining in the next month, and as far as I can tell this is a minimal amount of work that goes a long way to integrate Halyard into our toolchains.

skim1420:
SGTM for my comments.

ewiseblatt:
LGTM. I didnt see you update about the wording you mentioned (tone of document changing) but that's your call.

lwander:
@ewiseblatt I haven't updated the doc yet, and I agree the tone needs to change around that section.

docs/deploys.md
lwander: Up to you :)

docs/deploys.md
lwander: Good point, Reddis isn't a special case and I shouldn't handle it as one. Let me see what I can describe for generic service dependencies.

docs/deploys.md
```diff
@@ -0,0 +1,120 @@
...
+   when instances from both version with incompatable caches schemas will 
+   share the same Redis cache, flushing the cache is not an option. The only 
+   alternative is to scale down the old version to 0 instances, and the flush
+   the redis cache where applicable. The critical thing to note here is that
+   the only way Halyard can know to do that is when a services Major version
+   increases.
+
+#### Kubernetes 
+
+> Timeline: End of March. The core features are in, but a few touchups as
```
duftler: Should this still read "End of March"?

lwander: No, sorry, I re-prioritized once we shifted the release date.

PR #272, opened 2017-03-22
feat(deploy): Improve logging granularity from Orca

lwander:
error
Now it shows why the pipeline failed, and what stages it's running.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/provider/v1/OrcaRunner.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/provider/v1/OrcaRunner.java
@@ -1,0 +1,284 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/provider/v1/OrcaRunner.java
```diff
@@ -0,0 +1,280 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+@Component
+class OrcaRunner {
+  @Autowired
+  private ObjectMapper objectMapper;
+
+  void monitorTask(Supplier<String> submitTask, Orca orca) {
+    final String id = submitTask.get().substring(1);
```
jtk54: What char is at the beginning of this string? Can you add a comment why you `substring` here?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/provider/v1/OrcaRunner.java
```diff
@@ -0,0 +1,280 @@
...
+
+    Supplier<Pipeline> getPipeline = () -> {
+      Map<String, Object> execution = orca.getRef(id);
+      return objectMapper.convertValue(execution, Task.class).getExecution();
+    };
+    monitor(getPipeline, orca);
+  }
+
+  void monitorPipeline(Supplier<String> submitPipeline, Orca orca) {
+    final String id = submitPipeline.get().substring(1);
```
jtk54: Better yet, maybe add a function that is a `stripXChar` since you do this more than once to the id strings.

lwander: Good idea, the char is a leading '/' that is returned by orca as a reference to the pipeline.

PR #274, opened 2017-03-22
fix(deploy): Non-public service Load Balancer

lwander:


lwander:
Let's run those integration tests again :)

PR #276, opened 2017-03-22
fix(deploy): Show version in running service details

lwander:


lwander:
Not yet, I want to get more of a feel for how running services should be represented/what's helpful before adding too much else.

PR #283, opened 2017-03-26
refactor(config): Rethink config mapping

lwander:
Really sorry about the size, I didn't expect this refactor to carry on for so long. Don't worry glossing over the details here, I realize it's pretty unreadable at this length.

This originated from a change in the way I was generated profiles (configs) for services. Before this, a "profile" could contain multiple files, but only one of those could be backed by an external config found on GCS. This was a hack to quickly add support for custom apache2 config alongside settings.js, but was ugly and was preventing me from properly supporting multiple Rosco templates, and downloading monitoring dashboards for deployment via Halyard. To address this, I made each type of profile backed by a ProfileFactory, which can get a base profile from any number of sources (GCS, string, Jar resource, etc...). 

Once this was completed, I realized it would be nicer from a deployment standpoint to have each SpinnakerService describe exactly what profile factories it needed to configure itself, as well as be able to generate endpoint information given a halconfig. Before this refactor, SpinnakerServices lived outside of the Spring lifecycle, so it would be a huge pain/ugly use of ApplicationContext to instantiate those ProfileFactories into each SpinnakerService. So, I turned each SpinnakerService into a Spring Component making this much neater. As a result, it became really easy for a particular type of Deployment to just enumerate which SpinnakerServices it needed, and have each SpinnakerService be in charge of generating its own installation commands & deployment/update pipelines. 

That's what this refactor achieves, again, sorry it got out of hand.


jtk54:
Extremely cursory look. Taking this on

image

PR #289, opened 2017-03-29
fix(config): PrimaryAccount behavior

lwander:
Prior to this fix PrimaryAccount would not be correct if the account it referred to was removed.

@duftler this removes the TODO you added during rosco validation.


lwander:
ping

PR #306, opened 2017-04-03
docs(password): Add details about password prompts on stdin.

jtk54:


docs/commands.md
```diff
@@ -635,7 +635,7 @@ hal config webhook jenkins master add MASTER [parameters]
  * `--address`: (*Required*) The address your jenkins master is reachable at.
  * `--no-validate`: (*Default*: `false`) Skip validation.
  * `--options`: Get options for the specified field name.
- * `--password`: The password of the jenkins user to authenticate as.
+ * `--password`: (Sensitive data - user will be prompted on standard input.) The password of the jenkins user to authenticate as.
```
lwander: The period inside of the parens feels wrong.

lwander: Could you also make "Sensitive data" italicized to match the other arg decorators? 

(*Sensitive data* - user will be prompted on standard input)

jtk54: Yep, agree with both.

PR #313, opened 2017-04-04
fix(cli): Cleanup output during deploy

lwander:
Previously the daemon logger was producing a lot of noise during the deploy stage related to pipeline task executions. Moving this into a new `details` field allows for that bottom white bar to update whenever a sort-of sub-task completes. This looks a lot neater on the CLI, and cleans up the internal data model a bit.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/ResponseUnwrapper.java
```diff
@@ -16,16 +16,14 @@
 
 package com.netflix.spinnaker.halyard.cli.services.v1;
 
+import com.amazonaws.util.StringUtils;
```
duftler: In `AnsiParagraphBuilder`, you use `org.apache.commons.lang.StringUtils`.

PR #323, opened 2017-04-06
feat(deploy): Install vault using BOM version

lwander:
Everything is now in place for building GCE service-specific VM images.

You'll need to run the following:

```
hal deploy actuate --omit-config --no-validate --service-names $COMPONENT monitoring-daemon vault-client consul-client
```

As well as copy the directory `./services/vault` into `/opt/spinnaker/`, since we'll be calling `/opt/spinnaker/startup/mount-config.sh` as the startup script.


jtk54:
What order does the copy of `./services/vault` and running `actuate`?

lwander:
doesn't matter, both steps just stage stuff for when spinnaker starts later

PR #331, opened 2017-04-10
feat(cli): s/actuate/apply

lwander:


lwander:
ok ok, back to apply

duftler:
Was the plan that I wouldn't notice that you changed only the title?

Issue #280, opened 2017-03-26
Permission denied for k8s config


idoshamun:
Running `hal config provider kubernetes account add my-k8s-account --docker-registries my-gcr-account --kubeconfig-file ~/.kube/config` command, throws the following error:

```
Problems in default.provider.kubernetes.my-k8s-account:
! ERROR Cannot find provided path: /home/ido/.kube/config
  (Permission denied).
- Failed to add account my-k8s-account for provider kubernetes.
```

I didn't change the default permissions of k8s config file, of course.
`ls -l ~/.kube`
```
total 8
drwxr-xr-x 3 ido ido 4096 Mar 25 20:25 cache
-rw------- 1 ido ido 2189 Mar 25 23:08 config
```

idoshamun:
`chmod 744 ~/.kube/config` fixed this issue

lwander:
Thanks for reporting this, I've been meaning to have Halyard explain that the files need to be readable by "spinnaker".

PR #337, opened 2017-04-11
fix(daemon): Add no access remediation to FileNotFoundException

lwander:
Turns out some systems will put the permission denied error in the `FileNotFoundException`, while others put it in the `IOException`.


lwander:
Closes https://github.com/spinnaker/halyard/issues/280

PR #343, opened 2017-04-12
fix(install): non .conf files are omitted by nebula

lwander:


lwander:
I'm merging this quickly since halyard install is broken until this is in

PR #349, opened 2017-04-13
feat(cli): Print artifact-specific version

lwander:


lwander:
@jtk54 you'll want to supply `--color=false --quiet` as well

PR #353, opened 2017-04-18
feat(bake): Bake top-level startupscript into VM images

lwander:
This ensures services are started in order of priority (e.g. vault, then consul, then component). No changes needed on your part @jtk54


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/BakeDeployer.java
```diff
@@ -55,7 +57,13 @@ public RemoteAction deploy(
       return m1;
     });
 
-    String installCommand = serviceProvider.getInstallCommand(resolvedConfiguration, installCommands);
+    String startupCommand = String.join("\n", enabledServices.stream()
+        .map(BakeService::getStartupCommand)
+        .filter(Objects::nonNull)
+        .map(es -> es)
```
jtk54: This map is just doing an identity function.

lwander: oh whoops, it used to do a little more

PR #355, opened 2017-04-19
feat(bake): Bake vault-server, consul-server &redis

lwander:
You'll have to run the following commands to build the three external service dependencies respectively

```
hal deploy apply --no-validate --service-names redis
hal deploy apply --no-validate --service-names consul-server
hal deploy apply --no-validate --service-names vault-server
```

keep in mind, that since the image name will be static (we don't append build numbers to these) that you'll have to delete old images if they exist)


lwander:
There is also no need to include the other components like for the core services, i.e. no `monitoring-daemon consul-client vault-client` should be baked on these images.

Issue #119, opened 2017-02-05
Support Azure configuration


lwander:
The linked PRs & details in #71 will be helpful.

lwander:
@EricJizbaMSFT ready to close this?

EricJizbaMSFT:
I've got two things left to do:
1. Add Azure to setting.js
2. Add Azure Storage to front50

lwander:
Hey, when you get a chance could you also make sure the Azure OAuth provider works with `hal config security oauth2 edit --provider azure --client-id <> --client-secret <>`? I've implemented all the Azure bits, I just don't have any way to verify.

EricJizbaMSFT:
FYI I haven't forgotten about this - I've just been occupied with other tasks. I should be able to try out the OAuth provider next week and hope to get to the other items soon

Speaking of... @lwander any thoughts on how I should do the azure storage config? Both aws and gce use the 'bucket' terminology, but Azure does not. I could either add the Azure parameters to the current storage command (a little more confusing for users IMO) or we could abstract out the storage command for all providers. I'm leaning the second option, but it's obviously a bigger change so I thought I'd run it by you first

lwander:
No worries. I'd prefer if the tool felt native to users of each cloud provider, rather than tried to fit everything into the same abstraction - so trying to adopt the Azure terminology where appropriate feels best. What's the change in naming you're proposing?

EricJizbaMSFT:
For gcs or s3 it would be the exact same except you have to specify the provider before 'edit':
`hal config storage gcs edit --account-name my-google-account --bucket bucket123 --location westus --root-folder spinnaker`

For azs, I would do something like this:
`hal config storage azs edit --storage-account-name storageAccount123 --storage-account-key secretkey --container-name spinnaker`
> In Azure, an 'account' is analogous to a 'bucket'. So the 'storage-account-name' has nothing to do with a spinnaker account. In fact, azure just uses a key to authenticate and has no use for spinnaker accounts or location like gcs/s3.

Finally, since Azure doesn't use a spinnaker account, we can't automatically detect which provider to enable. I think we would want to add a command like this:
`hal config storage edit --provider <gcs|azs|s3|cassandra>`
> Not sure if this should be 'edit', 'set', 'change' or something else
Alternatively, we could add a "enable/disable" command under each provider, but I don't think that makes sense since only one can be enabled at a time.

lwander:
[...]

PR #359, opened 2017-04-21
feat(storage): Split persistentStorage into subclasses based on persistent store

EricJizbaMSFT:
This allows for unique parameter names and validation per store (needed for upcoming azs support #119)
Also added "hal config storage set" command to set the current store


halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/PersistentStorageController.java
```diff
@@ -46,8 +47,8 @@
 
   @RequestMapping(value = "/", method = RequestMethod.GET)
   DaemonTask<Halconfig, PersistentStorage> getPersistentStorage(@PathVariable String deploymentName,
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Severity severity) {
+                                                                @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
```
lwander: Small nit, could you preserve the existing indentation for multiline parameter definitions?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/PersistentStorage.java
```diff
@@ -1,22 +1,25 @@
...
+import java.util.Optional;
 
-/**
- * This is the configuration for S3/GCS storage options.
- */
 @Data
 @EqualsAndHashCode(callSuper = false)
-@Slf4j
 public class PersistentStorage extends Node {
+  CassandraPersistentStore cassandra = new CassandraPersistentStore() {{
```
lwander: We shouldn't include cassandra is it's formally deprecated.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/persistentStorage/SetPersistentStoreCommand.java
```diff
@@ -0,0 +1,52 @@
...
+@Parameters(separators = "=")
+public class SetPersistentStoreCommand extends AbstractConfigCommand {
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "set";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Set Spinnaker's current persistent store.";
+
+  @Parameter(
+      names = "--name",
```
lwander: I don't think `name` is correct, since everywhere else, `name` is completely user-definable. Instead `type` or `source` implies that there is a predefined list of storage options to chose from.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/PersistentStorageController.java
```diff
@@ -82,6 +83,49 @@
     builder.setRevert(() -> halconfigParser.undoChanges());
     builder.setSave(() -> halconfigParser.saveConfig());
 
-    return DaemonTaskHandler.submitTask(builder::build, "Edit storage settings");
+    return DaemonTaskHandler.submitTask(builder::build, "Set current persistent store");
+  }
+
+  @RequestMapping(value = "/persistentStore/{persistentStoreName:.+}", method = RequestMethod.GET)
+  DaemonTask<Halconfig, PersistentStore> getPersistentStore(@PathVariable String deploymentName,
```
lwander: How do you pick a persistent store to use? Do you have enable & disable explicitly?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/PersistentStorageValidator.java
```diff
@@ -0,0 +1,49 @@
...
+      if (persistentStore.enabled) {
+        numEnabled++;
+      }
+
+      child = nodeIterator.getNext();
+    }
+
+    if (numEnabled == 0) {
+      ps.addProblem(Severity.FATAL, "No persistent store is enabled.")
+          .setRemediation("Enable a persistent store.");
```
lwander: I don't think these remediations add much beyond what the problem message is telling the user.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/PersistentStorageController.java
```diff
@@ -57,19 +58,19 @@
...
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Severity severity,
-      @RequestBody Object rawPersistentStorage) {
-    PersistentStorage persistentStorage = objectMapper.convertValue(rawPersistentStorage, PersistentStorage.class);
+  @RequestMapping(value = "/currentPersistentStore", method = RequestMethod.PUT)
+  DaemonTask<Halconfig, Void> setCurrentPersistentStore(@PathVariable String deploymentName,
+                                                        @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
+                                                        @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Severity severity,
+                                                        @RequestBody Object rawPersistentStoreName) {
+    String persistentStoreName = objectMapper.convertValue(rawPersistentStoreName, String.class);
```
lwander: This endpoint breaks convention, usually you would pick a storage type by editing the top-level persistent storage object containing all storage types. This would be surfaced in halyard has `hal config storage edit --type azs`.

Take a look at the metric store controller to see how it's handled there.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/PersistentStorage.java
EricJizbaMSFT: I only included it so that we could have a default value. Otherwise your config will start out in a 'broken' state and you'll have to fix it before validation passes. Is there some better default or are you fine with that?

lwander: Maybe it's best to provide a `WARNING` when no source has been selected yet, and mention that your deployment will most likely fail until one is picked & configured.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/PersistentStorageController.java
EricJizbaMSFT: It can only be set through through the '/currentPersistentStore' (which I will refactor based on your above comment). I didn't want 'enable/disable' per store because you can only set one at a time. However, when you set 'currentPersistentStore, it gets mapped to 'enabled: true/false' for each persistent store here: https://github.com/spinnaker/halyard/pull/359/commits/3b707e0f1ac99e5ada39753d09e2aa567596b1d6#diff-b362be0f61df2de7c9543613c182b8edR111

Which brings up a larger question I have: how _exactly_ is the halconfig used? I'm only familiar with the current model in front50.yml where each persistent store has 'enabled: true/false'. If front50 switches to use the halconfig as the one source of truth (I assume that's the plan?) can front50 config be refactored to use a 'type: gcs' setting under the main persistentStorage node rather than 'enable/disable' under each child node?

lwander: It's used to generate the config for the other services -- so it would be nice to avoid the `enable/disable` flags in the halconfig, and then only have them show when the front50.yml is written using the corresponding profile factory. In fact, I'd recommend removing those flags from the halconfig since they'll likely only cause confusion (but I understand why you've added them).

lwander: Easiest way to ensure `enable/disable` only shows up in the generated config is probably to cast the whole storage object into a `Map` using the `@Autowired ObjectMapper objectMapper`, and then write `.put("enabled", val)` for the necessary providers. Or you could subclass that storage object and then copy the fields over, with `enabled` being the only addition to that subclass.

EricJizbaMSFT: I thought you were removing the `hal config generate` command? Does the config get generated some other way?

lwander: It's always generated & applied during `hal deploy apply`, but `hal config generate` is misleading since it doesn't apply your generated config.

lwander: I was thinking of changing the name/making it clear that it's more valuable for debugging than anything

EricJizbaMSFT:
@lwander let me know what you think and I'll rebase/squash

As an aside: I've always wondered why the PR forms recommend squashing as you go...am I the only one that likes to see the progression of commits in a PR (and squashing at the end)?

lwander:
You know I totally agree actually, the incremental commits make much more sense. Also, using just the "squash & merge" in the web-UI you don't even need to manually rebase using the git CLI, which is another point in favor of just pushing your work in small, consumable commits. I think in the past, github wasn't as fully-featured with respect to multi-commit diffs, and they sort of confused the review process, which is probably why it's still a habit to continually rebase & squash during a review.

EricJizbaMSFT:
@lwander Done

PR #364, opened 2017-04-24
feat(deploy): Add default Azure settings to deck

EricJizbaMSFT:
@lwander I created the corresponding commit in deck: https://github.com/spinnaker/deck/pull/3581, but do I need to do anything else in order for halyard to consume that after it's merged? Do I need to update spinnaker dependencies or something?

#119


lwander:
Nope, we just need to run the build to get the settings.js halconfig changes you made into our halconfig bucket

PR #366, opened 2017-04-24
fix(deploy): Incremental vault init fixups

lwander:
This corrects shebang headers being ignored by prepended edit warnings, broken secret loading for secrets in directories that don't exist yet, and vault/consul client provider-specific startup being ignored by a non-escaped bash sequence.

Also adds initial `spin-clouddriver-bootstrap` deployment, but is missing profile specialization and consul server members to join.


lwander:
Merging early to bake VM images in the nightly.

PR #377, opened 2017-04-26
chore(autocomplete): Adds small Make task for non-installed autocomplete

ttomsu:
@lwander  - You were right that you can't modify the parent shell from within Make, so I just printed a command the user can copy/paste to enable autocomplete during development.

### DIFF
--- a/halyard-cli/.gitignore
+++ b/halyard-cli/.gitignore
@@ -1,0 +1,1 @@
+complete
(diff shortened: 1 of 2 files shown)


halyard-cli/.gitignore
```diff
@@ -0,0 +1 @@
+complete
```
lwander: should this be `halyard-cli/complete`? since `make` is run in that directory

ttomsu: This file is in the `halyard-cli` directory

lwander: Ahh ok sounds good

PR #380, opened 2017-04-27
chore(install): Systemd startupfile

lwander:
@jtk54 FYI

### DIFF
--- a/halyard-web/lib/systemd/system/halyard.service
+++ b/halyard-web/lib/systemd/system/halyard.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=A tool for managing the lifecycle of a Spinnaker installation
+
+[Service]
+Restart=always
+RestartSec=10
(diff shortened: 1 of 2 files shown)


halyard-web/lib/systemd/system/halyard.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=A tool for managing the lifecycle of a Spinnaker installation
+
+[Service]
+Restart=always
+RestartSec=10
+Type=forking
+ExecStart=
```
lwander: This isn't a mistake, it clears this var since it can be assigned multiple times.

jtk54: Comment that.

lwander: Seems to be too basic to comment on when reading these configs.

PR #385, opened 2017-04-27
fix(storage): Stronger GCS validation

lwander:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/PersistentStorageService.java
```diff
@@ -103,9 +103,11 @@ public void setPersistentStore(String deploymentName, PersistentStore newPersist
   }
 
   public ProblemSet validatePersistentStorage(String deploymentName) {
+    PersistentStorage storage = getPersistentStorage(deploymentName);
     NodeFilter filter = new NodeFilter()
         .setDeployment(deploymentName)
-        .setPersistentStorage();
+        .setPersistentStorage()
```
jtk54: Is this just enabling storage?

lwander: It's enabling the validation of whatever storage source you picked.

Issue #305, opened 2017-04-03
Switch to long & short descriptions for each CLI command


lwander:
Since halyard is auto-generating its own documentation now, I want to make the documentation more useful/readable. This involves writing a "long" description for each command that gives a detailed explanation of the resource it's modifying, and what the command does, and a "short" description that's no longer than a sentence to help identify the command's purpose.

The "long" description will sit under the command's header in the docs, [like so](https://github.com/lwander/halyard/blob/7b73b33a1696535c72edd513b801d6cc494320fb/docs/commands.md#hal-config-provider-kubernetes), while the "short" description will sit next to a command in its list of subcommands, [like so](https://github.com/lwander/halyard/blob/7b73b33a1696535c72edd513b801d6cc494320fb/docs/commands.md#subcommands-17).

lwander:
https://github.com/spinnaker/halyard/pull/304

lwander:
Ideally each command's resident expert will supply documentation.

You can rebuild the docs using `make` in the `./halyard-cli/` directory. Please check in the up-to-date docs.

lwander:
Paging

@danielpeach for appengine
@emjburns for openstack
@duftler for gce
@EricJizbaMSFT for azure
@lwander for k8s
@ttomsu for authn/authz

There's no rush here since the default descriptions will keep getting used. But, if you find the time, this is a good place to preemptively answer questions about your providers/expertise before they percolate into slack.

spinnaker-release:
Migrated to https://github.com/spinnaker/spinnaker/issue#2372

PR #386, opened 2017-04-27
feat(authn): Adds SAML configuration support

ttomsu:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/saml/EditSamlCommand.java
```diff
@@ -0,0 +1,96 @@
...
+
+@Parameters(separators = "=")
+public class EditSamlCommand extends AbstractEditAuthnMethodCommand<Saml> {
+
+  @Getter
+  private AuthnMethod.Method method = AuthnMethod.Method.SAML;
+
+  @Parameter(
+      names = "--metadata",
+      description = "The address to your identity provider's metadata XML file. This can be a URL" +
```
lwander: This needs to be prefixed with `http` or `file`, right? If so, can you call that out here?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/saml/EditSamlCommand.java
```diff
@@ -0,0 +1,96 @@
...
+import com.netflix.spinnaker.halyard.config.model.v1.security.AuthnMethod;
+import com.netflix.spinnaker.halyard.config.model.v1.security.Saml;
+import lombok.Getter;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+@Parameters(separators = "=")
+public class EditSamlCommand extends AbstractEditAuthnMethodCommand<Saml> {
+
```
lwander: Can you add a `String longDescription = ...` and `String shortDescription = ...` that describes saml in a little more detail?

https://github.com/spinnaker/halyard/issues/305

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/saml/EditSamlCommand.java
ttomsu: `http` is needed for a URL, but for a local file, only a path is needed, and the `file:` is prepended later.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/saml/EditSamlCommand.java
ttomsu: Added.

PR #391, opened 2017-04-28
feat(core): task commands for interruption

lwander:
There seems to be some resource leakage going on in Halyard. This allows us to troubleshoot.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
@@ -28,6 +28,7 @@ import com.netflix.spinnaker.halyard.cli.ui.v1.*;
 import com.netflix.spinnaker.halyard.core.job.v1.JobExecutor;
 import com.netflix.spinnaker.halyard.core.job.v1.JobExecutorLocal;
 import com.netflix.spinnaker.halyard.core.resource.v1.JarResource;
+import com.netflix.spinnaker.halyard.core.tasks.v1.DaemonTaskInterrupted;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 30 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -173,6 +174,9 @@ private void safeExecuteThis() {
       AnsiUi.error(e.getMessage());
       AnsiUi.remediation("Try the command again with the --debug flag.");
       System.exit(1);
+    } catch (DaemonTaskInterrupted e) {
+      AnsiUi.warning("Interrupted. Task killed in unknown state.");
```
jtk54: Why is the state of the task unknown here? Is there a way to know it?

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/tasks/v1/TaskRepository.java
```diff
@@ -20,83 +19,98 @@
...
       } catch (HalException e) {
         log.info("Task " + task + " failed for reason: ", e);
-        task.setResponse(new DaemonResponse<>(null, new ProblemSet(e.getProblems())));
-        task.setFatalError(e);
-        task.setState(State.FATAL);
+        response = new DaemonResponse<>(null, new ProblemSet(e.getProblems()));
+        fatalError = e;
+        state = State.FATAL;
+      } catch (DaemonTaskInterrupted e) {
+        log.info("Task " + task + " interrupted: ", e);
```
jtk54: This and the previous catch block look quite similar, maybe capture it in a function.

lwander: Both just set a few variables, there isn't much to extract that doesn't require global vars or references

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
lwander: This happens when a different user kills your task. Since that can happen at anytime, there is no good way to know what just happened (beyond looking at the task output on your screen).

PR #400, opened 2017-05-01
fix(docker): Don't warn on whitespace errors when not needed

lwander:


lwander:
Not sure what's going with git, going to squash & merge this then refork.

PR #401, opened 2017-05-02
chore(install) auto-generate install scripts for release tracks

lwander:
It looks like the same file was checked in 3 times, but 2 are auto-generated based on the 3rd each time you invoke `./gradlew`

### DIFF
--- a/install/stable/InstallHalyard.sh
+++ b/install/stable/InstallHalyard.sh
@@ -1,0 +1,366 @@
+#!/usr/bin/env bash
+
+# This script installs Halyard.
+
+# _WARNING_ This file was auto generated, do not edit directly
+
+set -e
+set -o pipefail
+
+REPOSITORY_URL="https://dl.bintray.com/spinnaker-releases/debians"
+VERSION=""
+RELEASE_TRACK=stable
+
+# We can only currently support limited releases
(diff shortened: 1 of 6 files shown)


lwander:
ping

install/stable/InstallHalyard.sh
```diff
@@ -0,0 +1,366 @@
...
+    fi
+  fi
+  echo "deb $REPOSITORY_URL ${DISTRIB_CODENAME}-${RELEASE_TRACK} spinnaker" | tee /etc/apt/sources.list.d/halyard.list > /dev/null
+  # Java 8
+  # https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa
+  add-apt-repository -y ppa:openjdk-r/ppa
+  apt-get update ||:
+}
+
+function install_java() {
```
jtk54: Maybe at some point you should have a common bash functions file you can source rather than repeating these functions. That being said, I don't know what's being templated into these install scripts. Just a thought.

lwander: That's tricky, the other place this function appears is in an auto-generated file from the daemon, which has a different lifecycle than this file (this file only exists during install).

PR #410, opened 2017-05-03
feat(authz): Adds GitHub Teams as authz provider

ttomsu:
This gets the feature 95% of the way there, but as discussed yesterday, I am going to remove the extra `roles` layer to match the authn command

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -138,6 +138,8 @@
  * [**hal config security authz roles edit**](#hal-config-security-authz-roles-edit)
  * [**hal config security authz roles enable**](#hal-config-security-authz-roles-enable)
  * [**hal config security authz roles provider**](#hal-config-security-authz-roles-provider)
+ * [**hal config security authz roles provider github**](#hal-config-security-authz-roles-provider-github)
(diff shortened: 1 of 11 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/GoogleRoleProviderValidator.java
```diff
@@ -27,6 +27,10 @@
 public class GoogleRoleProviderValidator extends Validator<GoogleRoleProvider> {
   @Override
   public void validate(ConfigProblemSetBuilder p, GoogleRoleProvider n) {
+    if (!n.isEnabled()) {
```
lwander: I thought we were dropping this in favor of a `type` switch?

ttomsu: We talked about that after I cut this PR. Is getting fixed up in the next one.

lwander: Sounds good, feel free to merge now if that's easier for you

PR #412, opened 2017-05-03
fix(deploy): Fix the empty-clone case (no source)

lwander:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/RunningServiceDetails.java
```diff
@@ -18,22 +18,34 @@
...
+      versions.sort(Integer::compareTo);
+      versions.sort(Comparator.reverseOrder());
+    }
+
+    return versions.stream()
+        .map(i -> new ImmutablePair<>(i, instances.get(i)))
+        .filter(is -> is.getRight().stream().allMatch(i -> i.isHealthy() && i.isRunning()))
+        .findFirst()
+        .orElse(new ImmutablePair<>(null, null))
+    .getLeft();
```
jtk54: Formatting seems off.

lwander: Seems funny, but `getLeft()` is applied to the result of the other chain, which is why it's indented like this (chosen by IntelljJ).

PR #414, opened 2017-05-03
fix(authz): Collapse and options to mirror authn options. Adds valida… 

ttomsu:
…tion for Google and Github role providers.


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/AuthzValidator.java
```diff
@@ -0,0 +1,53 @@
...
+
+    switch (z.getGroupMembership().getService()) {
+      case GITHUB:
+        githubValidator.validate(p, z.getGroupMembership().getGithub());
+        break;
+      case GOOGLE:
+        googleValidator.validate(p, z.getGroupMembership().getGoogle());
+        break;
+      default:
+        // no-op.
```
lwander: I think you should "throw" or supply a warning here for future development

ttomsu: Updated the comment. Doing nothing is intentional for the SAML use case

lwander:
LGTM

PR #416, opened 2017-05-03
feat(backup): Support for config backups

lwander:
This feeds into a KMS story for remote backups.

Right now it takes your entire halconfig, determines all files it needs, and copies it and the files into `~/.hal/.backups/`, rewriting properties as needed.


lwander:
https://github.com/spinnaker/halyard/issues/417

PR #418, opened 2017-05-04
feat(config): Configure redis as a persistent store

lwander:


jtk54:
I thought we decided against supporting Redis as first class storage?

lwander:
It's the only option for people w/o a hosted cloud provider, plus the docs warn you: https://github.com/spinnaker/spinnaker.github.io/pull/81/files#diff-57d7bbafe720d184f8c4137a1de58d9bR11

lwander:
Yup it does

PR #419, opened 2017-05-04
feat(oraclebmcs): Add Oracle BMCS account configuration support to Ha…

simonlord:
…lyard

Two things of note:

1. We haven't committed to deck project yet, so i've not added to DeckProfile.java
2. Due to our Oracle BMCS SDK not being in maven, I haven't added clouddriverOracleBmcs gradle dependency as we wouldn't be able to use it. Our SDK should be in maven soon(ish).

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/oraclebmcs/OracleBMCSAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/oraclebmcs/OracleBMCSAccount.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2017 Oracle America, Inc.
+ *
+ * The contents of this file are subject to the Apache License Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.providers.oraclebmcs;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
+import com.netflix.spinnaker.halyard.config.model.v1.node.LocalFile;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class OracleBMCSAccount extends Account {
+  private String compartmentId;
+  private String userId;
+  private String fingerprint;
+  @LocalFile private String sshPrivateKeyFilePath;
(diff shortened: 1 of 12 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/oraclebmcs/OracleBMCSAccount.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class OracleBMCSAccount extends Account {
+  private String compartmentId;
+  private String userId;
+  private String fingerprint;
+  private String sshPrivateKeyFilePath;
```
lwander: I would annotate this with `@LocalFile` - this ensures it will be mounted as a required config file when doing a remote deployment.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/oraclebmcs/OracleBMCSAccountValidator.java
```diff
@@ -0,0 +1,38 @@
...
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import com.netflix.spinnaker.halyard.core.problem.v1.Problem.Severity;
+import org.springframework.stereotype.Component;
+
+@Component
+public class OracleBMCSAccountValidator extends Validator<OracleBMCSAccount> {
+  @Override
+  public void validate(ConfigProblemSetBuilder psBuilder, OracleBMCSAccount account) {
+
+    notNullOrEmpty(account.getCompartmentId(), "compartment id", psBuilder);
```
lwander: Since these are all required, in the `account add` subcommand I would add the `@Parameter` property `required = true`.

lwander:
https://github.com/spinnaker/halyard/issues/408

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/oraclebmcs/OracleBMCSAccount.java
simonlord: Woops - yes i totally thought i'd done this! It was in fact the LocalFile thing that made me realize i needed to change how the clouddriver config/accounts for oracle bmcs worked in the first place.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/oraclebmcs/OracleBMCSAccountValidator.java
simonlord: Cool adding this now. Should i still validate them as above?

lwander: Yup, this way people can edit the `~/.hal/config` by hand and still depend on the validation.

PR #427, opened 2017-05-05
refactor(deploy): Dynamically associate sidecars

lwander:
In the process I found that an update to kubectl caused contention when running multiple parallel proxies, this is fixed here as well by associating thread IDs with open connections to ensure no proxy is shared across Halyard Tasks.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/ResourceBuilder.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/ResourceBuilder.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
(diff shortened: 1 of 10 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/ResourceBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.distributed.kubernetes;
+
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.ConfigSource;
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.ServiceSettings;
+import io.fabric8.kubernetes.api.model.*;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
```
jtk54: Take out the intelliJ comments.

PR #428, opened 2017-05-05
feat(deploy): Connect to sidecar services

lwander:
This will be useful for reading health status from the consul API.


lwander:
https://github.com/spinnaker/halyard/issues/426

PR #432, opened 2017-05-05
fix(deploy): Retry failed vault calls

lwander:


halyard-deploy/src/main/resources/services/vault/client/startup/mount-config.py
```diff
@@ -5,31 +5,37 @@
...
+import time
 
 def mkdirs(path):
     subprocess.check_call(["mkdir", "-p", path])
 
+
+def subprocess_retry(supplier, retries):
+    if retries < 0:
+        logging.fatal("All retries attempted")
+        raise subprocess.CalledProcessError("")
```
jtk54: Can you log the process/command we were trying to retry? Do you have access to what the command is?

jtk54: Seems like you could just join the command list and add it as a param: `' '.join(cmd_tokens)` and then include that in the raised exception.

jtk54:
LGTM otherwise.

halyard-deploy/src/main/resources/services/vault/client/startup/mount-config.py
lwander: It's already logged, since called process error logs it, but it doesn't hurt to add again.

PR #438, opened 2017-05-07
feat(deploy): GCE red/black pipelines

lwander:
Not much going on here, there were two widespread changes
1. Switching ProviderType.id to return the id expected by clouddriver (since the google provider is referenced internal to spinnaker as "GCE").
2. Moving some of DeployConfig into a separate spring config class to allow provider accounts to generically modify themselves to include whatever config is needed if they used in the bootstrapping process. This was done to put "marketplace-spinnaker-release" into the list of external image projects in our google account.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
@@ -20,13 +20,15 @@ package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.distributed.go
 import com.google.api.services.compute.Compute;
 import com.google.api.services.compute.model.*;
 import com.netflix.frigga.Names;
+import com.netflix.spinnaker.clouddriver.google.deploy.GCEUtil;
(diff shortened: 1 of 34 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
```diff
@@ -369,7 +380,77 @@ default void ensureRunning(AccountDeploymentDetails<GoogleAccount> details,
...
+    deployDescription.put("freeFormDetails", detail);
+    deployDescription.put("network", network);
+    deployDescription.put("metadata", metadata);
+    deployDescription.put("serviceAccountEmail", serviceAccountEmail);
+    deployDescription.put("authScopes", scopes);
+    deployDescription.put("accountName", accountName);
+    deployDescription.put("account", accountName);
+    return deployDescription;
+
+    /* TODO(lwander): Google's credential class cannot be serialized as-is, making this type of construction impossible
```
jtk54: Saving this for later?

lwander: yup

PR #440, opened 2017-05-08
fix(cli): Rely on ObjectMapper to convert types

lwander:
I recently switch from Gson to Jackson to act as Retrofit's deserializer, which can't serialize types that have references to generic types. By instead having the interface deal only with `Object`s, and having `Daemon.java` convert the type directly, we avoid this problem.


jtk54:
Looks OK, is there something Jackson buys you that Gson doesn't have?

lwander:
Gson started failing (w/o any changes on my end) for every request, so I switched it out. I'm guessing some patch release went bad.

PR #443, opened 2017-05-08
feat(deploy): Improve SSH tunnel handling for GCP

lwander:
This now waits for the connection to open correctly, as well as uses a "halyard-only" ssh keypair for your instances


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleProviderUtils.java
```diff
@@ -50,12 +55,23 @@
...
 
+@Slf4j
 class GoogleProviderUtils {
   // Map from service -> the port & job managing the connection.
   private static ConcurrentHashMap<String, Proxy> proxyMap = new ConcurrentHashMap<>();
   public static String getNetworkName() {
     return "spinnaker-hal";
   }
+  public static String getSshKeyFile() {
+    return Paths.get(System.getProperty("user.home"), ".ssh/google_compute_halyard").toString();
```
jtk54: Could be configurable in the future.

lwander: Yup could be, I don't see the value (yet) in doing that so I'm skipping it for now.

PR #450, opened 2017-05-11
feat(deploy): Allow users to specify service settings

lwander:
Documentation in the github.io repo incoming.

This applies to runtime settings like which port a service is deployed on, or what address they should bind to.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/HasServiceSettings.java
```diff
@@ -20,8 +20,11 @@
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentConfiguration;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
 
+import java.util.Optional;
```
jtk54: Accidental import?

lwander: One could say it's Optional here

PR #452, opened 2017-05-12
feat(deploy): Multi VM monitoring

lwander:
This isn't complete, since the monitoring config folder isn't pointing at the expected /opt/spinnaker-monitoring/config/, but instead /opt/spinnaker-monitoring making it difficult for k8s to mount spinnaker-monitoring.yml. This will either take change to spinnaker/spinnaker-monitoring or the config mounting in halyard. I'm talking to @ewiseblatt about it.


ewiseblatt:
I'll look into changing monitoring later this afternoon after I finish what
I'm doing now.

ewiseblatt:
In the meantime, you can use the --config parameter to point to the config
dir.
When I change the default, you wont be affected other than having a
superfluous --config.

lwander:
I can make the change too after lunch, unless you have specific ideas about how to deprecate the old path.

lwander:
https://github.com/spinnaker/spinnaker-monitoring/pull/51

PR #457, opened 2017-05-15
feat(validation): Add validation for AWS accounts and S3

jervi:
#116: Add validation for AWS accounts and S3 persistent storage. The S3 bucket validation will create the bucket if it does not exist (this makes the S3 validation behave in the same way as the GCS validation).

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/S3Validator.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/S3Validator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Microsoft, Inc.
+ * Copyright 2017 Schibsted ASA.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


jervi:
@lwander PTAL

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/S3Validator.java
```diff
@@ -16,15 +16,50 @@
...
 
+import java.util.Optional;
+
 @Component
 public class S3Validator extends Validator<S3PersistentStore> {
   @Override
   public void validate(ConfigProblemSetBuilder ps, S3PersistentStore n) {
-    // TODO(lwander) find someone to handle https://github.com/spinnaker/halyard/issues/116
+    DaemonTaskHandler.message(String.format("Validating %s with %s", n.getNodeName(), getClass().getSimpleName()));
+    AWSCredentialsProvider credentialsProvider = DefaultAWSCredentialsProviderChain.getInstance();
```
lwander: It's probably best if you instantiate these credentials by importing front50 as a dependency. It already does things like [create a nonexistent bucket](https://github.com/spinnaker/front50/blob/0c0ed5848b6f46bfc315e976c86d690b4b850758/front50-s3/src/main/java/com/netflix/spinnaker/front50/model/S3StorageService.java#L61) and verify that you have access to it.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/aws/AwsAccountValidator.java
```diff
@@ -17,17 +17,36 @@
...
 
 @Component
 public class AwsAccountValidator extends Validator<AwsAccount> {
   @Override
   public void validate(ConfigProblemSetBuilder p, AwsAccount n) {
-    // TODO(https://github.com/spinnaker/halyard/issues/116)
-    p.addProblem(Severity.WARNING, "No validators exist for this provider.");
+    DaemonTaskHandler.message(String.format("Validating %s with %s", n.getNodeName(), getClass().getSimpleName()));
+
+    AWSCredentialsProvider credentialsProvider = DefaultAWSCredentialsProviderChain.getInstance();
```
lwander: This should similarly use the credential initializer in clouddriver.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/S3Validator.java
jervi: Halyard already has a dependency on front50, but it is pinned to version 1.72.0 in spinnaker-dependencies. Should I update the dependency there, or should I add an explicit dependency on a newer front50 version? In 1.72.0 your implementation of `ensureBucketExists()` doesn't exist yet.

jervi: I've sort of fixed this. I still do the initialisation of the credentials here, but I use the `S3StorageService` to validate and create the bucket. I've also added a region property to the s3 edit command.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/aws/AwsAccountValidator.java
jervi: I don't really understand how this is supposed to work, or what benefit it will provide...

lwander: Using the same method for initializing the credentials gives you a higher degree of confidence that your config will work once it passes Halyard's checks. If it turns out that it's too complicated we can skip this for now and move on to getting this to work on k8s.

lwander:
We talked this over in slack and it seems that to initialize the credential handling for AWS you need to do a lot of manual spring initialization, which might not be the best way forward. Perhaps we can look into PolicyTester in the future.

PR #463, opened 2017-05-15
fix(deploy): spinnaker user & group must own all files

lwander:
Looks like I was omitting group before, perhaps this is why some services didn't have trouble, but Rosco did.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverBootstrapProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverBootstrapProfileFactory.java
@@ -76,7 +76,7 @@ public class ClouddriverBootstrapProfileFactory extends SpringProfileFactory {
       log.warn("Attempting to perform a distributed deployment to account \"" + account.getName() + "\" without a discovery mechanism");
     }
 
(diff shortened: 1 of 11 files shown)


lwander:
Can we take a look at what state your VM was in again? For whatever reason I only remember a single user column (no group) from your `ls -l` output.

ewiseblatt:
ewiseblatt@hal-gce:~$ ls -l /home/ewiseblatt/.hal-staging/total 56-rw-------
1 root root 5936 May 15 19:39 <redacted>-config-r-------- 1 root root 1246
May 15 19:39 250392558-google-credentials.json-r-------- 1 root root 1246
May 15 19:39 339816717-google-credentials.jsondrwxr-xr-x 2 root root 4096
May 15 19:39 apache2-rw-r--r-- 1 root root 1714 May 15 19:39
clouddriver.yml-rw-r--r--
1 root root 847 May 15 19:39 echo.yml-rw-r--r-- 1 root root 1091 May 15
19:39 front50.yml-rw-r--r-- 1 root root 426 May 15 19:39 gate.yml-rw-r--r--
1 root root 538 May 15 19:39 igor.yml-rw-r--r-- 1 root root 1183 May 15
19:39 orca.yml-rw-r--r-- 1 root root 1732 May 15 19:39 rosco.yml-rw-r--r--
1 root root 3365 May 15 19:39 settings.js-rw-r--r-- 1 root root 3322 May 15
19:39 spinnaker.yml

Also, the yml files should be 0600 or 0400 because they often contain
passwords.

PR #467, opened 2017-05-16
feat(dcos): Add DC/OS configuration support to Halyard (#408)

willgorman:
There's still a little bit of additional work to be done around validating using clouddriver-dcos after the DC/OS changes are added to clouddriver but this covers all of the basics.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
@@ -1,0 +1,78 @@
+package com.netflix.spinnaker.halyard.cli.command.v1.config.providers.dcos;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.google.common.collect.Lists;
+import com.netflix.spinnaker.halyard.cli.command.v1.config.providers.account.AbstractAddAccountCommand;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Provider;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSAccount;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSAccount.ClusterCredential;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DockerRegistryReference;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.Objects.isNull;
+
+@Parameters(separators = "=")
+public class DCOSAddAccountCommand extends AbstractAddAccountCommand {
+  protected String getProviderName() {
+    return Provider.ProviderType.DCOS.getId();
+  }
+
+  @Parameter(
+      names = "--docker-registries",
+      description = DCOSCommandProperties.DOCKER_REGISTRIES_DESCRIPTION,
+      required = true,
+      variableArity = true
+  )
+  private List<String> dockerRegistries = new ArrayList<>();
+
+  @Parameter(
+      names = "--uid",
+      description = "User or service account identifier",
+      required = true
+  )
+  private String uid;
+
+  @Parameter(
+      names = "--cluster",
+      description = "Reference to the name of the cluster from the set of clusters defined for this provider",
+      required = true
+
+  )
+  private String cluster;
+
+  //TODO(willgorman): might be good to support loading from a file
+  @Parameter(
+      names = "--service-key",
+      description = "Secret key for service account authentication"
+  )
+  private String serviceKey;
+
+  @Parameter(
+      names = "--password",
+      description = "Password for a user account"
+  )
+  private String password;
+
+  @Override
+  protected Account buildAccount(String accountName) {
+    DCOSAccount account = (DCOSAccount) new DCOSAccount().setName(accountName);
+    dockerRegistries.forEach(registryName -> account.getDockerRegistries().add(new DockerRegistryReference().setAccountName(registryName)));
+
+    if (!isNull(serviceKey) && !isNull(password)) {
+      throw new IllegalArgumentException("Only one of --service-key or --password may be set");
+    }
+
+    account.setClusters(Lists.newArrayList(new ClusterCredential(cluster, uid, password, serviceKey)));
+
+    return account;
+  }
+
+  @Override
+  protected Account emptyAccount() {
+    return new DCOSAccount();
+  }
+}
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/cluster/DCOSListClusterCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/cluster/DCOSListClusterCommand.java
@@ -1,0 +1,53 @@
+package com.netflix.spinnaker.halyard.cli.command.v1.config.providers.dcos.cluster;
+
+import lombok.Getter;
+
+import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.cli.services.v1.Daemon;
+import com.netflix.spinnaker.halyard.cli.services.v1.OperationHandler;
+import com.netflix.spinnaker.halyard.cli.ui.v1.AnsiUi;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Provider;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSCluster;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSProvider;
+
+import java.util.List;
+
+/**
+ *
+ */
+@Parameters(separators = "=")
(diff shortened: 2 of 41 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Parameter(
+      names = "--cluster",
+      description = "Reference to the name of the cluster from the set of clusters defined for this provider",
+      required = true
+
+  )
+  private String cluster;
+
+  //TODO(willgorman): might be good to support loading from a file
```
lwander: Just curious, would this involve adding a config parameter to clouddriver to support a `serviceKeyFile` or something similar?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
```diff
@@ -0,0 +1,80 @@
...
+
+    return account;
+  }
+
+  @Override
+  protected Account emptyAccount() {
+    return new DCOSAccount();
+  }
+
+
```
lwander: Extra newline

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/cluster/DCOSListClusterCommand.java
```diff
@@ -0,0 +1,56 @@
...
+    return new OperationHandler<Provider>()
+        .setFailureMesssage("Failed to get provider " + providerName + ".")
+        .setOperation(Daemon.getProvider(currentDeployment, providerName, !noValidate))
+        .get();
+  }
+
+  @Override
+  protected void executeThis() {
+    Provider provider = getProvider();
+    if (!provider.providerType().equals(Provider.ProviderType.DCOS)) {
```
lwander: Why is this check needed? If this is the `DCOS` list cluster command, won't this always fail?

lwander: Maybe my concern is that the exception below should instead indicate an assertion was violated, and isn't a user error.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ClusterService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    for (int i = 0; i < dcosProvider.getClusters().size(); i++) {
+      Cluster cluster = dcosProvider.getClusters().get(i);
+      if (cluster.getNodeName().equals(clusterName)) {
+        dcosProvider.getClusters().set(i, newCluster);
+        return;
+      }
+    }
+
+    throw new HalException(new ConfigProblemBuilder(Problem.Severity.FATAL, "Cluster \"" + clusterName + "\" wasn't found").build());
```
lwander: `HalException` now has a constructor that takes `Severity, String` directly rather than making you use a builder.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ClusterService.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  public void setCluster(String deploymentName, String providerName, String clusterName, Cluster newCluster) {
+    Provider provider = providerService.getProvider(deploymentName, providerName);
+    if (!provider.providerType().equals(Provider.ProviderType.DCOS)) {
+      throw new IllegalArgumentException("Provider type does not accept cluster configuration");
+    }
+
+    // TODO(willgorman): could also introduce a ClusterProvider subclass of Provider instead of going straight to DC/OS here
+    DCOSProvider dcosProvider = (DCOSProvider) provider;
```
lwander: I would prefer if you could switch this to a provider subclass, similar to the `HasImageProvider<T>` so by the time the next person comes around to possibly support a cluster-like construct they aren't tempted to expand these functions into a series of `if (provider ==...)` statements.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dcos/DCOSAccountValidator.java
```diff
@@ -0,0 +1,156 @@
...
+ * TODO: use clouddriver components for full validation (e.g. account name))
+ */
+@Component
+public class DCOSAccountValidator extends Validator<DCOSAccount> {
+  @Override
+  public void validate(final ConfigProblemSetBuilder problems, final DCOSAccount account) {
+    DeploymentConfiguration deploymentConfiguration;
+
+    /**
+     * I have copied
```
lwander: Heh, it's OK and my fault ;)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dcos/DCOSAccountValidator.java
```diff
@@ -0,0 +1,156 @@
...
+
+      if (!Strings.isStringEmpty(c.getPassword()) && !Strings.isStringEmpty(c.getServiceKey())) {
+        problems.addProblem(ERROR, "Account has both a password and service key")
+            .setRemediation("Remove either the password or service key.");
+      }
+    });
+
+  }
+
+  private void validateDockerRegistries(ConfigProblemSetBuilder psBuilder, DCOSAccount account,
```
lwander: Could you pull this into a static method somewhere between k8s & dcos? I think this function looks identical (minus some of the DC/OS error message prefixes).

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
```diff
@@ -50,8 +50,8 @@
 
   @RequestMapping(value = "/", method = RequestMethod.GET)
   DaemonTask<Halconfig, List<Account>> accounts(@PathVariable String deploymentName, @PathVariable String providerName,
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
-      @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Severity severity) {
+                                                @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
+                                                @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Severity severity) {
```
lwander: Please unindent :)

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
willgorman: Possibly, yeah.  I was initially thinking of just reading a file in from a path on the machine running the CLI and adding the content to the config instead of needing to pass the content as an argument.  Looking at the `@LocalFile` stuff I assume that instructs Halyard to copy the files to the remote deployments and that might be a better way to go?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/dcos/ClusterController.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @RequestMapping(value = "/cluster/{clusterName:.+}", method = RequestMethod.PUT)
+  DaemonTask<Halconfig, Void> setCluster(
+      @PathVariable String deploymentName,
+      @PathVariable String providerName,
+      @PathVariable String clusterName,
+      @RequestParam(required = false, defaultValue = DefaultControllerValues.validate) boolean validate,
+      @RequestParam(required = false, defaultValue = DefaultControllerValues.severity) Problem.Severity severity,
+      @RequestBody Object rawCluster) {
+    DCOSCluster cluster = objectMapper.convertValue(
```
lwander: These can be converted to `HasClustersProvider` now, right?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dcos/DCOSAddAccountCommand.java
lwander: Yes that's right, anything annotated with `@LocalFile` is copied/mounted in a remote deployment. It's also stored & encrypted during halyard backups (almost complete).

willgorman: Ok, it sounds like we should definitely have clouddriver support loading that from a file to take advantage of the encryption with halyard.  I'll make sure we get issues logged to clouddriver and halyard to follow up on that.

lwander:
LGTM! Once last thing before we merge - can you add yourself or your company (whoever you're doing this work on behalf of) to this list of AUTHORS as a part of this PR? https://github.com/spinnaker/halyard/blob/master/AUTHORS

PR #469, opened 2017-05-16
feat(validate): Version mismatch warning

lwander:
I was going to add generic support for notifications, but realized that different kinds of notifications merit different treatment, and didn't want to hard-code all of those, or come up with something robust enough to handle all of our future notification worries. So instead I'm going to be splitting up the notification work into three categories:

1. Spinnaker update available: Addressed here, see below screenshots.
2. Halyard update available: Will be addressed similar to how it was done here.
3. Critical spinnaker announcement: I can only think of security-style patches that need to be applied by the Halyard operator, and therefore think that special cases of 1 & 2 are more appropriate. 
update-warning-2
update-warning


lwander:
@spinnaker/google-reviewers any ideas on other types of notifications we should support before we release?

lwander:
I'm going to merge this since I'm blocked on it, but will leave the above question open.

PR #473, opened 2017-05-18
fix(config): Ensure rosco config is validated on edit

lwander:


duftler:
Unit test failed.

lwander:
Classic unit tests

PR #484, opened 2017-05-19
fix(cli): version edit wording to match aliases

lwander:
@ewiseblatt What do you think about this wording?

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditVersionCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditVersionCommand.java
@@ -35,8 +35,8 @@ public class EditVersionCommand extends AbstractConfigCommand {
   @Parameter(
       names = "--version",
       required = true,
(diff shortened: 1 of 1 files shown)


ewiseblatt:
LGTM, except shouldnt there be an alias for the latest published patch
release for a given branch? I'd think you only care about specific versions
when rolling back.

lwander:
I think it's best to wait before adding that behavior - it would mean updates will be applied to stable releases without the user explicitly asking for them - we could of course start publishing these aliases, but I think it might confuse/surprise users at first. WDYT, @jtk54?

lwander:
For a little more context:

Halyard will either

1. complain with a warning that a new patch is out, or
2. prevent you from deploying an old version if we've explicitly made it "illegal"

lwander:
So I think for now we have the tools in place to ensure users will be up-to-date on the latest patch version

ewiseblatt:
ok

jtk54:
I think we should avoid doing implicit things until users are familiar with what is going on with Halyard.

ewiseblatt:
I think it should be easy to do the things we want them to be doing.

lwander:
Agreed, let's figure out how best to do that. Maybe the alias should instead by `1.4.*` or something to be clear you're matching the latest patch always.

PR #485, opened 2017-05-19
feat(cli): Show current version on `hal version list`

lwander:


lwander:
`hal config version` does that for you - this feature was requested by both Scott & Christopher independently

PR #499, opened 2017-05-24
fix(authz): Remove enable/disable option from individual role provide…

ttomsu:
…rs, rename NONE source to EXTERNAL to match Fiat naming.


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/AuthzValidator.java
```diff
@@ -44,8 +45,8 @@ public void validate(ConfigProblemSetBuilder p, Authz z) {
       case GOOGLE:
         googleValidator.validate(p, z.getGroupMembership().getGoogle());
         break;
+      case EXTERNAL: // fallthrough
       default:
-        // Used when SAML authentication is enabled. No additional config is needed.
```
lwander: Can we add an exception here since this should no longer be reachable?

ttomsu: Multiple things are falling through here, so we don't want that quite yet.

PR #508, opened 2017-05-25
feat(validation): Support AWS assumeRole and setting credentials at the command line

jervi:
Second take at implementing validation of AWS accounts and S3 buckets. Please take a look @lwander  @ewiseblatt.

It kind of supports the assumeRole, and you can set credentials in the cli. It's not properly tested, so expect the unexpected. I just wanted to get it out there and collect some feedback.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1,4 +1,4 @@
-_Version: 0.26.0-SNAPSHOT_
+_Version: 0.27.0-SNAPSHOT_
 
 # Table of Contents
 
@@ -1030,7 +1030,11 @@ hal config provider aws account add ACCOUNT [parameters]
 ```
 #### Parameters
 `ACCOUNT`: The name of the account to operate on.
- * `--account-id`: Your AWS account ID to manage. See http://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html for more information.
+ * `--access-key-id, --access-key`: Your AWS Access Key ID. If not provided, Halyard will try to find AWS as described at http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default
+ * `--account-id`: (*Required*) Your AWS account ID to manage. See http://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html for more information.
+ * `--assume-role, --role`: If set, Halyard will configure a credentials provider that use AWS Security Token Service to assume the specified role.
+
+Example: "user/spinnaker" or "role/spinnakerManaged"
  * `--default-key-pair`: Provide the name of the AWS key-pair to use. See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html for more information.
  * `--discovery`: The endpoint your Eureka discovery system is reachable at. See https://github.com/Netflix/eureka for more information.
 
@@ -1041,6 +1045,7 @@ Using {{region}} will make Spinnaker use AWS regions in the hostname to access d
  * `--no-validate`: (*Default*: `false`) Skip validation.
  * `--regions`: (*Default*: `[]`) The AWS regions this Spinnaker account will manage.
  * `--required-group-membership`: (*Default*: `[]`) A user must be a member of at least one specified group in order to make changes to this account's cloud resources.
+ * `--secret-key`: Your AWS Secret Key.
 
 ---
 ## hal config provider aws account delete
@@ -1066,9 +1071,13 @@ hal config provider aws account edit ACCOUNT [parameters]
 ```
 #### Parameters
 `ACCOUNT`: The name of the account to operate on.
+ * `--access-key-id, --access-key`: Your AWS Access Key ID. If not provided, Halyard will try to find AWS as described at http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default
  * `--account-id`: Your AWS account ID to manage. See http://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html for more information.
  * `--add-region`: Add this region to the list of managed regions.
  * `--add-required-group-membership`: Add this group to the list of required group memberships.
+ * `--assume-role, --role`: If set, Halyard will configure a credentials provider that use AWS Security Token Service to assume the specified role.
+
+Example: "user/spinnaker" or "role/spinnakerManaged"
  * `--default-key-pair`: Provide the name of the AWS key-pair to use. See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html for more information.
  * `--discovery`: The endpoint your Eureka discovery system is reachable at. See https://github.com/Netflix/eureka for more information.
 
@@ -1081,6 +1090,7 @@ Using {{region}} will make Spinnaker use AWS regions in the hostname to access d
  * `--remove-region`: Remove this region from the list of managed regions.
  * `--remove-required-group-membership`: Remove this group from the list of required group memberships.
  * `--required-group-membership`: A user must be a member of at least one specified group in order to make changes to this account's cloud resources.
+ * `--secret-key`: Your AWS Secret Key.
 
 ---
 ## hal config provider aws account get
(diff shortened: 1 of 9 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
```diff
@@ -34,48 +35,70 @@ protected String getProviderName() {
...
   private List<String> regions = new ArrayList<>();
 
+  @Parameter(
+    names = {"--assume-role", "--role"},
+    description = AwsCommandProperties.ASSUME_ROLE_DESCRIPTION
+  )
+  private String assumeRole;
+
+  @Parameter(
+    names = {"--access-key-id", "--access-key"},
```
lwander: I think there is a `password = true` parameter that ensures these don't have to appear in your .bash_history

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/aws/AwsAccount.java
```diff
@@ -17,24 +17,29 @@
...
+@ToString(exclude = {"awsCredentials"})
 public class AwsAccount extends Account {
 
   private String defaultKeyPair;
   private String edda;
   private String discovery;
   private String accountId;
   private List<AwsRegion> regions = new ArrayList<>();
+  private String assumeRole;
+  private AWSCredentials awsCredentials;
```
lwander: I don't think this should appear here, I don't know how it's serialized & this should be easily editable via yaml

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/persistentStorage/S3PersistentStore.java
```diff
@@ -25,6 +25,7 @@
 @Data
 @EqualsAndHashCode(callSuper = false)
 public class S3PersistentStore extends PersistentStore {
+  private String accountName;
```
lwander: Is this to avoid duplicating the secret key fields here? We originally took this approach with GCS but dropped it because it required granting our keys super extra privileges in clouddriver that weren't required for deploying infrastructure, only managing buckets

ewiseblatt: I noticed this pattern somewhere else so I didnt say anything since I figured it was a standard pattern.

However I'd like to note that it could be that the account for the storage is not something that you want to deploy infrastructure to (other than perhaps Spinnaker). So you wouldnt necessarily want it visible in clouddriver.

lwander: Agreed, we removed it from GCS for that reason

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
jervi: It is, but it works a little bit too well... When true, JCommander no longer accepts the parameter at all, which makes scripting difficult. It displays an interactive prompt even if the parameter is provided.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/aws/AwsAccount.java
jervi: Fixed

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/persistentStorage/S3PersistentStore.java
jervi: Ok, will remove and replace with access key and secret key.

jervi: What about assume role? Does that make sense here?

lwander: I think so - you can have front50 do assume role style authentication, right?

docs/commands.md
```diff
@@ -1030,7 +1030,9 @@ hal config provider aws account add ACCOUNT [parameters]
 ```
 #### Parameters
 `ACCOUNT`: The name of the account to operate on.
+ * `--access-key-id, --access-key`: Your AWS Access Key ID. If not provided, Halyard will try to find AWS as described at http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default
  * `--account-id`: Your AWS account ID to manage. See http://docs.aws.amazon.com/IAM/latest/UserGuide/console_account-alias.html for more information.
+ * `--assume-role, --role`: If set, Halyard will configure a credentials provider that use AWS Security Token Service to assume the specified role. Should either start with "arn:aws:iam::" or just the part after the last colon.
```
lwander: What is the role name used for, does the authentication mechanism use it, or is it just for our validation?

jervi: Yes, it is used by the authentication mechanism.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
```diff
@@ -34,48 +35,70 @@ protected String getProviderName() {
...
+    description = AwsCommandProperties.DISCOVERY_DESCRIPTION
   )
   private String discovery;
 
   @Parameter(
-      names = "--account-id",
-      description = AwsCommandProperties.ACCOUNT_ID_DESCRIPTION
+    names = "--account-id",
+    description = AwsCommandProperties.ACCOUNT_ID_DESCRIPTION,
+    required = true
```
jervi: Am I wrong in assuming that account ID is required?

lwander: I think you're right here - can @gardleopard confirm?

docs/commands.md
lwander: Nice

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
lwander: Yeah it sets the arity to 0, which can be confusing

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
jervi: I guess the only thing I can think of is that we in some hypothetical scenario could use AWS account lookup to find the account id, but I don't think we should support that.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
jervi: So, the conclusion is to not use `password = true`, right?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
gardleopard: We tested with some local configs and it seems like account id is required.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/aws/AwsAddAccountCommand.java
lwander: No, since that encourages putting secrets into your bash history.

lwander: If it becomes a problem we should look at other ways to supply those secrets, but I think we should be wary of the security implications first

PR #510, opened 2017-05-28
fix(bake): Don't install monitoring api until after bake

lwander:


lwander:
Context: datadog pip package seems to be corrupting our bakes, this is a temporary workaround

PR #511, opened 2017-05-28
fix(version): 'nightly' is no longer a valid alias

lwander:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentConfiguration.java
```diff
@@ -45,7 +45,7 @@
    *
    * @see Halconfig#halyardVersion
    */
-  String version = "nightly";
+  String version = "master-latest-unvalidated";
```
ewiseblatt: why is this "master"?
doesnt "master" refer to the branch?

lwander: Yup - it's effectively what "nightly" was before we stopped using that alias.

ewiseblatt: I mean, how do you know you are in master? This shouldnt be a static string. In fact none of the tag should be. The naming policy should be defined externally.

lwander: This is a default/placeholder before we get have a proper "latest" alias that points to the latest validated Spinnaker. I know that for the next week before we release, this alias is just a QOL improvement that effectively points at what "nightly" pointed at.

PR #530, opened 2017-06-05
fix(deploy): Set default bake account in orca

lwander:
@ewiseblatt FYI


lwander:
Orca & clouddriver don't otherwise share config properties, so there is no way to set a static default that depends on SPEL

PR #536, opened 2017-06-06
fix(validate): Propagate s3 errors correctly

lwander:


lwander:
dupe of https://github.com/spinnaker/halyard/pull/540

PR #540, opened 2017-06-08
Don't validate/create S3 bucket twice

jervi:
`new S3Config().s3StorageService(s3Client, s3Properties);` calls `s3StorageService.ensureBucketExists();`, so no need to do it again.
Added bonus: the catch clause actually catches exceptions.


lwander:
Thanks!

I have a similarish PR here: https://github.com/spinnaker/halyard/pull/536 

Can you move the rest of the credential construction into the catch block and I'll close my PR?

PR #555, opened 2017-06-12
fix(bom): Retry storage request in case of 503 error

EricJizbaMSFT:


EricJizbaMSFT:
@lwander PTAL. I was able to test the happy-path, but I didn't know how to simulate a 503 to make sure this actually works.

In case you're curious, I'm mainly fixing this because I'm updating our quickstart templates to leverage Halyard and those run 'hal deploy apply' as part of setup. The error only happens ~10% of the time, but we want our templates to be very stable

PR #557, opened 2017-06-13
feat(timezone): Makes timezone CLI-configurable

lwander:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentService.java
```diff
@@ -42,8 +45,26 @@
...
 
+  public void setDeploymentConfiguration(String deploymentName, DeploymentConfiguration deploymentConfiguration) {
+    Halconfig halconfig = halconfigParser.getHalconfig();
+    List<DeploymentConfiguration> deploymentConfigurations = halconfig.getDeploymentConfigurations();
+
+    for (int i = 0; i < deploymentConfigurations.size(); i++) {
+      DeploymentConfiguration test = deploymentConfigurations.get(i);
+      if (test.getName().equals(deploymentName)) {
+        deploymentConfigurations.set(i, deploymentConfiguration);
+        return;
```
duftler: Might read easier to have one block to locate the index of the matching config, and then another to set the new config. As it stands now, it looks like you are setting a new config while you are still searching through the list.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/DeploymentConfigurationValidator.java
```diff
@@ -39,6 +41,12 @@
 
   @Override
   public void validate(ConfigProblemSetBuilder p, DeploymentConfiguration n) {
+    String timezone = n.getTimezone();
+
+    if (Arrays.stream(TimeZone.getAvailableIDs()).noneMatch(t -> t.equals(timezone))) {
+      p.addProblem(Problem.Severity.WARNING, "Timezone " + timezone + " does not match any known canonical timezone ID");
```
duftler: Maybe provide a link to the list of canonical timezone ids?

PR #583, opened 2017-06-30
Some minor grammar changes in the account validator

dlovitch:
Hi,

Just made a couple of grammar changes that I came across when doing the setup.


lwander:
Thanks!

PR #587, opened 2017-06-30
feat(google/xpn): Add support for Google shared VPC networks.

duftler:


duftler:
Relates to https://github.com/spinnaker/rosco/pull/212.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/google/GoogleEditBakeryDefaultsCommand.java
```diff
@@ -45,6 +45,12 @@ protected String getProviderName() {
   private String network;
 
   @Parameter(
+      names = "--network-project-id",
```
lwander: Why `-id`? Everywhere else we just refer to it as `project`, in component config & halyard commands.

duftler: For consistency with packer, and also because it's actually the project id.

PR #593, opened 2017-07-07
feat(deploy): Scale down orca & rosco during deploys

lwander:
When possible, orca & rosco will be scaled down during deployments (if they aren't handling any work).

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
@@ -276,16 +276,45 @@ public class DistributedDeployer<T extends Account> implements Deployer<Distribu
     List<Integer> allRoscos = new ArrayList<>(executionsByServerGroupVersion.keySet());
     allRoscos.sort(Integer::compareTo);
 
-    int roscoCount = allRoscos.size();
-    if (roscoCount <= MAX_REMAINING_SERVER_GROUPS) {
+    cleanupServerGroups(details, roscoService, roscoSettings, executionsByServerGroupVersion, allRoscos);
+  }
+
+  private <S, T extends Account> void cleanupServerGroups(AccountDeploymentDetails<T> details,
+      DistributedService<S, T> service,
+      ServiceSettings settings,
+      Map<Integer, Integer> workloadByVersion,
+      List<Integer> runningVersions) {
+    int runningVersionCount = runningVersions.size();
+
+    if (runningVersionCount <= 1) {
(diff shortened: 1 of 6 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
```diff
@@ -318,18 +346,6 @@ public RemoteAction deploy(DistributedServiceProvider<T> serviceProvider,
...
     allOrcas.sort(Integer::compareTo);
 
-    int orcaCount = allOrcas.size();
-    if (orcaCount <= MAX_REMAINING_SERVER_GROUPS) {
-      return;
-    }
-
-    allOrcas = allOrcas.subList(0, orcaCount - MAX_REMAINING_SERVER_GROUPS);
-    for (Integer orcaVersion : allOrcas) {
-      // TODO(lwander) consult clouddriver to ensure this orca isn't enabled
```
lwander: TODO was dropped because the orca versions outside of MAX_REMAINING_SERVER_GROUPS cannot be enabled, since the rollback operation requested by halyard ensures the highest sequence number is always the only one serving traffic.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
```diff
@@ -276,16 +276,44 @@ public RemoteAction deploy(DistributedServiceProvider<T> serviceProvider,
...
+  }
+
+  private <S, T extends Account> void cleanupServerGroups(AccountDeploymentDetails<T> details,
+      DistributedService<S, T> service,
+      ServiceSettings settings,
+      Map<Integer, Integer> workloadByVersion,
+      List<Integer> runningVersions) {
+    int runningVersionCount = runningVersions.size();
+
+    if (runningVersionCount <= 1) {
```
jtk54: This is not the usual case, correct? Maybe add logging here so you avoid possibly silently failing.

lwander: This happens on every fresh deploy - you'll only have one of each service. I'll add a log statement but it's expected to be hit

jtk54: Got it, still maybe worth it.

lwander: Agreed & logged

Issue #563, opened 2017-06-16
The `postInstall.sh` restarts the service without a user


TYsewyn:
At the end of the `postInstall.sh` script the halyard service will be restarted.
This in turn will execute following command:

`sudo -u $(cat /var/spinnaker/config/halyard-user) -g $(cat /var/spinnaker/config/halyard-user) /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &`

Since `/var/spinnaker/config/halyard-user` will only be written after the installation, and therefor after the `postInstall.sh` script, this results in a faulty restart of the service.

lwander:
This one's a bit tricky to fix, since the user needs to specify who to run Halyard as before the install can be considered successful. The only default that comes to mind is "root", and that feels off. The service restart was added to the postInstall script to ensure package updates were successful, after the Halyard had been installed correctly once.

PR #602, opened 2017-07-13
fix(install): Use halyard-user to assign ownership to log dir

lwander:
There are duplicate changes to the install script, because the InstallHalyard.sh.gen generates (with some small templating) the other two InstallHalyard.sh files.

By first setting halyard-user (before Halyard is installed) we can be sure that the log dir can be correctly assigned on both installs & updates of Halyard.


lwander:
Fixes breakage introduced here: https://github.com/spinnaker/halyard/pull/572 and also uses @TYsewyn's approach to fixing the install path in https://github.com/spinnaker/halyard/issues/563

PR #606, opened 2017-07-18
chore(install): Change installer to download install script

lwander:
This breaks out the nightly & stable installers, which were previously
autogenerated. I want to prepare for a more generic, platform
independent installer that doesn't require debians or upstart to
function. In the process I'll be updating the nightly installer &
halyard CLI behavior in the nightly release before migrating changes to
the stable installer.


install/nightly/InstallHalyard.sh
```diff
@@ -1,427 +1,23 @@
...
 
-set -e
-set -o pipefail
+echo "Downloading nightly halyard installer..."
 
-REPOSITORY_URL="https://dl.bintray.com/spinnaker-releases/debians"
-SPINNAKER_REPOSITORY_URL="https://dl.bintray.com/spinnaker-releases/debians"
-SPINNAKER_DOCKER_REGISTRY="gcr.io/spinnaker-marketplace"
-SPINNAKER_GCE_PROJECT="marketplace-spinnaker-release"
+curl -O https://raw.githubusercontent.com/spinnaker/halyard/master/install/Installer.sh
```
jtk54: Seems like we would need to produce the Halyard JARs and publish them to GCS before we flip this switch, correct? Moreover, are we preserving the old method of installing Halyard for folks who aren't ready to migrate yet?

lwander: No switch being flipped here, just switching to a single install script for the old path rather than two that are written from a template. The goal was just to lower the cognitive overhead required to understand all the different new scripts in the `./install` directory.

jtk54: Ok SGTM.

PR #607, opened 2017-07-18
fix(backup): Only close archive entries with data

lwander:
Several users have now seen the backup controller fail to create a tar archive due to unclosed entries with fairly simply hal config directories. I haven't been able to reproduce this error, but my best guess is that archive entries created for any directory are causing grief for their local tar archiver. Reading [the spec](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveOutputStream.html#closeArchiveEntry--) it could (?) be interpreted that only entries with data can be closed.


lwander:
Also possible: A `HalException` was thrown during the tar operation, causing the archive to be closed prematurely. I've added logging to catch this

PR #609, opened 2017-07-19
chore(install): New install path

lwander:
The idea is:

1. We package up the JARs & generated wrapper scripts built by gradle
2. Users download that package as well as a script that starts the daemon when it's not running yet

Each system will need to supply a script to download java8, as well as provide a script to start the daemon when it's not running.

### DIFF
--- a/install/debian/InstallHalyard.sh
+++ b/install/debian/InstallHalyard.sh
@@ -1,0 +1,218 @@
+#!/usr/bin/env bash
+
+set -e
+
+function process_args() {
+  while [ "$#" -gt "0" ]
+  do
+    local key="$1"
+    shift
+    case $key in
+      --user)
+        echo "user"
+        HAL_USER="$1"
+        shift
+        ;;
+      --version)
+        echo "version"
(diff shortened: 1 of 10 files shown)


install/debian/InstallHalyard.sh
```diff
@@ -0,0 +1,218 @@
...
+  artifacts:
+    debianRepository: $SPINNAKER_REPOSITORY_URL
+    dockerRegistry: $SPINNAKER_DOCKER_REGISTRY
+    googleImageProject: $SPINNAKER_GCE_PROJECT
+EOL
+
+  echo $HAL_USER > /opt/spinnaker/config/halyard-user
+
+  cat > $halconfig_dir/uninstall.sh <<EOL
+#!/usr/bin/env bash
```
jtk54: Very meta.

lwander: Good thing that script doesn't write any scripts.

lwander:
No, whoops - thanks.

PR #613, opened 2017-07-19
chore(run): Restart daemon when PID is unknown

lwander:


startup/debian/hal
```diff
@@ -40,16 +40,27 @@ function start_daemon() {
     echo ""
 }
 
+function kill_daemon() {
+    set +e
+    pkill -f '/opt/halyard/lib/halyard-web'
```
jtk54: could `|| true` this also.

lwander: Ah neat

PR #623, opened 2017-07-21
Support talking to a remote Daemon from the CLI

johnrengelman:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
```diff
@@ -643,15 +643,12 @@ private static ObjectMapper getObjectMapper() {
     return objectMapper;
   }
 
-  // TODO(lwander): setup config file for this
```
lwander: Thank you :)

johnrengelman:
I think a config file would make sense as an improvment after this.

lwander:
Agreed

PR #627, opened 2017-07-25
To be able to use PREPARE_ENVIRONMENT variable disabling dependency installation.

cookingcodewithme:
Today, halyard has a problem with the following error. It is openjdk-8-jdk issue.
But, it breaks halyard deployment and installation.

To prevent the issue, it will be better to have option to disable the dependency installation.
```
The following packages have unmet dependencies:
 openjdk-8-jdk : Depends: openjdk-8-jre (= 8u141-b15-2~14.04)
                 Depends: openjdk-8-jdk-headless (= 8u141-b15-2~14.04)
E: Unable to correct problems, you have held broken packages.
~
Install finished with exit code (100)... cleaning up
```

2 of Installer.sh have the dependency installation issue.
I can use the following command for installing new halyard.
```
curl -O https://raw.githubusercontent.com/spinnaker/halyard/master/install/Installer.sh

sed -i "s/^install_java/#install_java/g" Installer.sh 
#export DOWNLOAD="test"
export RELEASE_TRACK="stable"
bash Installer.sh
```
However, the "hal deploy apply" still has the issue. Because, generated install.sh script ,which is located at /home/ubuntu/.hal/default/install.sh, is not allowing to disable the dependency installation.

If we can setup export PREPARE_ENVIRONMENT="false", we can handle this issue based on the code.

I suggest you it will be better to change the default value of bindings.put("prepare-environment", "true"); to "false" at https://github.com/spinnaker/halyard/blob/02864f7d407eb3789562b2c79bf9bc09684326ed/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/debian/LocalDebianServiceProvider.java

The first installation script(https://raw.githubusercontent.com/spinnaker/halyard/master/install/Installer.sh) can handle the dependency installation.


aritratony:
Why not just check if a JAVA_HOME is present, and the major version matches the requirements? If it's not present, then trigger an install.

cookingcodewithme:
You can find the following code from https://raw.githubusercontent.com/spinnaker/halyard/master/install/Installer.sh.

Please read the comment inside of the function. 
I think today's issue is known issue. At that time, the way to fix was adding "$DOWNLOAD" variable.

I think It will be better to remove the java installation logic from the function both installation script and deployment script.

```
function install_java() {
  if [ -z "$DOWNLOAD" ]; then
    apt-get install -y --force-yes openjdk-8-jdk

    # https://bugs.launchpad.net/ubuntu/+source/ca-certificates-java/+bug/983302
    # It seems a circular dependency was introduced on 2016-04-22 with an openjdk-8 release, where
    # the JRE relies on the ca-certificates-java package, which itself relies on the JRE. D'oh!
    # This causes the /etc/ssl/certs/java/cacerts file to never be generated, causing a startup
    # failure in Clouddriver.
    dpkg --purge --force-depends ca-certificates-java
    apt-get install ca-certificates-java
  elif [[ "x`java -version 2>&1|head -1`" != *"1.8.0"* ]]; then
    echo "you must manually install java 8 and then rerun this script; exiting"
    exit 13
  fi
}```

aritratony:
Yes, I have seen this before. I am wondering if spinnaker has a hard dependency on openjdk? If a server already has some flavor of java 8 jdk installed, why would spinnaker try to force install openjdk?

lwander:
Thanks for the PR :)

There is no hard dependency on openjdk - a check like this: https://github.com/spinnaker/halyard/blob/master/install/debian/InstallHalyard.sh#L148-L157 is probably preferred to requiring users to toggle an env var.

cookingcodewithme:
FYI, I have confirmed that new openjdk(8u141-b15-3~14.04) was uploaded 2 hours ago. 
The issue wasn't reproduced anymore.

https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa

PR #629, opened 2017-07-26
Allow CLI to configure location for spinnaker deployment.

johnrengelman:


lwander:
Oh and please add Target to the AUTHORS file :)

edwinavalos:
I probably need to do some more review of the hal config file, but going to bring this situation up.

We are going to be working on making a k8s deploy, be able to accept a --bootstrap-only flag of some sort. Where that deployment would end up being Spinnaker can be deployed to, but not an account that Spinnaker has access to manipulate resources in.

We'll be adding say two accounts 

  ```
  kubernetes:
    enabled: true
    accounts:
    - name: cluster1
      requiredGroupMembership: []
      context: null
      cluster: null
      user: null
      namespaces: []
      omitNamespaces: []
      dockerRegistries:
      - accountName: registry-name
        namespaces: []
      kubeconfigFile: /root/.kube/config
  - name: cluster2
      requiredGroupMembership: []
      context: null
      cluster: null
      user: null
      namespaces: []
      omitNamespaces: []
      dockerRegistries:
      - accountName: registry-name
        namespaces: []
      kubeconfigFile: /root/.kube/config
    primaryAccount: inception
```

If we take your advice, should we move both flags to result in something like this? cluster1 will be a bootstrap only cluster and we do not want spinnaker to have access to it.
```
  kubernetes:
    enabled: true
    **bootstrap-account: cluster1**
    **bootstrap-location: spinnaker-lab**
    accounts:
    - name: cluster1
      requiredGroupMembership: []
      context: null
      cluster: null
      user: null
      namespaces: []
      omitNamespaces: []
      dockerRegistries:
      - accountName: registry-name
        namespaces: []
      kubeconfigFile: /root/.kube/config
  - name: cluster2
      requiredGroupMembership: []
      context: null
      cluster: null
      user: null
      namespaces: []
      omitNamespaces: []
      dockerRegistries:
      - accountName: registry-name
        namespaces: []
      kubeconfigFile: /root/.kube/config
    primaryAccount: cluster2
```

lwander:
We chatted about this offline - I think we came to the agreement that the `DeploymentEnvironment` makes the most sense given that it already houses deployment-specific details: https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java

edwinavalos:
Because of the differences in approaches we will probably close this PR and open a new one with the changes we chatted about here.

johnrengelman:
Closing this for now.

PR #635, opened 2017-08-01
feat(deploy): Disable nu-orca instances during deploy

lwander:
This gracefully handles a mix of nu & old orcas by only disabling (and then flagging for either deletion or scaling down depending on the age) orca instances that implement the new `/admin/instance/enabled` endpoint. All other orcas are untouched & reported as "unkillable". See sample behavior here: 
hal-nu-orca

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
@@ -21,14 +21,18 @@ import com.netflix.spinnaker.halyard.core.DaemonResponse;
 import com.netflix.spinnaker.halyard.core.RemoteAction;
 import com.netflix.spinnaker.halyard.core.error.v1.HalException;
 import com.netflix.spinnaker.halyard.core.problem.v1.Problem;
+import com.netflix.spinnaker.halyard.core.problem.v1.ProblemBuilder;
 import com.netflix.spinnaker.halyard.core.tasks.v1.DaemonTaskHandler;
 import com.netflix.spinnaker.halyard.deploy.services.v1.GenerateService.ResolvedConfiguration;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.RunningServiceDetails;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSettings;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.*;
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.ConfigSource;
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.LogCollector;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.OrcaService.Orca;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.OrcaService.Orca.ActiveExecutions;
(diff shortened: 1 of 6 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
```diff
@@ -570,6 +604,36 @@ default RunningServiceDetails getRunningServiceDetails(AccountDeploymentDetails<
...
+    List<String> command = KubernetesProviderUtils.kubectlPortForwardCommand(details,
+        namespace,
+        instanceId,
+        targetPort,
+        localPort);
+    JobRequest request = new JobRequest().setTokenizedCommand(command);
+    String jobId = getJobExecutor().startJob(request);
+
+    // Wait for the proxy to spin up.
+    DaemonTaskHandler.safeSleep(TimeUnit.SECONDS.toMillis(5));
```
jtk54: Maybe poll this a few times rather than waiting and trying once.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
```diff
@@ -189,9 +198,16 @@ public RemoteAction deploy(DistributedServiceProvider<T> serviceProvider,
...
 
-    reapOrcaServerGroups(deploymentDetails, runtimeSettings, serviceProvider.getDeployableService(SpinnakerService.Type.ORCA));
+    DistributedService<Orca, T> orca = serviceProvider.getDeployableService(SpinnakerService.Type.ORCA);
+    Set<Integer> unknownVersions = reapOrcaServerGroups(deploymentDetails, runtimeSettings, orca);
     reapRoscoServerGroups(deploymentDetails, runtimeSettings, serviceProvider.getDeployableService(SpinnakerService.Type.ROSCO));
 
+    if (!unknownVersions.isEmpty()) {
+      String versions = String.join(", ", unknownVersions.stream().map(orca::getVersionedName).collect(Collectors.toList()));
+      throw new HalException(new ProblemBuilder(Problem.Severity.ERROR, "The following orca versions (" + versions + ") could not safely be drained of work.")
+          .setRemediation("Please make sure that no pipelines are running, and manually destroy the server groups at those versions.").build());
```
jtk54: What is the correct user action if Halyard can't reap the old Orcas? Is the remediation is to manually drain the Orcas, destroy them, and then retry the deploy again?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
lwander: Tricky thing is that there is nothing clean to poll on (no clear success condition). If we try to open a connection to the local port, it's possible the container isn't accepting requests yet, or that there is already something else bound to the local port (false negative and false positive response respectively).

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/deployment/v1/DistributedDeployer.java
lwander: It should be exactly that help text, namely making sure that you aren't running pipelines, and then deleting those orcas by hand. Checking for no running pipelines is kinda race-condition prone since there is no guarantee that those nu-orca nodes won't pick up work again. The safest thing to do is to do this by hand.

PR #640, opened 2017-08-02
Added options for a configurable namespace.

shazy792:
This is the direct fix for concerns raised in https://github.com/spinnaker/halyard/pull/629

h/t @markus-silpala-tgt

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
@@ -94,6 +94,7 @@ public class DeploymentEnvironment extends Node {
   private String accountName;
   private Consul consul = new Consul();
   private Vault vault = new Vault();
+  private String location = "spinnaker";
 
   @Data
   public static class Consul {
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
@@ -94,7 +94,7 @@ public class DeploymentEnvironment extends Node {
   private String accountName;
   private Consul consul = new Consul();
   private Vault vault = new Vault();
-  private String location = "spinnaker";
+  private String location;
 
   @Data
   public static class Consul {
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
@@ -96,14 +96,17 @@ public class KubernetesAccount extends Account implements Cloneable {
   public void makeBootstrappingAccount(ArtifactSourcesConfig artifactSourcesConfig) {
     super.makeBootstrappingAccount(artifactSourcesConfig);
 
+    // We may want this to use the value from the deployment config rather than the hard-code.
+    String location = "spinnaker";
+
     // These changes are only surfaced in the account used by the bootstrapping clouddriver,
     // the user's clouddriver will be unchanged.
-    if (!namespaces.isEmpty() && !namespaces.contains("spinnaker")) {
-      namespaces.add("spinnaker");
+    if (!namespaces.isEmpty() && !namespaces.contains(location)) {
+      namespaces.add(location);
     }
 
-    if (!omitNamespaces.isEmpty() && omitNamespaces.contains("spinnaker")) {
-      omitNamespaces.remove("spinnaker");
+    if (!omitNamespaces.isEmpty() && omitNamespaces.contains(location)) {
+      omitNamespaces.remove(location);
     }
   }
 }
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentEnvironmentService.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentEnvironmentService.java
@@ -17,10 +17,9 @@
 
 package com.netflix.spinnaker.halyard.config.services.v1;
 
-import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentConfiguration;
-import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentEnvironment;
-import com.netflix.spinnaker.halyard.config.model.v1.node.NodeFilter;
+import com.netflix.spinnaker.halyard.config.model.v1.node.*;
 import com.netflix.spinnaker.halyard.core.problem.v1.ProblemSet;
+import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -36,6 +35,12 @@ public class DeploymentEnvironmentService {
 
   @Autowired
   private ValidateService validateService;
+
+  @Autowired
+  private AccountService accountService;
+
+  @Autowired
+  private ProviderService providerService;
 
   public DeploymentEnvironment getDeploymentEnvironment(String deploymentName) {
     NodeFilter filter = new NodeFilter().setDeployment(deploymentName).setDeploymentEnvironment();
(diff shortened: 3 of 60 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
```diff
@@ -94,6 +94,7 @@ public static Size fromString(String name) {
   private String accountName;
   private Consul consul = new Consul();
   private Vault vault = new Vault();
+  private String location = "spinnaker";
```
lwander: I made a comment of this in the other PR, but every provider needs to implement some interface that returns a default deployment location that is referenced/read when you configure which account you're deploying to. The reason being, "spinnaker" only makes sense for k8s - when deploying to GCE you'd need something like "us-central1-f".

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -96,14 +96,17 @@ public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
   public void makeBootstrappingAccount(ArtifactSourcesConfig artifactSourcesConfig) {
     super.makeBootstrappingAccount(artifactSourcesConfig);
 
+    // We may want this to use the value from the deployment config rather than the hard-code.
+    String location = "spinnaker";
```
lwander: This will definitely need to come from the deployment environment config.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesFiatService.java
```diff
@@ -40,11 +40,10 @@ public DistributedLogCollector getLogCollector() {
     return getLogCollectorFactory().build(this);
   }
 
-
   @Override
   public Settings buildServiceSettings(DeploymentConfiguration deploymentConfiguration) {
     Settings settings = new Settings();
-    String location = "spinnaker";
+    String location = deploymentConfiguration.getDeploymentEnvironment().getLocation();
```
lwander: Sorry for all the hard-coded constants; thank you for fixing :)

markus-silpala-tgt: We accept thank-you payment in kudos, tacos, and beer. ;-)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
markus-silpala-tgt: @lwander Are you recommending that we:
* leave it unset here;
* set the "spinnaker" default only in the k8s and openstack providers;
* set sane defaults in all the other providers?

lwander: Yes exactly. You will only need to give this treatment to k8s for now, the other cloud providers won't call `getLocation()` until they've been given the same refactor as you have provided here. I would probably have the semantics be defined in `getLocation()` which only looks up a statically defined method on the cloud provider of the account used to deploy to when `String location` is null (hasn't been user defined yet).

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesFiatService.java
lwander: heh ;)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
shazy792: This _spinnaker_ value is the actual key that is being used in the config yaml so we might not want to change that.

lwander: What do you mean by "might not want to change that"?

shazy792:
@lwander Could you take an other look at this, Thanks!

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
lwander: I think this still needs to be resolved

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/DeploymentEnvironmentService.java
```diff
@@ -56,14 +62,26 @@ public DeploymentEnvironment getDeploymentEnvironment(String deploymentName) {
 
   public void setDeploymentEnvironment(String deploymentName, DeploymentEnvironment newDeploymentEnvironment) {
     DeploymentConfiguration deploymentConfiguration = deploymentService.getDeploymentConfiguration(deploymentName);
+
+    String defaultLocation = lookupDefaultLocation(deploymentName, newDeploymentEnvironment.getAccountName());
+    String location = newDeploymentEnvironment.getLocation();
+    newDeploymentEnvironment.setLocation(StringUtils.isEmpty(location) ? defaultLocation : location);
```
lwander: I don't think this is the right place to set this. Once a user first chooses a provider they will be locked into that providers choice of location until they manually set it. I'd prefer that the location doesn't have to be modified by the user unless they explicitly start making edits to it. 

Imagine a user selects a K8S account to deploy to, immediately changes their mind and switches to a GCE account. The deploy would fail for the weird reason that there is no "spinnaker" zone. On the other hand, if they'd manually edited the location to be something like "spinnaker-prod", and then changed to a GCE account and saw that "spinnaker-prod" is not an available zone they'd at least understand where that failure is coming from, and what they edited to cause it.

shazy792: If we want to change the location whenever the account changes then we will have to do it on the CLI side of things, Do you have a better place in mind for the location change to take place or should we just make the change on the CLI side of things.

lwander: It's OK to change the location with the account - but if the user hasn't overriden the location yet they shouldn't have to set it each time after the first edit to the account name.

markus-silpala-tgt: @lwander They don't have to repeatedly set it. Once set via CLI, it retains that value until another CLI override changes it. The default only kicks in if (A) the deployment environment doesn't have a location set, and (B) the provider defines a default location.

The case of changing account types from k8s and having it fail because of "spinnaker" is definitely a confusing bit of UX. We mulled that a little and decided to punt it to when we implement the validator.

What would be the "correct" behavior when:
* User sets account type to k8s without setting location (which sets default for that provider)
* User changes account type to openstack

We considered the option of only setting the default location at deploy time, in each micro service (essentially in every place that "spinnaker" was hardcoded before). Maybe that's the best choice after all. The main issue we saw with it was that the default namespace doesn't appear in the config; that feels inconsistent and obscures useful information from the user, but perhaps that's actually the smaller issue here.

lwander: Sorry, I wasn't clear - I meant each consecutive change that switches the provider with the account change.

In your example, the correct behavior would be to use the default location for openstack during deploy time.

I see why you want to keep the namespace explicit - but if a user wants to make the location explicit they are welcome to do so in the config, and once they do, the location sticks.

There are two reasons why I don't like having the location set on the first edit to DeploymentEnvironment:

1. The way it's currently configured lends itself to weird side-effects implicitly tied to the order your carry out certain operations. If you first decide to deploy to GCE, then K8s, you're deploying to a different namespace than you would if you had just selected K8s in the first place.

2. While we should have proper validation of the deployment environment that checks that the specified location is valid, not all providers (e.g. Kubernetes) will be able to catch accidents where a user doesn't update the location. Techincally "us-central1-f" is a valid k8s namespace, but it's kind-of sloppy.

markus-silpala-tgt: @shazy792 Looks like your original proposal is going to be the winner after all.

shazy792:
Concerns fixed in https://github.com/spinnaker/halyard/pull/652

Issue #649, opened 2017-08-09
Overriding settings.js on Kubernetes setups breaks the deck service


cristim:
This was first reported as a comment in #638 and #639 but it seems to be unrelated to those.

**Steps to reproduce**

```
cp ~/.hal/default/staging/settings.js ~/.hal/default/profiles/
```
Then modify it slightly to be able to tell the difference from the initial version.
```
hal deploy apply  --service-names=deck
```

**Results**

The Deck service returns 403 and fails to serve the expected UI, rendering Spinnaker unavailable to human users interacting to it via GUI.

**Expected results**

The Deck service would load as expected, and have the customized settings applied.

**Workaround**

In the Kubernetes Secrets you can manually edit the `hal-spin-deck0-*-profiles` entry and replace the existing base64 data with the base64 of your desired configuration. Once replaced you can visualize the file in the Kubernetes dashboard to confirm it matches what you expect. 

This works but it will be reverted by the next halyard configuration change.

**Detailed information**

The deck Kubernetes pod is missing all the content of `/opt/deck/html` as soon as the changed `settings.js` is deployed. This could be seen on the pods using the following `kubectl exec` command:

```kubectl --namespace spinnaker exec -it spin-deck-v013-6cbcv -c spin-deck bash``` (make sure to use your pod name)

```
# ls -la /opt/deck/html                                                                               
total 0
drwxrwxrwt 3 www-data root 100 Aug  8 15:25 .
drwxr-xr-x 1 root     root  67 Aug  8 15:25 ..
drwxr-xr-x 2 www-data root  60 Aug  8 15:25 ..8988_08_08_15_25_03.760686059
lrwxrwxrwx 1 www-data root  31 Aug  8 15:25 ..data -> ..8988_08_08_15_25_03.760686059
lrwxrwxrwx 1 www-data root  18 Aug  8 15:25 settings.js -> ..data/settings.js
```

I extracted this from the apache log, which complains about the missing index.html and returns a 403:
```
[Tue Aug 08 15:58:30.163804 2017] [authz_core:debug] [pid 4212:tid 140059814586112] mod_authz_core.c(809): [client 100.109.113.64:32830] AH01626: authorization result of Require all granted: granted
[Tue Aug 08 15:58:30.163839 2017] [authz_core:debug] [pid 4212:tid 140059814586112] mod_authz_core.c(809): [client 100.109.113.64:32830] AH01626: authorization result of <RequireAny>: granted
[...]

PR #651, opened 2017-08-09
fix(deploy): Correctly mount custom settings.js in k8s

lwander:
@cristim FYI


lwander:
Prior to this settings.js was being mounted in /opt/deck/html directly, while the docker container expects to find setting.js in /opt/spinnaker/config and copy it into /opt/deck/html. The reason for the copy instead of a direct mount is that when k8s mounts a file in a directory, it overwrites the rest of the directory. Doing so overwrites all generated html and js when a custom settings.js is supplied.

lwander:
https://github.com/spinnaker/halyard/issues/649

PR #652, opened 2017-08-09
Dynamic deploy namespace

shazy792:
This PR is the fix for concerns raised in https://github.com/spinnaker/halyard/pull/640

PR #653, opened 2017-08-10
Dynamic Deploy Location

shazy792:
Same changes as https://github.com/spinnaker/halyard/pull/652


shazy792:
@lwander Thought that the branch name was a problem with the travis build but its not, do you mind looking into this?

lwander:
Could you try squashing your commits? I've seen similar things happen when there are too many commits between what's released & what's being added.

PR #660, opened 2017-08-15
Validate all Providers in ProvidersSpec.

markus-silpala-tgt:
READ FIRST.

I beefed up the ProvidersSpec to validate that _all_ current providers show up as children in the iterator. 

In addition, I restructured the spec to be a little more groovy and a little less iffy. But I couldn't decide _which_ of two structures to go with. So I'm interested in @lwander's opinion and preference. See the comments above each method for high-level descriptions of why you might prefer one or the other.

### DIFF
--- a/halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ProvidersSpec.groovy
+++ b/halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ProvidersSpec.groovy
@@ -17,48 +17,79 @@
 package com.netflix.spinnaker.halyard.config.model.v1
 
 import com.netflix.spinnaker.halyard.config.model.v1.node.Providers
+import com.netflix.spinnaker.halyard.config.model.v1.providers.appengine.AppengineProvider
+import com.netflix.spinnaker.halyard.config.model.v1.providers.aws.AwsProvider
 import com.netflix.spinnaker.halyard.config.model.v1.providers.azure.AzureProvider
+import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSProvider
 import com.netflix.spinnaker.halyard.config.model.v1.providers.dockerRegistry.DockerRegistryProvider
 import com.netflix.spinnaker.halyard.config.model.v1.providers.google.GoogleProvider
 import com.netflix.spinnaker.halyard.config.model.v1.providers.kubernetes.KubernetesProvider
+import com.netflix.spinnaker.halyard.config.model.v1.providers.openstack.OpenstackProvider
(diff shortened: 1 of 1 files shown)


lwander:
Nice work, I generally prefer `@Unroll` because as you pointed out, the messages point exactly to your failure. Up to you on what you think is cleanest/best, though.

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ProvidersSpec.groovy
```diff
@@ -17,48 +17,47 @@
...
+    setup:
+    def iterator = new Providers().getChildren()
 
-      child = iterator.getNext()
+    when:
+    List actualProviders = []
+    def child
+    while (child = iterator.next) {
+      actualProviders << child.class
     }
```
markus-silpala-tgt: @lwander A question for you: is there a reason that getChildren returns a hand-rolled iterator instead of a Iterable of some sort? If it were Iterable then lines 37-45 here could have been just two lines, like this:

```
when:
List actualProviders = new Providers().children.collect { child -> child.class }
```

Not a big deal either way; but when we're using Spock it's nice to be able to _really_ use groovy, too.

lwander: Ah right - nothing stopping us from fulfilling the `iterator` interface, last I checked `remove` would cause problems for us. What's important is that we keep `next(NodeFilter)` around.

lwander:
Looks good! I can squash on the merge via github - ready to merge?

markus-silpala-tgt:
Yup. Thanks!

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ProvidersSpec.groovy
markus-silpala-tgt: Ok. I might poke around and see how clean it is to make the switch. The other thing is that the name getChildren implies returning a collection rather than an iterator. I'm also not sure how to groovily use the collection operators (collect, each, etc) on just an iterator. Sounds like an interesting space to explore during pockets of free time.

markus-silpala-tgt:
It's running again because I just merged master into it and pushed again.

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ProvidersSpec.groovy
lwander: That's a very good point

Issue #659, opened 2017-08-14
Fix serialization of error messages


lwander:
The daemon failed to send a failure resulting in the following stack trace

```
java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 48 path $.
	at com.google.gson.stream.JsonReader.nextString(JsonReader.java:837)
	at com.google.gson.internal.bind.TypeAdapters$13.read(TypeAdapters.java:358)
	at com.google.gson.internal.bind.TypeAdapters$13.read(TypeAdapters.java:346)
	at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.read(TypeAdapterRuntimeTypeWrapper.java:40)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:187)
	at com.google.gson.internal.bind.MapTypeAdapterFactory$Adapter.read(MapTypeAdapterFactory.java:145)
	at com.google.gson.Gson.fromJson(Gson.java:810)
	at com.google.gson.Gson.fromJson(Gson.java:775)
	at retrofit.converter.GsonConverter.fromBody(GsonConverter.java:63)
	at retrofit.RetrofitError.getBodyAs(RetrofitError.java:137)
	at retrofit.RetrofitError.getBody(RetrofitError.java:112)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:664)
	at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:690)
	at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:155)
	at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:704)
	at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:690)
[...]

PR #662, opened 2017-08-15
fix(deploy): Fix deserialization bug showing strange errors

lwander:
When kubernetes returns a 503 for a non-existent service, retrofit tries to cast the error message into that of the original response body type.


lwander:
Closes: https://github.com/spinnaker/halyard/issues/659

PR #663, opened 2017-08-15
CLI Specs: first cut

markus-silpala-tgt:
Here's a proposed structure for wrapping some specs/tests/documentation around how all the subcommands are structured. While making it I discovered and fixed a duplicated =subcommands= member in HalConfig.

Most of our little team at Target has given this a thumbs-up. Looking to @lwander for opinions and suggestions.

### DIFF
--- a/halyard-cli/src/test/groovy/com/netflix/spinnaker/halyard/cli/command/v1/CommandTreeSpec.groovy
+++ b/halyard-cli/src/test/groovy/com/netflix/spinnaker/halyard/cli/command/v1/CommandTreeSpec.groovy
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2017 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1
+
+import com.netflix.spinnaker.halyard.cli.command.v1.AdminCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.BackupCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.ConfigCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.DeployCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.HalCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.NestableCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.TaskCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.VersionCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.DeploymentEnvironmentCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.EditConfigCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.FeaturesCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.GenerateCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.MetricStoresCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.PersistentStorageCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.SecurityCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.VersionConfigCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.ci.CiCommand
+import com.netflix.spinnaker.halyard.cli.command.v1.config.providers.ProviderCommand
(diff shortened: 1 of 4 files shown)


halyard-cli/src/test/groovy/com/netflix/spinnaker/halyard/cli/command/v1/CommandTreeSpec.groovy
```diff
@@ -0,0 +1,122 @@
...
+import com.netflix.spinnaker.halyard.cli.command.v1.config.security.ui.UiSecurityCommand
+import spock.lang.Specification
+import spock.lang.Unroll
+
+import static com.netflix.spinnaker.halyard.cli.command.v1.config.security.authn.AuthnMethodEnableDisableCommandBuilder.AuthnMethodEnableDisableCommand
+
+class CommandTreeSpec extends Specification {
+
+  /**
+   * Documents and validates the tree structure of the hal command and all of its subcommands.
```
lwander: I'm wondering where this is useful - it feels like this will be just another thing that has to be touched everytime someone wants to add a command. This is kind-of my frustration with unit tests in general (although I know that's sort of a heretical stance) where they really just seem to impede velocity even if they've been added with good intention. Of course that's not true of all unit tests.

The tree structure is already documented in `./docs/commands.md`, and (as far as I can tell) it's straight forward enough to add/remove commands that rogue/accidental code changes shouldn't be breaking the command tree.

If the intent is to drive discussion around adding/removing commands by having failing tests force the issue, I feel that the act of having the auto-generated docs change already accomplishes this.

markus-silpala-tgt: No need to worry about heresy: I know there are lots of views on this stuff. Your questions and concerns are legit, and I'll do my best to address them.

I see these as the main drivers for this sort of thing:

* I like to drive every code change with a failing test. Watching the test fail for the right reason, and then watching it go green after the coding, tells me that the code works _and_ that the test actually validates it. Over the years I have spent a lot of time chasing bugs that turn out to have been from a really silly little change which I (or someone) didn't validate because I/they assumed such a simple change would just work. By the time someone realizes it, that bug is several hours or days old and can be very tricky to track down. From that perspective, adding a single row to this table before I add my new command is a really quick step which guards against that kind of bug hunt.

* A test that runs on every build guards against regressions and other accidents. Stuff happens; and if, say a subcommand were dropped or modified accidentally, it might be a while before anyone notices. Again, a quick test case guards against that without adding too much weight (hopefully).

* A spec like =CommandTreeSpec= documents intent: expressing what the code _should_ do instead of only what it _actually_ does. Because =commands.md= is auto-generated, it won't catch erroneous changes.

* Adding subcommands has just enough little details in just enough different places that it's not completely straightforward unless you've done it a few times. My motivation for this came about because we had just begun to add LDAP support and I thought, "boy, it would be nice if we could drive this with tests instead of poking around and copying things and manually testing as we are."

* Test-driving code has a way of uncovering or preventing bugs because it exercises the code in ways that make the bugs much more visible than just running the whole program. Absent this =CommandTreeSpec= suite, I doubt we would have discovered that =HalCommand= duplicates and conceals the =subcommands= field from its base class. Maybe that wouldn't have caused any harm; but if it did, it would likely have been very hard to track down without these unit tests.

Still, I do recognize that my opinions on these things aren't universal. All the benefits above are predicated on the test maintenance being very lightweight, and the meaning of lightweight is subjective. That's okay.

Also: TDD only really shines when it's adopted by the whole team rather than a subset of them. If some people go straight to the code and make functional changes first, they'll perceive the tests as more hassle than value: "I made this little change and now I have to change a test, too!" In my experience that's what drives the feeling you describe, where tests seem like an impediment more than a benefit. Tools that you don't use make your tool belt heavier without making your job easier. Unit tests add the most value when you keep them front of mind; if half the team won't do that it might be best to just remove their weight from the tool belt.

On the other hand, when everyone _does_ agree to drive changes via tests, and to look at tests as the place to find the intent and the proper usage of the functional code, then it becomes uncannily simple to make changes, to isolate bugs, to avoid writing new bugs, and to cross-train new developers. The most magically productive and enjoyable times of my career have been when I worked on teams that did that. Every place I go now, I hope to recreate that level of productivity and excellence, and test-driving code is one of the core ingredients to getting there.

I know this was long-winded, and I apologize. It's a surprisingly deep topic, and I find that when someone hasn't experienced highly-productive TDD firsthand, explaining it is not trivial. I don't know whether I explained it effectively here.

If I haven't, that's okay too. Passionate though my opinions may be, it's more important to find the right practice for the whole community than it is to satisfy my preferences.

Not sure where that leaves us. What do you think?

lwander: Alright - you've convinced me :) Let's give it a shot.

lwander: Let me know if you're ready to merge

markus-silpala-tgt: Awesome. Thank you. Ready any time.

Two things to call out:
* Next time I'm in New York I'd like to buy you a beer or alternate beverage of your choice. #appreciation 
* Please be vigilant and frank for when the tests feel heavy to you. Let's nip that problem in the bud when it comes up.

lwander: Sounds good to me! :) I already owe you a beer for the other code cleanup anyhow

PR #664, opened 2017-08-15
refactor(pipline_templates): update orca profile to pipelineTemplates

idoshamun:
Due to https://github.com/spinnaker/orca/pull/1548, the profile stanza is now
in plural form and not single form as it used to.


lwander:
https://github.com/spinnaker/halyard/pull/665

idoshamun:
@danielpeach yeah, you're totally right

PR #666, opened 2017-08-16
fix(config/orca): append singular form and plural form of pipelineTemplates

idoshamun:
To support old versions of orca, it is needed to append also the singular form and not just the plural.

#664


idoshamun:
Please note that I didn't change orca-bootstrap profile as older version don't enable it by default so there is no need to add backward compatibility there, if needed let me know

lwander:
I've added to the older orca-bootstrap profile as well because even if it's not added by default it will still be picked up in the `orca.yml` profile

PR #673, opened 2017-08-21
Add authn Ldap settings to halyard

loomisdf:
I've added settings and corresponding tests to enable ldap authentication.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -171,6 +171,10 @@
  * [**hal config security api ssl edit**](#hal-config-security-api-ssl-edit)
  * [**hal config security api ssl enable**](#hal-config-security-api-ssl-enable)
  * [**hal config security authn**](#hal-config-security-authn)
+ * [**hal config security authn ldap**](#hal-config-security-authn-ldap)
+ * [**hal config security authn ldap disable**](#hal-config-security-authn-ldap-disable)
+ * [**hal config security authn ldap edit**](#hal-config-security-authn-ldap-edit)
+ * [**hal config security authn ldap enable**](#hal-config-security-authn-ldap-enable)
  * [**hal config security authn oauth2**](#hal-config-security-authn-oauth2)
  * [**hal config security authn oauth2 disable**](#hal-config-security-authn-oauth2-disable)
  * [**hal config security authn oauth2 edit**](#hal-config-security-authn-oauth2-edit)
@@ -3344,8 +3348,79 @@ hal config security authn [parameters] [subcommands]
  * `--no-validate`: (*Default*: `false`) Skip validation.
 
 #### Subcommands
+ * `ldap`: Configure the ldap method for authenticating.
  * `oauth2`: Configure the oauth2 method for authenticating.
  * `saml`: Configure the saml method for authenticating.
+
+---
+## hal config security authn ldap
+
+Configure the ldap method for authenticating.
+
+#### Usage
+```
+hal config security authn ldap [parameters] [subcommands]
+```
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+
+#### Subcommands
+ * `disable`: Set the ldap method as disabled
+ * `edit`: Configure authentication using a LDAP identity provider.
+ * `enable`: Set the ldap method as enabled
+
+---
+## hal config security authn ldap disable
+
+Set the ldap method as disabled
+
+#### Usage
+```
+hal config security authn ldap disable [parameters]
+```
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+
+
+---
+## hal config security authn ldap edit
+
+Edit the Ldap authentication method
+
+#### Usage
+```
+hal config security authn ldap edit [parameters]
+```
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+ * `--url`: ldap:// or ldaps:// url of the LDAP server
+ * `--user-dn-pattern`: User DN pattern, Base + Unique e.g. uid={0},ou=users,dc=my-org,dc=com
+ * `--user-search-base`: Used when a simple DN pattern is not possible
+ * `--user-search-filter`: Used When a simple DN pattern is not possible
+ 
+Go [here](https://www.spinnaker.io/setup/security/authentication/ldap/) for more information about Ldap parameters and spinnaker.
+
+
+---
+## hal config security authn ldap enable
+
+Set the ldap method as enabled
+
+#### Usage
+```
+hal config security authn ldap enable [parameters]
+```
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+
 
 ---
 ## hal config security authn oauth2
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/ldap/EditLdapCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/ldap/EditLdapCommand.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1.config.security.authn.ldap;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.converters.URIConverter;
+import com.netflix.spinnaker.halyard.cli.command.v1.config.security.authn.AbstractEditAuthnMethodCommand;
+import com.netflix.spinnaker.halyard.config.model.v1.security.AuthnMethod;
+import com.netflix.spinnaker.halyard.config.model.v1.security.Ldap;
+import lombok.Getter;
+
+import java.net.URI;
+
+public class EditLdapCommand extends AbstractEditAuthnMethodCommand<Ldap> {
+
+  @Getter
+  private String shortDescription = "Configure authentication using a LDAP identity provider.";
+
+  @Getter
+  private String longDescription = String.join(" ", "Lightweight Directory Access Protocol (LDAP)",
+      "is a standard way many organizations maintain user credentials and group memberships.",
+      "Spinnaker uses the standard “bind” approach for user authentication.",
+      "This is a fancy way of saying that Gate uses your username and password to login to the LDAP server,",
+      "and if the connection is successful, you’re considered authenticated.");
+
+  @Getter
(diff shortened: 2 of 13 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ldap.java
```diff
@@ -0,0 +1,40 @@
...
+  private final Method method = Method.LDAP;
+  private final String nodeName = "ldap";
+
+  private String url;
+  private String userDnPattern;
+  private String userSearchBase;
+  private String userSearchFilter;
+
+  @Override
+  public void accept(ConfigProblemSetBuilder psBuilder, Validator v){
```
lwander: Missing space before {

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
```diff
@@ -0,0 +1,51 @@
...
+
+    if (StringUtils.isEmpty(ldap.getUrl())) {
+      p.addProblem(Problem.Severity.ERROR, "LDAP url must be provided.");
+    }
+
+    if(StringUtils.isEmpty(ldap.getUserDnPattern())) {
+      p.addProblem(Problem.Severity.ERROR, "LDAP User Dn Pattern must be provided.");
+    }
+
+    if(ldap.getUserSearchBase() != null && ldap.getUserSearchBase().equals("")) {
```
lwander: Why not `StringUtils.isEmpty()`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ldap.java
```diff
@@ -0,0 +1,40 @@
...
+package com.netflix.spinnaker.halyard.config.model.v1.security;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode(callSuper = true)
+@Data
+public class Ldap extends AuthnMethod{
```
lwander: Missing space before {

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/LdapConfig.java
```diff
@@ -0,0 +1,47 @@
...
+  String userSearchFilter;
+
+  public LdapConfig(Security security) {
+    if (!security.getAuthn().getLdap().isEnabled()) {
+      return;
+    }
+
+    Ldap ldap = security.getAuthn().getLdap();
+
+    this.enabled = ldap.isEnabled();
```
lwander: So far in halyard the internal constructs generally map quite identically to those that are generated. Would you be able to just use the `Ldap` class here?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
```diff
@@ -0,0 +1,51 @@
...
+@Component
+public class LdapValidator extends Validator<Ldap> {
+
+  @Override
+  public void validate(ConfigProblemSetBuilder p, Ldap ldap) {
+    if (!ldap.isEnabled()) {
+      return;
+    }
+
+    if (StringUtils.isEmpty(ldap.getUrl())) {
```
lwander: You could validate that this is a URL here.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
loomisdf: The field can be null, but can't be an empty string. isEmpty checks for null and empty strings.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
loomisdf: I'll have to do some testing and see how to validate our ldap url.

loomisdf: Ok, I can validate the url by using LdapURL, but the build throws a bunch of warnings about the class being proprietary.
`23: warning: LdapURL is internal proprietary API and may be removed in a future release`
`import com.sun.jndi.ldap.LdapURL;`
Their code is available [here](https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/com/sun/jndi/ldap/LdapURL.java)

Is it ok to use this class for validation purposes?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/LdapConfig.java
loomisdf: I'm having trouble understanding what you're asking about. Are you asking if it's possible to use `Ldap ldap = new Ldap();`?

loomisdf: I was just following the convention set in the [SamlConfig.java](https://github.com/spinnaker/halyard/blob/2dd661040ee4da79b76aa7cb28e50e194e3e3a99/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SamlConfig.java)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/Ldap.java
loomisdf: Oops, good catch. I'll fix this up along with the other issues.

docs/commands.md
```diff
@@ -3338,10 +3343,79 @@ hal config security authn [parameters] [subcommands]
...
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+
+
+---
+## hal config security authn ldap edit
+
+placeholder text...markus says fix this.
```
MatthewKohner: fix this?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/ldap/EditLdapCommand.java
```diff
@@ -0,0 +1,84 @@
...
+import com.netflix.spinnaker.halyard.config.model.v1.security.Ldap;
+import lombok.Getter;
+
+public class EditLdapCommand extends AbstractEditAuthnMethodCommand<Ldap> {
+
+  @Getter
+  private String shortDescription = "Configure authentication using a LDAP identity provider.";
+
+  @Getter
+  private String longDescription = String.join(" ", "placeholder text...markus says fix this.");
```
MatthewKohner: fix this too.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
markus-silpala-tgt: How about using a URL instead of a String, as in the [SamlValidator](https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/SamlValidator.java#L112-L115)?

Using the same technique, we could validate that the URL uses samls as its protocol.

markus-silpala-tgt: Doh. I do see now that java.net.URL doesn't support the ```saml``` and ```samls``` protocols, at least not by default. Time for a doc dive.

To @loomisdf's original question, I don't think we want to use a proprietary API.

docs/commands.md
```diff
@@ -3338,10 +3343,81 @@ hal config security authn [parameters] [subcommands]
...
+#### Usage
+```
+hal config security authn ldap edit [parameters]
+```
+
+#### Parameters
+ * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--no-validate`: (*Default*: `false`) Skip validation.
+ * `--url`: ldap:// or ldaps:// url of the LDAP server
+ * `--user-dn-pattern`: User Dn pattern, Base + Unique e.g. uid={0},ou=users,dc=my-org,dc=com
```
lwander: DN is capitalized in the below descriptions but not this one.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
```diff
@@ -0,0 +1,58 @@
...
+      return;
+    }
+
+    if (ldap.getUrl() == null) {
+      p.addProblem(Problem.Severity.ERROR, "LDAP url missing.");
+    } else if (ldap.getUrl().getScheme() == null) {
+      p.addProblem(Problem.Severity.ERROR, "LDAP url scheme is missing.");
+    } else if (ldap.getUrl().getPort() == -1) {
+      p.addProblem(Problem.Severity.ERROR, "LDAP url port is undefined");
+    } else if (! ldap.getUrl().getScheme().equalsIgnoreCase("ldaps") && ! ldap.getUrl().getScheme().equalsIgnoreCase("ldap")) {
```
lwander: Why the space after `!`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/LdapValidator.java
lwander: Ah I misread, sounds good

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/LdapConfig.java
lwander: Sorry for the wait. This class looks like a duplicate of Ldap, which makes me wonder why we need it.

markus-silpala-tgt: @lwander I agree that the data within these classes is nearly identical; it feels weird to combine them based on that, if they play different roles within the system. Are there other examples where we use the same class for the halyard-deploy model and for the halyard-config profile?

lwander: I think most do. All of the providers and ci systems should. From a maintainability perspective it feels easier to break them apart when we have to, and keep them identical as long as they conceptually map to the same configuration resource.

markus-silpala-tgt: That certainly makes sense. I guess we just missed that most cases use the same class for both purposes.

_Steps away to review some of the other code and see how it's structured._

markus-silpala-tgt: I'm curious whether the intent of this [SpringConfig class](https://github.com/spinnaker/halyard/blob/master/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringConfig.java) was to contain all the different authn implementations, rather than just the OAuth2 one.

The name (SpringConfig) is way more generic than the role it currently plays.

lwander: Yes I think the intent was to capture Spring-specific config (nested under spring:). if that includes other authn mechanisms those should likely live there too.

edwinavalos:
@loomisdf if you have a second could you rebase and squash the commits on this PR?

loomisdf:
@edwinavalos done

PR #687, opened 2017-09-08
fix(springBoot): Update Gate's Spring Boot upgrade version

ttomsu:


lwander:
1.2.2 is on 0.6.0 - so we should be good

ttomsu:
it assumes all gate versions beyond 0.7 will have boot 1.5.4 - https://github.com/spinnaker/halyard/pull/687/files#diff-97de087204983a5ea7f11564f3b5ceddR87

Issue #680, opened 2017-09-01
Halyard - adding docker repo with self signed cert fails


laurieodgers:
I'm having some issues within the docker container mentioned at https://www.spinnaker.io/setup/install/halyard/

`docker run --name halyard --rm -v ~/.hal:/root/.hal gcr.io/spinnaker-marketplace/halyard:stable`

I am currently attempting to add a local docker repository (v2) which uses a self signed certificate:

```
hal config provider docker-registry account add staging-registry \
    --address localhost \
    --username user \
    --repositories "10.1.1.1:5000/test 10.1.1.1:5000/test2" \
    --password
```

results in the following error:
```
! ERROR Unable to reach repository:
  sun.security.validator.ValidatorException: PKIX path building failed:
  sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid
  certification path to requested target.
```

Is there any way to disable SSL verification? Or any way that I can supply a truststore in order to get communication working between `hal` and our local docker repo?

I'm not sure if this is related, but our docker repo is v2 and supports the catalog endpoint (I have tested with curl) and omitting the --repositories flag results in the following error:
```
Problems in default.provider.dockerRegistry.staging-registry:
! ERROR Your docker registry has no repositories specified, and the
  registry's catalog is empty.
? Manually specify some repositories for this docker registry to
  index.
```

lwander:
1. Self-signed certs: This doesn't work today because the underlying HTTP client needs to be hand-loaded with your custom certs. I don't have the time to implement this in the near future, but if anyone is interested: the relevant code lives here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy#L128. I'm guessing this will help: https://stackoverflow.com/questions/23103174/does-okhttp-support-accepting-self-signed-ssl-certs

2. Catalog: Have you checked that the catalog contains repos with your given credentials? Halyard complains about the empty catalog endpoint because when it comes time to deploy Spinnaker, not having any repositories to deploy images from is confusing. If it's not empty, what's the endpoint your catalog lives at?

Spellchaser:
I believe I have the same kind of error. I'm trying to get Halyard working on a corp network (read firewalls and cert modifications). I've configured my proxies correctly but I'm pretty sure my corp is messing with SSL certs. I've seen something about loading BOMs from disk?

[...]

PR #690, opened 2017-09-11
feat: Added insecure registry support

podnov:
Added --insecure-registry=[true|false] as a command line flag
when adding a docker-registry account.

This is dependent on spinnaker/clouddriver#1887 and will
require a dependency version bump once merged and released.

This fixes spinnaker/halyard#680.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/dockerRegistry/DockerRegistryEditAccountCommand.java
```diff
@@ -88,6 +88,12 @@ protected String getProviderName() {
   )
   private Long cacheIntervalSeconds;
 
+  @Parameter(
+      names = "--insecure-registry",
+      description = DockerRegistryCommandProperties.INSECURE_REGISTRY_DESCRIPTION
```
lwander: Can you make the arity=1 here? This allows you to set the flag back to false.

podnov: I've set arity=1 for both the add and edit command.  Were you also suggesting setting the default value of the flag to false?  I intentionally left that default null in the edit command for the isSet() check.

podnov:
Figured how to point this at the a new version of cloudriver for the api changes. This should be gtg as far as I'm concerned.

podnov:
At kubecon watching the keynote about spinnaker, hoping to get this merged soon.

lwander:
Sorry! nice work

PR #708, opened 2017-09-20
X509 config

omawhite:
adds x509 config enable and edit commands

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/X509.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/X509.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.security;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class X509 extends AuthnMethod {
+
+  private final Method method = Method.X509;
+  private final String nodeName = "x509";
+
+  private String roleOid;
+
+  @Override
+  public void accept(ConfigProblemSetBuilder psBuilder, Validator v ) { v.validate(psBuilder, this); }
+}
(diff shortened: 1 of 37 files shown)


edwinavalos:
@omawhite squash the commits into one, and I'll go ahead and merge.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/X509.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class X509 extends AuthnMethod {
+
+  private final Method method = Method.X509;
+  private final String nodeName = "x509";
+
+  private String roleOid;
```
lwander: I'm surprised there's no need for certs in the config?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/X509Validator.java
```diff
@@ -0,0 +1,38 @@
...
+@Component
+public class X509Validator extends Validator<X509> {
+
+  @Override
+  public void validate(ConfigProblemSetBuilder p, X509 x509) {
+    if (!x509.isEnabled()) {
+      return;
+    }
+
+    if (x509.getRoleOid() != null && (x509.getRoleOid() == "")) {
```
lwander: from apache.lang3 `StringUtils.isEmpty`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/GateProfileFactory.java
```diff
@@ -60,17 +60,11 @@ public void setProfile(Profile profile,
     SpringConfig spring;
     SamlConfig saml;
     LdapConfig ldap;
+    X509Config x509;
 
     GateConfig(ServiceSettings gate, Security security) {
       super(gate);
       server.ssl = security.getApiSecurity().getSsl();
-      if (security.getAuthn().getOauth2().isEnabled()) {
```
edwinavalos: I assume these are taken care of in the two profile factories instead of the original one.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/X509.java
edwinavalos: It uses the cert configs from the api ssl settings. We might want to add a validator step to make sure api ssl is turned on then.

lwander: Ah that's neat. Yeah probably a good idea...

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/X509Validator.java
edwinavalos: This is one of those java things I just get tripped up on a lot.

Basically here I want to validate if they are setting roleOid, that it is not an empty string. But they don't have to set roleOid, so it can be set to null. does the .isEmpty cover that?

lwander: Oh you're right, I'm just not reading carefully. This looks good

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/X509.java
edwinavalos: I'm going to break that into a separate issue, and go ahead and merge as it is here.

PR #711, opened 2017-09-22
docs: link to spinnaker.io docs for halyard

dblackdblack:
It's not obvious that the github docs are non-exhaustive
and that the main docs are actually elsewhere


dblackdblack:
@lwander trivial README.md change: add link from github to spinnaker.io for docs

Issue #683, opened 2017-09-07
Adding container shaping to Halyard deployed replicasets


edwinavalos:
Discussed this with @lwander and we decided that this will be our first approach:

```
deploymentEnvironment:
    size: SMALL
    customSizing:
      clouddriver:
        requests:
          memory: "64Mi"
          cpu: "250m"
        limits:
         memory: "128Mi"
         cpu: "500m"
      echo:
        request:
          memory: "64Mi"
          cpu: "250m"
        limits:
         memory: "128Mi"
         cpu: "500m"
    type: Distributed
    accountName: my-k8s-cluster
    consul:
      enabled: false
    vault:
      enabled: false
    location: my-spinnaker-namespace
```

Any service that doesn't have a customSizing section associated with it will receive no sizing at the moment, and at some point if size: SMALL or size: MEDIUM have presets it would fall back to those.

Open to discussion, but Target will be working on this approach soon.

lwander:
Looking good!

lwander:
Thanks to @greenkiwi we have the following env var `JAVA_DOCKER_MEMORY_OPTS='-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MaxRAMFraction=1 -XshowSettings:vm'` to ensure that the java heap size matches what k8s allocates.

markus-silpala-tgt:
@lwander @edwinavalos Any objections if we leave out the ```size: SMALL``` config line for now? Better to add that to the config only when there's actual functionality to use it, IMO.

lwander:
We can remove it, just be careful that no serialization error occurs that might wipe out a user's halconfig.

markus-silpala-tgt:
I hadn't realized that the size element was already present in the config (though not in the command or parsers). Edwin set me straight. We're not going to touch it for this round; we'll just implement the overrides.

dylanbernhardt:
Possible options for the halyard command:

Parse a YAML file with all the custom sizings
hal config deploy edit --custom-sizing PATH_TO_YAML_FILE_WITH_CUSTOM_SIZING
Use the current convention for deploy edits
hal config deploy edit --custom-sizing-clouddriver-requests-memory MEMORY \
--custom-sizing-clouddriver-requests-cpu CPU \
--custom-sizing-clouddriver-limits-memory MEMORY \
--custom-sizing-clouddriver-limits-cpu CPU
[...]

PR #724, opened 2017-10-09
feat(deployments): allows for custom sizing for kubernetes distrbuted deployments in halconfig

omawhite:
see spinnaker [issue 683](https://github.com/spinnaker/halyard/issues/683)

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2017 Target, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.node;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A CustomSizing is a map of maps where you can hack in provider-specific settings related to instance/container/pod sizes.
+ */
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class CustomSizing implements Map<String, Map> {
+  Map<String, Map> componentSizings = new HashMap<>();
+
+  public static String stringOrNull(Object value) {
+    return value != null ? String.valueOf(value) : null;
+  }
+
+  @Override
+  public int size() {
+    return componentSizings.size();
+  }
+
+  @Override
+  public boolean isEmpty() {
+    return componentSizings.isEmpty();
+  }
+
+  @Override
+  public boolean containsKey(Object key) {
+    return componentSizings.containsKey(key);
+  }
+
+  @Override
+  public boolean containsValue(Object value) {
+    return componentSizings.containsValue(value);
+  }
+
+  @Override
+  public Map get(Object key) {
+    return componentSizings.get(key);
+  }
+
+  @Override
+  public Map put(String key, Map value) {
+    return componentSizings.put(key, value);
+  }
+
+  @Override
+  public Map remove(Object key) {
+    return componentSizings.remove(key);
+  }
+
+  @Override
+  public void putAll(Map m) {
+    componentSizings.putAll(m);
+  }
+
+  @Override
+  public void clear() {
+    componentSizings.clear();
+  }
+
+  @Override
+  public Set<String> keySet() {
+    return componentSizings.keySet();
+  }
+
+  @Override
+  public Collection<Map> values() {
+    return componentSizings.values();
+  }
+
+  @Override
+  public Set<Entry<String, Map>> entrySet() {
+    return componentSizings.entrySet();
+  }
+}
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
@@ -32,11 +32,13 @@ import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergro
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesHttpGetAction;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesImageDescription;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesProbe;
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesResourceDescription;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesSecretVolumeSource;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesTcpSocketAction;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeMount;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeSource;
 import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeSourceType;
+import com.netflix.spinnaker.halyard.config.model.v1.node.CustomSizing;
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentEnvironment;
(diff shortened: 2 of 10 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
```diff
@@ -23,20 +23,7 @@
...
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesHandlerType;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesHttpGetAction;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesImageDescription;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesProbe;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesSecretVolumeSource;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesTcpSocketAction;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeMount;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeSource;
-import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.KubernetesVolumeSourceType;
+import com.netflix.spinnaker.clouddriver.kubernetes.deploy.description.servergroup.*;
```
markus-silpala-tgt: @omawhite I think IntelliJ did this. We probably want to change it back to list all the packages explicitly.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
```diff
@@ -426,6 +405,36 @@ default KubernetesContainerDescription buildContainer(String name, ServiceSettin
...
+    KubernetesResourceDescription requests = new KubernetesResourceDescription();
+    requests.setCpu(stringOrNull(componentSizing.get(resourceType).get("cpu")));
+    requests.setMemory(stringOrNull(componentSizing.get(resourceType).get("memory")));
+    return requests;
+  }
+
+  // This is super-goofy. What can we do differently for this? Anything in the config parsing logic?
+  default String stringOrNull(Object value) {
+    return value != null ? String.valueOf(value) : null;
+  }
```
markus-silpala-tgt: @lwander We'd be interested in your thoughts on this little nugget. The issue arises when the size specification is in form with only numeric characters: ```"1"``` as opposed to ```"1000m"``` or ```"1Gi"```.

We used ```String.valueOf``` to ensure that those cases were parsed out of the config as strings. That, in turn, caused problems when a value was unset (null), because ```String.valueOf(null)``` evaluates to ```"null"``` as a string, which is _not_ what we want to be in place when something isn't set.

Our tactical solution was this ```stringOrNull``` method, which just feels weird. Is there something more framework-y in the wider halyard code base that we might leverage for this? Or a natural util-ish place we could stick this little method instead of implementing it twice in the deployment code? We also considered putting it in the CustomSizing class and calling that from both places where it's needed.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
lwander: Yeah there's some setting to keep those expanded by default

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A CustomSizing is a map of maps where you can hack in provider-specific settings related to instance/container/pod sizes.
+ */
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class CustomSizing implements Map<String, Map> {
```
lwander: Is the intent here just to provide a type alias?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
```diff
@@ -373,14 +359,7 @@ default KubernetesContainerDescription buildContainer(String name, ServiceSettin
     readinessProbe.setHandler(handler);
     container.setReadinessProbe(readinessProbe);
 
-    /* TODO(lwander) this needs work
```
lwander: You guys rock

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
lwander: Yeah agreed this is a little funky - but this is also the first time where the type isn't explicit in the config. Would it be possible to make it clear that component sizing is mapping `string -> string -> string`? Or is the fear that this is too restrictive?

edwinavalos:
@lwander should we be append the JAVA_OPTS in this code, or should we be instructing people that they should introduce that in to their service-settings files for individual services?

lwander:
I think it'd be nice to append this in this code for the k8s deployments. They can always be overridden using custom service settings.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
markus-silpala-tgt: We were concerned about being too restrictive, but at this point it probably makes sense to explicitly make it all maps with strings.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
markus-silpala-tgt: :-)

To be completely honest, we only moved your TODO into the applyCustomSize method. Because we didn't implement handling of the SMALL/MEDIUM/LARGE specifiers, we kept your placeholder in its new home, below.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
markus-silpala-tgt: Heh. Not really.

Initially this class had some structure to it, because it assumed the kubernetes "requests" and "limits" nomenclature. We also extended Node initially, and then removed that after we realized we had nothing to validate because there's no CLI interaction with this data (and we couldn't think of any reasonable validation that wasn't k8s-specific). By the time we removed those things, this had nothing left but the Map implementation; you were the first to realize this. :-)

But now, we've given this class a new reason to exist: it holds our goofy little ```stringOrNull``` method, so that it doesn't have to be duplicated in the manual and orca deployment flows.

(We abandoned the idea of explicitly making this a three-layer map with strings at its leaf nodes. No implementation felt clean, and we also realized that when we add replicaSetSize to it later this week, that will be an integer, at the top tier of the map. So we'd be breaking that constraint ourselves.)

omawhite:
Hey @lwander in reference to this environment variable, we noticed that the change wasn't added until java version 8u131, but it appears that the spinnaker components are using 8u111.  
```
ENV JAVA_DOCKER_MEMORY_OPTS '-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap -XX:MaxRAMFraction=1 -XshowSettings:vm'
```

omawhite:
@lwander so we found that java version 8u131 isn't available until alpine version 3.6, the containers currently run 3.4. So until that changes people will just have to manually set the JAVA_OPTS in addition to setting custom sizing.

omawhite:
squashed commits

lwander:
Good catch @omawhite on the java version.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
lwander: Hmm it might still be valuable to validate even if there is no CLI command to generate entries here - but maybe not. 

Sounds good

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesDistributedService.java
lwander: Heh I saw that after

lwander:
I'm ready to merge when you guys feel ready (or you can go ahead @edwinavalos)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/CustomSizing.java
markus-silpala-tgt: We'd be wide open to it if you can guide us on _what_ to validate in a way that doesn't assume a specific cloud provider, or on a way to detect that we're deploying to k8s and only then do our k8s-specific validations. I would definitely feel better if we validated some basics in the k8s case.

On a later PR, of course. :-)

PR #726, opened 2017-10-12
fix(build) Fix build warnings

markus-silpala-tgt:
Quick fix for a the loudest of the warnings that come up when building halyard.

The deprecated member seems to only be used in one place, where it is set. So removing it should have no functional effect.

The two @EqualsAndHashCode changes only make explicit what is already being built. If these _should_ be calling super, then we have a sneaky bug in the current master.


lwander:
:+1:

PR #729, opened 2017-10-16
feat(bootstrapping): Adds --bootstrap-only flag to accounts, which re…

ttomsu:
…moves the flagged account from the deployed Spinnaker instance.

cc: @edwinavalos This isn't exactly what your group was looking for, but it's a jumping off point to pare down the unnecessary docker registries.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/ContainerAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/ContainerAccount.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.providers.containers;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class ContainerAccount extends Account {
+
+  List<DockerRegistryReference> dockerRegistries = new ArrayList<>();
+
+}
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/DockerRegistryReference.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/DockerRegistryReference.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google, Inc.
+ * Copyright 2017 Google, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License")
  * you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.halyard.config.model.v1.providers.kubernetes;
+package com.netflix.spinnaker.halyard.config.model.v1.providers.containers;
 
 import lombok.Data;
 
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverBootstrapProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverBootstrapProfileFactory.java
@@ -18,21 +18,39 @@
 package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile;
 
 import com.netflix.spinnaker.halyard.config.config.v1.ArtifactSourcesConfig;
-import com.netflix.spinnaker.halyard.config.model.v1.node.*;
+import com.netflix.spinnaker.halyard.config.error.v1.ConfigNotFoundException;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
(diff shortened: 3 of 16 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AbstractEditAccountCommand.java
```diff
@@ -58,6 +58,12 @@
   )
   List<String> requiredGroupMembership;
 
+  @Parameter(
+      names = "--bootstrap-only",
+      description = AccountCommandProperties.BOOTSTRAP_ONLY
+  )
+  Boolean bootstrapOnly = false;
```
lwander: Don't set a default here, otherwise any edit that doesn't include this flag will turn off the --bootstrap-only mode.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/ProviderValidator.java
```diff
@@ -18,20 +18,33 @@
...
     if (n.isEnabled()) {
       if (n.getAccounts() == null || n.getAccounts().isEmpty()) {
         p.addProblem(Problem.Severity.WARNING, "Provider " + n.getNodeName() + " is enabled, but no accounts have been configured.");
       }
+
+      List bootstrapAccounts = (List) n.getAccounts()
+                                       .stream()
+                                       .filter(a -> ((Account) a).getBootstrapOnly())
+                                       .collect(Collectors.toList());
+      if (bootstrapAccounts.size() >= 2) {
```
lwander: You're only validating that there is at most 1 bootstrap-only account per provider. You want to do this check in the `Providers` class and validate across all accounts

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverProfileFactory.java
```diff
@@ -34,10 +38,32 @@ public SpinnakerArtifact getArtifact() {
...
+    Provider provider;
+    NodeIterator ni = clonedProviders.getChildren();
+    while((provider = (Provider) ni.getNext()) != null) {
+      Optional<Account> account = provider.getAccounts()
+                                          .stream()
+                                          .filter(a -> ((Account) a).getBootstrapOnly())
+                                          .findFirst();
+      // Silly Java won't let me do this in the above chained call way.
+      if (account.isPresent()) {
+        provider.getAccounts().remove(account.get());
```
lwander: You can with `.map` on the optional

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverBootstrapProfileFactory.java
```diff
@@ -62,16 +69,26 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
...
-      if (child.providerType() != Provider.ProviderType.DOCKERREGISTRY && !isBootstrappingAccount) {
-        child.setEnabled(false);
+      if (provider.providerType() != Provider.ProviderType.DOCKERREGISTRY && !isBootstrappingProvider) {
+        provider.setEnabled(false);
       }
 
-      child = (Provider) children.getNext();
+      if (isBootstrappingProvider) {
+        List bootstrapAccounts = (List) provider.getAccounts()
+                                                .stream()
```
lwander: The rest of the project aligns these call chains two indent levels rather than against the first method call.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AccountCommandProperties.java
```diff
@@ -19,4 +19,5 @@
 
 public class AccountCommandProperties {
   public static final String REQUIRED_GROUP_MEMBERSHIP_DESCRIPTION = "A user must be a member of at least one specified group in order to make changes to this account's cloud resources.";
+  public static final String BOOTSTRAP_ONLY = "A bootstrap-only account is the account in which Spinnaker itself is deployed. When true, this account will not be included the accounts managed by Spinnaker. Similarly, the bootstrap instance of Spinnaker will only contain this account. Only one account can be marked as bootstrap-only.";
```
lwander: It feels tedious that a user has to manually flag this account after they've  (in a separate place in the config) already marked the account as being used for "bootstrapping". It might make more sense to have this command be a flag in teh `DeploymentEnvironment`, so they can say something like `hal deploy edit --account-name 'my-boostrapping-account' --bootstrap-only`

lwander: Also it gets confusing if they've configured their k8s account with a docker registry that would have to be flagged too - which halyard should automatically address.

ttomsu:
I moved the `--bootstrap-only` flag to the `hal config deploy` command as suggested. I also took a stab at intelligently removing and disabling docker registries/accounts when `--bootstrap-only` is specified. Namely, if one is deploying Spinnaker in a kubernetes environment (with a supporting docker registry), but is only managing other non-container based environments, the deployed Spinnaker will have both Kuberentes + Docker providers disabled.

Furthermore, this logic tries to identify and remove docker registries that do not have any associated container-based accounts. So if you add docker registry A, don't associate it with any thing, and enable `--bootstrap-only`, it will not appear in your Spinnaker deployment (basically just wasting cycles).

ttomsu:
cc @willgorman

willgorman:
The DC/OS changes look fine to me.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/ContainerAccount.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2017 Netflix, Inc.
```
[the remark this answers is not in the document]
ttomsu: More like "CYA"

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/containers/DockerRegistryReference.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google, Inc.
+ * Copyright 2017 Netflix, Inc.
```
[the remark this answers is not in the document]
ttomsu: Done.

PR #779, opened 2017-11-30
feat(xenial_builds): Apache start separately from Spinnaker services.

brandonnelson3:


halyard-deploy/src/main/resources/debian/init.sh
```diff
@@ -1,19 +1,34 @@
...
     service \$i start
   done
 end script
 EOL
+elif [ "$VERSION_ID" = "16.04" ]; then
+  cat > /lib/systemd/system/spinnaker.service <<EOL
+[Unit]
+Description=All Spinnaker services
+After=network.target
+Wants=gate.service orca.service clouddriver.service front50.service rosco.service igor.service echo.service fiat.service
```
lwander: Should this somehow come from {%services%} as well?

lwander: The reason being, you can use Halyard to install just a subset of components on several instances. Requiring all services each time would break that.

brandonnelson3: Ah IC, uh... well sadly it seems that Systemd wants the ".service" part too. I've added a new template variable for systemd-service-configs.

PR #789, opened 2017-12-20
Halyard separation

brandonnelson3:
Halyard can now be configured to deploy referenced files by value instead of by filesystem reference.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
@@ -18,22 +18,24 @@
 package com.netflix.spinnaker.halyard.cli.command.v1.converter;
 
 import com.beust.jcommander.IStringConverter;
-
 import com.netflix.spinnaker.halyard.cli.command.v1.GlobalOptions;
 import com.netflix.spinnaker.halyard.core.error.v1.HalException;
 import com.netflix.spinnaker.halyard.core.problem.v1.Problem;
-import java.io.File;
-import java.io.IOException;
 import org.aspectj.util.FileUtil;
 
+import java.io.File;
+import java.io.IOException;
+
 public class LocalFileConverter implements IStringConverter<String> {
+
   @Override
   public String convert(String value) {
     if (GlobalOptions.getGlobalOptions().isUseRemoteDaemon()) {
       try {
         return FileUtil.readAsString(new File(value));
       } catch (IOException e) {
-        throw new HalException(Problem.Severity.FATAL, "Config references file that is unreadable: " + value);
+        throw new HalException(Problem.Severity.FATAL,
+            "Config references file that is unreadable: " + value);
       }
     }
     return new File(value).getAbsolutePath();
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
@@ -35,7 +35,7 @@ public class LocalFileConverter implements IStringConverter<String> {
         return FileUtil.readAsString(new File(value));
       } catch (IOException e) {
         throw new HalException(Problem.Severity.FATAL,
-            "Config references file that is unreadable: " + value);
+            "Was passed parameter " + value + " to unreadable file: " + e.getMessage());
       }
     }
     return new File(value).getAbsolutePath();
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
@@ -19,11 +19,23 @@ package com.netflix.spinnaker.halyard.cli.command.v1.converter;
 
 import com.beust.jcommander.IStringConverter;
 
+import com.netflix.spinnaker.halyard.cli.command.v1.GlobalOptions;
+import com.netflix.spinnaker.halyard.core.error.v1.HalException;
+import com.netflix.spinnaker.halyard.core.problem.v1.Problem;
 import java.io.File;
+import java.io.IOException;
+import org.aspectj.util.FileUtil;
 
-public class PathExpandingConverter implements IStringConverter<String> {
+public class LocalFileConverter implements IStringConverter<String> {
   @Override
   public String convert(String value) {
+    if (GlobalOptions.getGlobalOptions().isUseRemoteDaemon()) {
+      try {
+        return FileUtil.readAsString(new File(value));
+      } catch (IOException e) {
+        throw new HalException(Problem.Severity.FATAL, "Config references file that is unreadable: " + value);
+      }
+    }
     return new File(value).getAbsolutePath();
   }
 }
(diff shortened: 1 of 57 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
```diff
@@ -18,12 +18,26 @@
...
 
-public class PathExpandingConverter implements IStringConverter<String> {
   @Override
   public String convert(String value) {
+    if (GlobalApplicationOptions.getInstance().isUseRemoteDaemon()) {
+      try {
+        return FileUtil.readAsString(new File(value));
+      } catch (IOException e) {
+        throw new HalException(Problem.Severity.FATAL,
+            "Config references file that is unreadable: " + value);
```
lwander: I don't think this should read "config", instead "Was passed parameter " + value + " to unreadable file: " + e.getMessage()

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
```diff
@@ -140,6 +155,8 @@
     builder.setValidate(doValidate);
     builder.setRevert(() -> halconfigParser.undoChanges());
     builder.setSave(() -> halconfigParser.saveConfig());
+    Path stagingPath = halconfigDirectoryStructure.getConfigPath(deploymentName);
```
lwander: Should this be named "configPath"? (Here and in a few other spots).

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/GlobalApplicationOptions.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.IOException;
+
+/**
+ * This is the collection of general, top-level flags that come from the application configuration
+ * for halyard.
+ */
+@Data
+public class GlobalApplicationOptions {
+
+  private static final String CONFIG_PATH = "/tmp/hal.config";
```
lwander: I think `tmp` is fair game to be deleted

lwander: I think it's better to keep this close to other halyard config, so something like /opt/halyard/config or ~/.hal

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -141,6 +158,42 @@ public void saveConfig() {
...
+  public void cleanLocalFiles(Path stagingDirectoryPath) {
+    if (GlobalApplicationOptions.getInstance().isUseRemoteDaemon()) {
+      Halconfig halconfig = getHalconfig();
+      Set<String> referencedFiles = new HashSet<String>();
+      Consumer<Node> fileFinder = n -> referencedFiles.addAll(n.localFiles().stream().map(f -> {
+        try {
+          f.setAccessible(true);
+          return (String) f.get(n);
+        } catch (IllegalAccessException e) {
+          throw new HalException(FATAL, "Failed to clean staging directory: " + e.getMessage(), e);
```
lwander: This feels more like a runtime error than a halexception - by convention the latter is reserved for errors the user can do something about.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -141,6 +158,42 @@ public void saveConfig() {
     saveConfigTo(Paths.get(halconfigPath));
   }
 
+  /**
+   * Deletes all files in the staging directory that are not referenced in the hal config.
+   */
+  public void cleanLocalFiles(Path stagingDirectoryPath) {
+    if (GlobalApplicationOptions.getInstance().isUseRemoteDaemon()) {
```
lwander: This needs to be configured another way - those config options are used by the CLI - this should use spring.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
brandonnelson3: If by a few, you mean 30. Yes ;)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
brandonnelson3: SGTM.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/converter/LocalFileConverter.java
brandonnelson3: Done.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
brandonnelson3: Done.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -141,6 +162,42 @@ public void saveConfig() {
     saveConfigTo(Paths.get(halconfigPath));
   }
 
+  /**
+   * Deletes all files in the staging directory that are not referenced in the hal config.
+   */
+  public void cleanLocalFiles(Path stagingDirectoryPath) {
+    if (globalApplicationOptions.isUseRemoteDaemon()) {
```
ttomsu: Suggest returning early:

```
if (!globalApplicationOptions.isRemoteDaemon()) {
  return
}
```

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -169,6 +183,30 @@ public void parentify() {
     return res;
   }
 
+  public void stageLocalFiles(Path outputPath) {
+    if (GlobalApplicationOptions.getInstance().isUseRemoteDaemon()) {
```
ttomsu: Return early

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/GlobalApplicationOptions.java
```diff
@@ -0,0 +1,73 @@
...
+ * This is the collection of general, top-level flags that come from the application configuration for halyard.
+ * This class is both @AutoWired'able and can act as a singleton via the getInstance() function. This is so the same
+ * class may be reused between halyard's CLI and daemon.
+ */
+@Data
+@Configuration
+@PropertySource("file:" + GlobalApplicationOptions.CONFIG_PATH)
+public class GlobalApplicationOptions {
+
+  static final String CONFIG_PATH = "/opt/spinnaker/hal.properties";
```
ttomsu: Is this config shared between the client and the server? Based on the javadoc string, it appears so.

All of the other components follow a convention for server configuration options that hooks into Spring's parsing and options logic (here's a guide on how that works: https://www.travistomsu.com/2016/12/19/configuring-spinnaker/)

If possible, I'd suggest using the yaml format instead of `.properties` files.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
brandonnelson3: Done.

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/GlobalApplicationOptions.java
brandonnelson3: The reason this is done this way is because the CLI doesn't use spring and i'd prefer to share the implementation between the two.

brandonnelson3: Also its still Autowire-ed in the halyard-web backend.

PR #791, opened 2017-12-29
feat(artifacts): Add support for GitHub artifact account

wjoel:
Allows Halyard to configure clouddriver to download artifacts from GitHub, see https://github.com/spinnaker/clouddriver/pull/2231

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/github/GitHubAddArtifactAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/github/GitHubAddArtifactAccountCommand.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2017 Joel Wilsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/github/GitHubEditArtifactAccountCommand.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2017 Google, Inc.
```
lwander: You can drop Google from all of these

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/github/GitHubAddArtifactAccountCommand.java
```diff
@@ -0,0 +1,81 @@
...
+  )
+  private String password;
+  @Parameter(
+      names = "--username-password-file",
+      converter = PathExpandingConverter.class,
+      description = "File containing \"username:password\" to use for GitHub authentication"
+  )
+  private String usernamePasswordFile;
+  @Parameter(
+      names = "--token",
```
wjoel: @lwander should this also have `password = true`?

lwander: ah good catch, yes

wjoel: done

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/github/GitHubEditArtifactAccountCommand.java
wjoel: done

PR #794, opened 2018-01-04
feat(bom): read application default creds if available

lwander:


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/GoogleProfileReader.java
```diff
@@ -54,8 +55,16 @@
...
     JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
     String applicationName = "Spinnaker/Halyard";
-
-    return new Storage.Builder(GoogleCredentials.buildHttpTransport(), jsonFactory, GoogleCredentials.retryRequestInitializer())
+    HttpRequestInitializer requestInitializer;
+    try {
+      requestInitializer = GoogleCredentials.setHttpTimeout(GoogleCredential.getApplicationDefault());
+      log.info("Loaded application default credential for reading BOMs & profiles.");
+    } catch (Exception e) {
+      requestInitializer = GoogleCredentials.retryRequestInitializer();
```
ewiseblatt: Because credentials are not normally required in a standard use case scenario, the message details are probably distracting -- at least if they are because there no credentials. If the error is a problem with actual supplied credentials then that would be interesting.

lwander: That's fair - I added this since I thought you might find it useful when getting started, especially because there will be no other indication (other than a 401 when trying to read) that you're creds may not have been loaded. Is that OK?

ewiseblatt: your call. I'm not sure of the context of this call. I assumed that the generated exception would be different if you did not have credentials vs if they were invalid. For example if you arent on GCP and dont have the environment variable set then I dont think the exception message would be useful and unlikely relevant.

lwander: relegated to `log.debug`

PR #806, opened 2018-01-17
feat(halyard_installation): Prevent accidental second install.

brandonnelson3:
In the jar based installation, first check to make sure the debian package is not installed. If it is, provide a link to the user to migrate.

### DIFF
--- a/install/debian/InstallHalyard.sh
+++ b/install/debian/InstallHalyard.sh
@@ -6,7 +6,7 @@ function check_migration_needed() {
   set -e
   hash dpkg &> /dev/null
 
-  if [ "$?" != "0" ]; then
+  if [ "$?" = "0" ]; then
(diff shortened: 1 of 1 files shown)


install/debian/InstallHalyard.sh
```diff
@@ -2,6 +2,18 @@
 
 set -e
 
+function check_migration_needed() {
+  set -e
+  dpkg -s spinnaker-halyard &> /dev/null
+
+  if [ "$?" != "1" ]; then
```
lwander: This can fail if `dpkg` isn't installed

brandonnelson3: Added a check around this.

peterkomar:
Hello @brandonnelson3 . If it first installation on Ubuntu 14.04.5 installation script exit without any messages. After this change.

I use vagrant ubuntu/trusty64 image

PR #818, opened 2018-01-23
fix(localgit): Remove and then add git remote

davidxia:
If we run `git remote add upstream` when `upstream` already
exists, we get the annoying output `fatal: remote upstream already exists.`
for every component.

Run `git remote remove upstream` first to avoid printing this error.
This also mimics the behavior of the newer git command `git remote
set-url` that we use when git is above a certain version.

PR #821, opened 2018-01-24
fix(localgit): Do not fail if removing git remote fails

davidxia:
The remote "upstream" might not exist. Do not fail if so.

fixes bug introduced in https://github.com/spinnaker/halyard/pull/818


davidxia:
cc @lwander

lwander:
Sorry, should have caught this in review

PR #828, opened 2018-01-26
fix(core): Fix NPE when no clean step provided

lwander:
I'm not sure if this is correct "long term" fix, but it at least allows us to run requests that don't have a "clean" step specified.


lwander:
@brandonnelson3 PTAL when you get back

PR #834, opened 2018-01-28
Kubernetes URL fixed

croeck:
Was:
https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/to

Should be:
https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/


lwander:
That directory is autogenerated, can you update the entry in the halyard-cli source? https://github.com/spinnaker/halyard/blob/master/docs/adding-a-config-parameter.md#4a-cli-docs

croeck:
Done, changed the source fill as well. I did not know that the file is auto generated.

lwander:
Thanks! the spinnaker.io docs will reflect this next time we publish halyard

PR #842, opened 2018-02-12
feat(jenkins_master): Add CSRF config for Jenkins Masters.

jtk54:
Relies on https://github.com/spinnaker/igor/pull/222.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/jenkins/JenkinsAddMasterCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/jenkins/JenkinsAddMasterCommand.java
@@ -49,12 +49,19 @@ public class JenkinsAddMasterCommand extends AbstractAddMasterCommand {
   )
   public String password;
 
+  @Parameter(
+      names = "--csrf",
+      description = JenkinsCommandProperties.CSRF_DESCRIPTION
+  )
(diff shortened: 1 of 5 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/jenkins/JenkinsAddMasterCommand.java
```diff
@@ -49,12 +49,19 @@ protected String getCiName() {
...
+  public Boolean csrf;
+
   @Override
   protected Master buildMaster(String masterName) {
     JenkinsMaster master = (JenkinsMaster) new JenkinsMaster().setName(masterName);
     master.setAddress(address)
         .setPassword(password)
-        .setUsername(username);
+        .setUsername(username)
+        .setCsrf(csrf);
```
lwander: OK if this is null?

jtk54: Yes, it defaults to false in Igor. Should I add a default in hal also?

lwander: No that's fine then, just checking

jtk54:
Docs generated.

Issue #838, opened 2018-02-09
Document Services for localgit setup


brandonnelson3:
We should probably have a doc that acknowledges the fact that Rosco and Monitoring are not currently setup to be run when using localgit type.

spinnaker-release:
Migrated to https://github.com/spinnaker/spinnaker/issues/2434

PR #849, opened 2018-02-16
feat(rosco): Enable rosco for halyard local git installation

ezimanyi:


ezimanyi:
@lwander

lwander:
https://github.com/spinnaker/halyard/issues/838

PR #852, opened 2018-02-22
fix(localgit): Properly build destination paths for custom profile fi…

duftler:
…les.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
@@ -37,7 +37,6 @@ import com.netflix.spinnaker.halyard.core.resource.v1.StringReplaceJarResource;
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
-import org.apache.commons.lang.NotImplementedException;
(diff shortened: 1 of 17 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/NestableCommand.java
```diff
@@ -37,7 +37,6 @@
 import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
-import org.apache.commons.lang.NotImplementedException;
```
duftler: Swapped this for `UnsupportedOperationException` since jackson seems to enter an infinitely recursive loop when trying to serialize `NotImplementedException`.

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/GitProfileReader.java
```diff
@@ -21,9 +21,7 @@
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
-import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
duftler: Removed all these internal references.

PR #855, opened 2018-02-23
fix(kubernetes): fix gcloud binary path

lwander:
When deploying to a k8s cluster using gcp (gcloud) as the auth provider,
it was possible that the path set was different from the path provided
inside the clouddriver container. Setting the path here to be `gcloud`
in only the staged kubeconfig fixes this.


lwander:
https://github.com/spinnaker/spinnaker/issues/2456

PR #861, opened 2018-03-05
feat(kubernetes): Allow custom pod annotations for spinnaker services

polarbizzle:
When deploying spinnaker to kubernetes, it is sometimes necessary for
some of the services to have extra annotations. For example when using
kube2iam to specify the role certain pods should get. This change adds a
map for podAnnotations to the KubernetesSettings and applies it to both
the replicasets in the v1 provider and the deployments for the v2
provider.


jonathancole-h1:
How does one use this feature to add the specific kube2iam annotation to the hal config?

jonathancole-h1:
Looks like putting it in `~/.hal/default/service-settings/front50.yml` seems to work:
```
kubernetes:
  podAnnotations:
    iam.amazonaws.com/role: your_iam_role
```

PR #863, opened 2018-03-16
chore(install): add missing install params

lwander:


install/debian/InstallHalyard.sh
```diff
@@ -152,6 +192,24 @@ usage: $0 [-y] [--version=<version>] [--user=<user>]
     -y                              Accept all default options during install
                                     (non-interactive mode).
 
+    --halyard-bucket <name>         The bucket the Halyard JAR to be installed
```
ewiseblatt: how do the jars get into this bucket?

lwander: Talked offline: https://github.com/spinnaker/halyard/blob/master/release/publish.sh

install/debian/InstallHalyard.sh
```diff
@@ -93,6 +118,18 @@ function configure_defaults() {
...
   echo "$(tput bold)Halyard version will be $HALYARD_VERSION $(tput sgr0)"
 
+  if [ -z "$HALYARD_BUCKET_BASE_URL" ]; then
+    HALYARD_BUCKET_BASE_URL="gs://spinnaker-artifacts/halyard"
+  fi
+
+  echo "$(tput bold)Halyard will be downloaded from $HALYARD_BUCKET_BASE_URL $(tput sgr0)"
+
+  if [ -z "$CONFIG_BUCKET" ]; then
+    CONFIG_BUCKET="halconfig"
```
ewiseblatt: As noted offline, should another storage system be supported in the future (e.g. s3) then this will likely need to be complemented by something to turn it into a url. Changing this to a URL here would likely be a big change because of all the configuration references and not worth doing. The least impact change would be to add a second parameter (when needed) to override the gs:// assumption. Other courses of action can be considered when the time comes that it is needed.

Overall, I think introducing the bucket base as a URL above is still the right way to go even though it is inconsistent with how the config bucket is specified.

PR #873, opened 2018-03-18
feat(kayenta): Add initial bits for kayenta support.

duftler:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/git/LocalGitKayentaService.java
```diff
@@ -0,0 +1,67 @@
...
+import lombok.EqualsAndHashCode;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.stereotype.Component;
+
+@EqualsAndHashCode(callSuper = true)
+@Data
+@Component
+// TODO(duftler): Remove this once the kayenta repo is public.
+@ConditionalOnExpression("${canary.enabled:false}")
```
lwander: Why is https://github.com/spinnaker/halyard/pull/873/files#diff-93e9e64ef687d26dcb22df09afc9aca3R56 not enough?

duftler: Not sure how I missed this comment earlier. If I don't make the component conditional it will always be autowired into `LocalGitServiceProvider`, even if `Canary` is disabled. Since the repo is still private, the operator may not have access and the `hal deploy apply` command will fail when the daemon attempts to clone it.

lwander: Ah got it

PR #876, opened 2018-03-20
fix(localgit): Properly capture pid for deck.

duftler:


halyard-deploy/src/main/resources/git/deck-start.sh
```diff
@@ -0,0 +1,29 @@
...
+  fi
+
+ ${STOP_SCRIPT}
+
+  pushd $ARTIFACT_DIR
+  {%start-command%} \
+    2> ${LOG_DIR}/${ARTIFACT}.err \
+    > ${LOG_DIR}/${ARTIFACT}.log &
+
+  while ! nc -z localhost 9000; do sleep 1; done;
```
duftler: Not thrilled about the `9000` here, but I think it's ok for localgit.

PR #877, opened 2018-03-21
feat(localgit): Add --prep-only flag to just clone repos.

duftler:
Didn't bother to add the `--stage-only` flag to `ApplyDeployCommand` yet.

WDYT of the approach? Options include:

1) Go with what's here.
2) Contain the `stageOnly` checks to just `ApplyDeployCommand`. This will result in the repo's still being cloned, but no files will be generated in `~/.hal/default/staging`.
3) Decide it's too messy and not worth it and abandon it.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/deploy/ApplyDeployCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/deploy/ApplyDeployCommand.java
@@ -64,6 +64,13 @@ public class ApplyDeployCommand extends AbstractRemoteActionCommand {
   boolean flushInfrastructureCaches;
 
   @Parameter(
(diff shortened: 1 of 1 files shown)


lwander:
This might get a little confusing, since only the local git, debian & bake deployments really make use of that remote action. Everything else completes their deployment before that script is sent back to the user.

duftler:
Yup, I agree; so (1) is out. That leaves (2) or (3).

lwander:
Do you think 2 is worth it? Maybe there is some happy alternate path that creates the config as a separate command.

brandonnelson3:
I agree. I think 2 is easy enough to just knock out. If it starts to get out of hand then maybe reconsider but I think it will likely be reasonable.

duftler:
Updated to just do the cloning for now. PTAL.

duftler:
ping

PR #878, opened 2018-03-21
feat(kayenta): Support initial set of top-level parameters.

duftler:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/EditCanaryCommand.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * Copyright 2018 Netflix, Inc.
```
lwander: Hmmm

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/EditCanaryCommand.java
```diff
@@ -0,0 +1,132 @@
...
+      description = "Location of web components to use for Atlas metric configuration."
+  )
+  private String atlasWebComponentsUrl = null;
+
+  @Parameter(
+      names = "--templates-enabled",
+      arity = 1,
+      description = "Whether or not to enable custom filter templates for canary configs in deck."
+  )
+  private Boolean templatesEnabled = true;
```
lwander: Since this is an "edit" command, if this flag isn't supplied by a user it will always be set to `true`. I would drop these defaults.

duftler: I want it to default to `true`. Is there a better way to accomplish that?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/EditCanaryCommand.java
duftler: Fixed.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/EditCanaryCommand.java
duftler: Fixed as per offline discussion.

PR #881, opened 2018-03-23
fix(changelog): Stop fetching placeholder changelog

ezimanyi:
When running an unvalidated version of spinnaker, set the changelog
to an empty string so we skip changelog fetching rather than try to
fetch the placeholder {%changelog.gist.id%} from GitHub.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
@@ -82,11 +82,14 @@ public class DeckProfileFactory extends RegistryBackedProfileFactory {
 
     Optional<Versions.Version> validatedVersion = versionsService.getVersions().getVersion(version);
 
-    validatedVersion.ifPresent(v -> {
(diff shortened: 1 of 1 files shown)


ezimanyi:
This fixes the numerous console errors about being unable to fetch a gist when running an unvalidated version of Spinnaker.  These were harmless but cluttered up the JavaScript console.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
```diff
@@ -82,11 +82,14 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
 
     Optional<Versions.Version> validatedVersion = versionsService.getVersions().getVersion(version);
 
-    validatedVersion.ifPresent(v -> {
-      String changelog = v.getChangelog();
+    if (validatedVersion.isPresent()) {
```
lwander: Why switch from the `.ifPresent` to `if (.isPresent())`?

lwander: Oops I see - I'm an idiot

PR #889, opened 2018-04-02
feat(kayenta/prometheus): First pass at support for configuring Prome…

duftler:
…theus canary service integration.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/prometheus/EditCanaryPrometheusCommand.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "edit";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Edit Spinnaker's canary analysis Prometheus service integration settings.";
+
+  @Parameter(
+      names = "--metadata-caching-interval-ms",
+      arity = 1,
```
lwander: Arity is 1 by default for non bool/list params

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/KayentaProfileFactory.java
```diff
@@ -108,6 +114,19 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
           enabled = googleSvc.isStackdriverEnabled();
         }
       }
+
+      @Data
```
lwander: If this is needed for serialization/deserialization you might need to add a `@NoArgsConstructor` here since you overrode the default

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/canary/prometheus/EditCanaryPrometheusCommand.java
duftler: Fixed here and in `EditCanaryGoogleCommand`.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/KayentaProfileFactory.java
duftler: Do the profiles get deserialized somewhere (only supporting `localgit` so far, so maybe there is a path)?

duftler:
Also re-generated command docs.

PR #896, opened 2018-04-06
feat(provider/appengine): add feature flag to enable container image url deployments

sbwsg:


danielpeach:
That's here: https://github.com/spinnaker/deck/pull/5126, but that means that the `@ValidForSpinnaker` should be for a later version, since the halconfig for 1.6.0 won't have that change.

sbwsg:
Whoops, yes, thankyou: https://github.com/spinnaker/deck/pull/5126/files

sbwsg:
Pushed it back to a minimum version of 1.7.0

PR #898, opened 2018-04-09
fix(deploy/aws): mount aws creds in spinnaker home

lwander:
This is to allow spinnaker to be deployed using the non-root docker images into k8s when managing aws resources.


lwander:
@duftler this will likely need to be done with kayenta, but we only need to support /home/spinnaker in that case as it's being released after the non-root docker images. https://github.com/spinnaker/spinnaker/issues/2606

PR #910, opened 2018-04-12
feat(aws/bakery): Add initial aws bakery configuration commands and s…

duftler:
…upport for ~/.hal/$DEPLOYMENT/profiles/images-local.yml.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/aws/AwsBaseImage.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/aws/AwsBaseImage.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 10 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/aws/AwsBaseImage.java
```diff
@@ -0,0 +1,62 @@
...
+  private List<AwsVirtualizationSettings> virtualizationSettings;
+
+  @EqualsAndHashCode(callSuper = true)
+  @Data
+  @ToString(callSuper = true)
+  public static class AwsImageSettings extends BaseImage.ImageSettings {
+  }
+
+  @Data
+  @ToString
```
lwander: `@ToString` comes for free with `@Data` https://projectlombok.org/features/Data

duftler: Fixed.

PR #916, opened 2018-04-17
feat(deploy/ecs): add ecs support to halyard

melbit-michaelw:
PR to add ECS support into Halyard. This also requires [Deck PR 5190](https://github.com/spinnaker/deck/pull/5190)

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/ecs/EcsAccountValidator.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/ecs/EcsAccountValidator.java
@@ -1,0 +1,44 @@
+package com.netflix.spinnaker.halyard.config.validate.v1.providers.ecs;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Account;
+import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
+import com.netflix.spinnaker.halyard.config.model.v1.providers.ecs.EcsAccount;
+import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
+import com.netflix.spinnaker.halyard.config.services.v1.AccountService;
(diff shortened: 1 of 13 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/ecs/EcsAccountValidator.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Autowired
+    AccountService accountService;
+
+    @Autowired
+    ConfigService configService;
+
+    @Override
+    public void validate(ConfigProblemSetBuilder p, EcsAccount n) {
+        p.addProblem(Severity.WARNING, "Very limited validation for the ECS provider has been implemented.");
```
lwander: Could you rephrase this to "This only validates that a corresponding AWS account has been created for your ECS account."

I think that'll potentially avoid some questions down the road when people see a WARNING in their deploy logs.

lwander:
Nice done! To generate the docs that we will promote to spinnaker.io on the next stable release of halyard, can you run `make` in the `./halyard-cli` directory, and check in the changes?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/ecs/EcsAccountValidator.java
melbit-michaelw: @lwander Sorry, I'm new at this. I've squashed my commits and force pushed which seems to be the correct approach based on the contributing page, but that makes the comments in this PR look odd since they no longer reflect the code. Should I instead be just adding new commits, and they can be squashed when merged into master ?

Anyway, added the docs and updated the warning message.

lwander: No worries, that's the right approach. I (personally) prefer keeping all commits separate until we merge for clarity, maybe we can update the contributor guide to reflect that.

PR #918, opened 2018-04-18
fix(deploy): Gracefully handle missing artifacts

ezimanyi:
When deploying a version of Spinnaker before Kayenta, the distributed GCE deployment currently throws an NPE.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
@@ -134,9 +134,13 @@ public interface GoogleDistributedService<T> extends DistributedService<T, Googl
   default String getArtifactId(String deploymentName) {
     String artifactName = getArtifact().getName();
     String version = getArtifactService().getArtifactVersion(deploymentName, getArtifact());
-    return String.format("projects/%s/global/images/%s",
-        getGoogleImageProject(deploymentName),
-        String.join("-", "spinnaker", artifactName, version.replace(".", "-").replace(":", "-")));
+    if (version == null) {
+      return null;
+    } else {
+      return String.format("projects/%s/global/images/%s",
+          getGoogleImageProject(deploymentName),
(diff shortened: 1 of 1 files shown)


ezimanyi:
With this change (and after re-generating some missing GCE images), I was able to successfully deploy both 1.6.0 and 1.6.1 using a distributed GCE setup.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/google/GoogleDistributedService.java
```diff
@@ -134,9 +134,13 @@ default String getStartupScript() {
   default String getArtifactId(String deploymentName) {
     String artifactName = getArtifact().getName();
     String version = getArtifactService().getArtifactVersion(deploymentName, getArtifact());
-    return String.format("projects/%s/global/images/%s",
-        getGoogleImageProject(deploymentName),
-        String.join("-", "spinnaker", artifactName, version.replace(".", "-").replace(":", "-")));
+    if (version == null) {
```
lwander: I think this should be handled upstream from here. I realize this is where the exception will be thrown,  but I think the deployer in charge of picking which services to create should be smarter about asking for "artifact ids" of services that can't be created.

ezimanyi: I'm sort of torn on whether this should be handled here or upstream.  I actually think that it's this function (or perhaps the downstream one `getArtifactVersion`) that's in the best position to know whether or not a given service should be built.  Effectively, the question of whether a service should be built is answered by looking at whether it's in the BOM---which is what this function is doing to get the version to deploy.  So I think it make sense that asking "what version of X should I deploy?" would return `null` if the answer is not to deploy X.

That being said, I agree that it's a bit silly that we still end up with an entry in `endpoints.services` for a service that's not being deployed...so maybe we should write a `boolean serviceExists()` that just checks if there's a version at all and short-circuit all the settings-building if it's false.  Let me see what this last idea would look like.

ezimanyi:
Tested by doing the following deployments: LocalDebian, BakeDebian, Kubernetes (V1), Kubernetes (V2), Distributed GCE.

ezimanyi:
@lwander : Mind taking a look at this updated change?

PR #924, opened 2018-04-25
fix(Dockerfile): Create home dir for spinnaker user in Dockerfile

posox:
This patch fixes issue spinnaker/spinnaker#2703


lwander:
Thanks!

lwander:
Being promoted to :stable now...

PR #926, opened 2018-04-27
fix(localgit): Fix version validation for localgit deployment

ezimanyi:


ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/2705.

Now `hal config version edit --version 1.5.0` warns `Version "1.5.0" was patched by "1.5.4". Please upgrade` but `hal config version edit --version 1.5.4` does not warn.

PR #927, opened 2018-04-27
fix(rosco): Preserve file permissions for local git installation

ezimanyi:
Fixes an issue whereby executable files in the packer config directory
would get their execute bit unset when deploying halyard in the local
git mode.


ezimanyi:
It's sort of annoying that I had to write functions to convert between `java.nio` file permissions and `apache.commons` file bits, but I couldn't find a reasonable library to do this.

PR #931, opened 2018-05-02
Add support for setting a certificate chain

jasonmcintosh:
Should help with SSL validation of the UI.


jasonmcintosh:
SO apparently you don't need this and it's deprecated in 2.4.8 and newer.  From then on they support adding the chain to the file.  The quirk was it seems apache wants the intermediate 2nd in the file while most systems seem to want it first.

PR #933, opened 2018-05-04
Change the kubernetes proxyServiceEndpoint

kangwoo:
Fixed an issue https://github.com/spinnaker/spinnaker/issues/2728

hal deploy apply uses the old style proxy path for kubernetes which was removed in v1.10


kangwoo:
It works fine with version 1.2 or later.

ezimanyi:
One small point---the new endpoint requires the `services/proxy` RBAC permission whereas the old one did not.  I'm going to push a small change to improve the error message when that permission is lacking to make it easier for anyone who runs into that issue to diagnose and fix it.

kangwoo:
Thanks

PR #938, opened 2018-05-08
fix(ssl): Enable deck https endpoint for local git deployment.

stewchen:


ttomsu:
Whoops - I think I meant to only LGTM this if other folks wanted to look at it.

PR #941, opened 2018-05-11
fix(deploy/local): add version to profile override

lwander:
without this, it's impossible to have more than 1 local version stored, as the subcomponent versions will always have to be identical.

https://github.com/spinnaker/spinnaker/issues/2640


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/LocalDiskProfileReader.java
```diff
@@ -58,64 +58,66 @@
 
   @Override
   public InputStream readProfile(String artifactName, String version, String profileName) throws IOException {
-    String path = profilePath(artifactName, profileName);
+    version = version.substring("local:".length());
+    String path = profilePath(artifactName, version, profileName);
```
danielpeach: This means my profiles will have to live under e.g. `.hal/.boms/clouddriver/my-clouddriver-version`, right? Can this be optional, possibly by trying to look under that path, and otherwise picking the old path?

This breaks my current system and would require me to check in a new set of profiles for every version bump.

lwander: Yeah we can fall back to the non-versioned one if needed

lwander:
@danielpeach updated

danielpeach:
Thanks

PR #942, opened 2018-05-11
chore(kayenta/aws): Add support for access/secret keys.

duftler:


duftler:
Closes https://github.com/spinnaker/kayenta/issues/304.

PR #943, opened 2018-05-14
fix(ci/jenkins): Adds argument to CSRF flag.

jtk54:
FYI @dorbin (sorey).


jtk54:
Closes https://github.com/spinnaker/spinnaker/issues/2602.

PR #944, opened 2018-05-15
Trivial Doc update: use api key instead of password

nparks-kenzan:
Halyard command documentation says _password_ where one should get the APIKEY for the user per IGOR's own documentation. 

Keeps the password flag (to not break anyone's scripting), just updates the documentation to say **api key** in the description for the flag in commands.md


nparks-kenzan:
Oh yeah (egg on face). I will close this and open again when done.

PR #948, opened 2018-05-21
feat(customization): Enables per-service override of binary source re…

ttomsu:
…pository.

Example custom BOM that replaces Rosco with my own build would look like:

```
version: 1.7.5
timestamp: '2018-05-21 14:16:06'
services:
  ...
  rosco:
    version: local:foo
    artifactSources:
      dockerRegistry: gcr.io/ttomsu-dev-spinnaker
  ...
dependencies:
  ...
artifactSources:
  debianRepository: https://dl.bintray.com/spinnaker-releases/debians
  dockerRegistry: gcr.io/spinnaker-marketplace
  googleImageProject: marketplace-spinnaker-release
  gitPrefix: https://github.com/spinnaker
```

This results in the container having image `gcr.io/ttomsu-dev-spinnaker/rosco:foo`, and Rosco configs being pulled from my local `~/.hal/.boms/bom/rosco/foo` directory. 

I only tested this against a Kubernetes V2 deployment. Additional work (and testing) will need to be done to make this work with Debian packages.

cc/ @ewiseblatt @danielpeach

### DIFF
--- a/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
+++ b/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
@@ -48,6 +48,10 @@ public class BillOfMaterials {
     String getArtifactCommit(String artifactName) {
       return getFieldArtifact(Dependencies.class, this, artifactName).getCommit();
     }
+
+    ArtifactSources getArtifactSources(String artifactName) {
+      return getFieldArtifact(Dependencies.class, this, artifactName).getArtifactSources();
+    }
   }
 
   @Data
@@ -85,12 +89,17 @@ public class BillOfMaterials {
     String getArtifactCommit(String artifactName) {
       return getFieldArtifact(Services.class, this, artifactName).getCommit();
     }
+
+    ArtifactSources getArtifactSources(String artifactName) {
+      return getFieldArtifact(Services.class, this, artifactName).getArtifactSources();
+    }
   }
 
   @Data
   static class Artifact {
     String version;
     String commit;
+    ArtifactSources artifactSources;
   }
 
   static private <T> Artifact getFieldArtifact(Class<T> clazz, T obj, String artifactName) {
@@ -114,7 +123,7 @@ public class BillOfMaterials {
       if (result == null && !artifactName.equals("defaultArtifact")) {
         result = getFieldArtifact(clazz, obj, "defaultArtifact");
       }
-      
+
       if (result == null) {
         result = new Artifact();
       }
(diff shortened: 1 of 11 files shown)


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
```diff
@@ -85,12 +89,17 @@ String getArtifactVersion(String artifactName) {
     String getArtifactCommit(String artifactName) {
       return getFieldArtifact(Services.class, this, artifactName).getCommit();
     }
+
+    ArtifactSources getArtifactSources(String artifactName) {
+      return getFieldArtifact(Services.class, this, artifactName).getArtifactSources();
```
lwander: Should this default to the top-level artifact sources if `null`?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/services/v1/ArtifactService.java
```diff
@@ -61,17 +62,63 @@
...
+  ArtifactSourcesConfig artifactSourcesConfig;
+
   BillOfMaterials getBillOfMaterials(String deploymentName) {
     DeploymentConfiguration deploymentConfiguration = deploymentService.getDeploymentConfiguration(deploymentName);
     String version = deploymentConfiguration.getVersion();
     return versionsService.getBillOfMaterials(version);
   }
 
+  /**
+   * Should use {@link #getArtifactSources(String, SpinnakerArtifact)} when it supports all types of
```
lwander: Which types of deployments are missing?

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/BillOfMaterials.java
lwander: nvm I see your merging from the other direction below

lwander:
~/.hal/.boms/bom/rosco/foo is that right? it should be ~/.hal/.boms/rosco/foo

that .bom directory added by a community member mirrors the gs://halconfig structure.

ttomsu:
Nope, you're right - it's just `~/.hal/.boms/rosco/foo`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/services/v1/ArtifactService.java
lwander: Please test against the v1 k8s deployment, and localdebian, or provide detailed instructions for someone down the road to implement this.

ttomsu: `BakeDebianServiceProvider` and `LocalDebianServiceProvider`

ttomsu:
Tested V1 deployment - small modification made, but it works now too.

PR #951, opened 2018-05-25
fix(Halyard): s3 --bucket not accept s3:// prefix

robertflaig:
Solution to "Halyard: s3 --bucket will not accetp s3:// prefix #2408"

fixes #2408

Better solution might be is to find out why the parameter cannot be stored with the s3:// because it should be needed to access the s3 bucket. Possibly the validate function cannot handle the s3:// prefix and the code needed when the bucket is accessed would need updating so this might be the easiest solution.


robertflaig:
What should be in it?

Beginner here.

lwander:
Oh was this change made via the web-UI? if so, I think I can just edit it from here.

robertflaig:
I think I was able to fix the commit message

lwander:
Nice! thanks :)

PR #953, opened 2018-05-29
fix(deck/auth): Add the correct security env variables in deck.

stewchen:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/git/LocalGitDeckService.java
```diff
@@ -57,17 +58,26 @@
...
+      setEnv("FIAT_ENABLED");
+    }
+    return new Settings(security.getUiSecurity())
         .setArtifactId(getArtifactId(deploymentConfiguration.getName()))
-        .setHost(authEnabled ? "0.0.0.0" : getDefaultHost())
+        .setHost(security.getAuthn().isEnabled() ? "0.0.0.0" : getDefaultHost())
         .setEnabled(true);
   }
 
+  private void setEnv(String var) {
```
ttomsu: Recommend naming to `setEnvTrue`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/git/LocalGitDeckService.java
```diff
@@ -57,17 +58,26 @@
...
+    setStartCommand(String.join("\n", "export " + var + "=true", getStartCommand()));
```
ttomsu: Genuine question - do we need to `export` the var? [This SO post](https://stackoverflow.com/a/10856348/5569046) appears to use some bash magic that I've seen @ewiseblatt use with the python scripts.

PR #954, opened 2018-05-29
feat(permissions): Add permissions support for accounts in halyard.

stewchen:
This change also removes the deprecated --requiredGroupMembership flags,
as they are no longer functional.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AbstractAddAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AbstractAddAccountCommand.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.halyard.cli.command.v1.config.providers.account;
 
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.fiat.model.Authorization;
 import com.netflix.spinnaker.halyard.cli.command.v1.NestableCommand;
 import com.netflix.spinnaker.halyard.cli.services.v1.Daemon;
 import com.netflix.spinnaker.halyard.cli.services.v1.OperationHandler;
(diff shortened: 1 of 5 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AbstractAddAccountCommand.java
```diff
@@ -42,10 +43,17 @@
 
   @Parameter(
       variableArity = true,
-      names = "--required-group-membership",
-      description = AccountCommandProperties.REQUIRED_GROUP_MEMBERSHIP_DESCRIPTION
+      names = "--read-permissions",
```
lwander: These flags may no longer be functional in the latest spinnaker release, but still work in older ones. I would not remove them here.

stewchen: Done

stewchen:
Reverted the deletions

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AbstractAddAccountCommand.java
dorbin: @lwander 
Is there a deprecation process by which these flags should be removed at some time in the future?

lwander: There is an annotation for flags that don't work _until_ a certain release -- we need to add/extend that annotation for flags that no longer work _after_ a certain release. That can be done in a separate PR.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/account/AccountCommandProperties.java
```diff
@@ -21,4 +21,8 @@
   public static final String REQUIRED_GROUP_MEMBERSHIP_DESCRIPTION = "A user must be a member of at least one specified group in order to make changes to this account's cloud resources.";
 
   public static final String PROVIDER_VERSION_DESCRIPTION = "Some providers support multiple versions/release tracks. This allows you to pick the version of the provider (not the resources it manages) to run within Spinnaker.";
+
+  public static final String READ_PERMISSION_DESCRIPTION = "A user must have matching roles of all of these read permissions in order to view this account's cloud resources.";
```
ttomsu: A user must have _at least one`_ of these roles in order to be granted the `READ` permission. Same with the write description

stewchen:
Docs update pending the release: https://github.com/spinnaker/spinnaker.github.io/pull/813

PR #955, opened 2018-05-30
fix(version/bom): BOM version command displays message if Spinnaker version is not properly formatted.

joonlim:
* When calling `hal version bom`, checks to make sure that the Spinnaker version in the halconfig file is a properly formatted version (string of numbers with 0 or more dots in between, with no dots directly next to each other).

Before:
```
➜  hal version bom        
+ Get current deployment
  Success
+ Get Spinnaker version
  Success
+ Get current deployment
  Success
+ Get Spinnaker version
  Success
! ERROR 400 
```

After:
```
➜  ./hal version bom                                                                                             
+ Get current deployment
  Success
+ Get Spinnaker version
  Success
! ERROR Failed to get BOM versions because Spinnaker version is not
  properly formatted in halconfig: version='branch:upstream/master'
```

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
@@ -64,25 +64,17 @@ public class BomVersionCommand extends AbstractConfigCommand {
   }
 
   public String getVersion() {
-    String version;
     switch (versions.size()) {
       case 0:
-        version = new OperationHandler<String>()
+        return new OperationHandler<String>()
             .setOperation(Daemon.getVersion(getCurrentDeployment(), false))
             .setFailureMesssage("Failed to get version of Spinnaker configured in your halconfig.")
             .get();
-        break;
       case 1:
-        version = versions.get(0);
-        break;
+        return versions.get(0);
       default:
         throw new IllegalArgumentException("More than one version supplied");
     }
-
-    if (!BomVersionCommand.isProjectVersionFormat(version)) {
-      throw new RuntimeException(String.format("Failed to get BOM versions because Spinnaker version is not properly formatted in halconfig: version='%s'", version));
-    }
-    return version;
   }
 
   @Override
@@ -104,10 +96,5 @@ public class BomVersionCommand extends AbstractConfigCommand {
     }
 
     AnsiPrinter.out.println(result);
-  }
-
-  private static boolean isProjectVersionFormat(String version) {
-    // Matches #, #.#, #.#.#, ...
-    return version.matches("\\d+(\\.\\d+)*");
   }
 }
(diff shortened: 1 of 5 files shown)


joonlim:
@lwander Could you please review this?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
```diff
@@ -97,4 +105,9 @@ protected void executeThis() {
 
     AnsiPrinter.out.println(result);
   }
+
+  private static boolean isProjectVersionFormat(String version) {
```
duftler: Maybe add a simple unit test for this method?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
```diff
@@ -64,17 +64,25 @@ public String getMainParameter() {
...
+        break;
       case 1:
-        return versions.get(0);
+        version = versions.get(0);
+        break;
       default:
         throw new IllegalArgumentException("More than one version supplied");
     }
+
+    if (!BomVersionCommand.isProjectVersionFormat(version)) {
```
lwander: Versions like `master-latest-unvalidated` are perfectly find but would be rejected here. I don't think we want to do the check in this place, but instead when trying to lookup the BOM.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
```diff
@@ -97,4 +105,9 @@ protected void executeThis() {
 
     AnsiPrinter.out.println(result);
   }
+
+  private static boolean isProjectVersionFormat(String version) {
+    // Matches #, #.#, #.#.#, ...
+    return version.matches("\\d+(\\.\\d+)*");
```
lwander: I would delegate to https://github.com/spinnaker/halyard/blob/master/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/Versions.java#L54 to check if a version is a branch or local, and assume all other versions can at least be looked up.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/versions/BomVersionCommand.java
lwander: This is where the BOM lookup happens: https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/VersionsService.java#L63

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
```diff
@@ -727,6 +727,9 @@
   @GET("/v1/versions/bom/{version}")
   DaemonTask<Halconfig, Object> getBillOfMaterials(@Path("version") String version);
 
+  @GET("/v1/versions/bom")
+  DaemonTask<Halconfig, Object> getBillOfMaterialsV2(@Query("version") String version);
```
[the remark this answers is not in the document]
joonlim: Thanks! Done.

PR #956, opened 2018-05-31
fix(halyard): Allow none standard uri for github oauth2 authn

awalker125:
This pull request implements an EGITHUB provider for authn in halyard e.g

```
hal config security authn oauth2 edit \
--client-id YYYYYYYYYYYYYYYYY\
--client-secret XXXXXX \
--provider egithub \
--access-token-uri https://github.example.com/login/oauth/access_token \
--user-authorization-uri https://github.example.com/login/oauth/authorize \
--user-info-uri https://github.example.com/api/v3/user 

```

It allows overriding the uris to be none github. I guess this could be used as any custom provider.

I created this because of https://github.com/spinnaker/spinnaker/issues/2856

I've put this together quickly to get round the problem I have but happy to make amendments if needed/required.


lwander:
@stewchen @ttomsu could you have a look at this?

awalker125:
Have squashed the commits and named as per the convention.

ttomsu:
This looks like a more specific implementation of https://github.com/spinnaker/spinnaker/issues/2664. I'd prefer to see a generic solution for bring-your-own-provider.

@awalker125 Would you be against us closing this in favor of a more generic approach?

PR #959, opened 2018-05-31
refactor(provider/oracle, storage/oracle): oraclebmcs provider/storage is renamed to oracle

guoyongzhang:
oraclebmcs provider is renamed to oracle;  oraclebmcs storage is renamed to oracle.


docs/commands.md
```diff
@@ -250,15 +250,15 @@
  * [**hal config provider openstack bakery edit**](#hal-config-provider-openstack-bakery-edit)
  * [**hal config provider openstack disable**](#hal-config-provider-openstack-disable)
  * [**hal config provider openstack enable**](#hal-config-provider-openstack-enable)
- * [**hal config provider oraclebmcs**](#hal-config-provider-oraclebmcs)
```
lwander: I'm worried this kind of renaming will break any existing users... is this necessary?

markxnelson: It is not strictly necessary, but it is highly desirable from our point of view.  The original name was probably an unfortunate choice, and was tied to a name that is no longer used.  So leaving it as it is could also cause possible confusion.  I would like to suffer any pain now while it is hopefully a smaller impact of changing this.

markxnelson: If this change would force people who do not even use that provider to take some action, like update their hal config file, then we would definitely not want to do that.

guoyongzhang:
The provider/storage config will read both old name (oraclebmcs) and new name (oracle), and consolidate them under just new name (oracle) for hal config to display and hal deploy to generate deployment files.  With this PR, spinnaker with old oraclebmcs provider/storage config should be ported to the new name "oracle" automatically.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Providers.java
```diff
@@ -46,16 +51,33 @@
...
   }
 
+  public OracleProvider getOracle() {
+    return OracleProvider.mergeOracleBMCSProvider(oracle, oraclebmcs);
+  }
+  
   @Override
   public NodeIterator getChildren() {
-    return NodeIteratorFactory.makeReflectiveIterator(this);
+    List<Node> nodes = new ArrayList<Node>();
```
lwander: Can we keep using reflective iterator, but maybe extend it to take preconditions? e.g. name != oraclebmcs. Otherwise this just adds to the overhead for adding a new provider.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/PersistentStorage.java
```diff
@@ -28,10 +34,27 @@ public void accept(ConfigProblemSetBuilder psBuilder, Validator v) {
...
+  
   @Override
   public NodeIterator getChildren() {
-    return NodeIteratorFactory.makeReflectiveIterator(this);
+    List<Node> nodes = new ArrayList<Node>();
+    nodes.add(azs);
+    nodes.add(gcs);
+    nodes.add(redis);
+    nodes.add(s3);
+    nodes.add(OraclePersistentStore.mergeOracleBMCSPersistentStore(oracle, oraclebmcs));
```
lwander: Same comment here about relying on the reflective iterator

guoyongzhang:
changes made according to the suggestion.

guoyongzhang:
Please review.  This PR contains multiple commits, should I squash them or the merge can do it?

markxnelson:
LGTM

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Providers.java
```diff
@@ -46,16 +51,35 @@
   GoogleProvider google = new GoogleProvider();
   KubernetesProvider kubernetes = new KubernetesProvider();
   OpenstackProvider openstack = new OpenstackProvider();
+  @JsonProperty(access = Access.WRITE_ONLY)
```
lwander: If this is write only, and someone has the written in their halconfig, how does halyard read it to do the merge?

lwander:
I can squash the commits as a part of the merge.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Providers.java
guoyongzhang: According to JsonProperty.Access javadoc:  WRITE_ONLY means that the property may only be written (set) for deserialization, but will not be read (get) on serialization, that is, the value of the property is not included in serialization.
So, the existing "oraclebmcs" will be deserialized into PersistenceStorage and will be consolidated with the deserialized "oracle" to produce a new "oracle" for serialization.  The serialized "persistentStorage" will not contain "oraclebmcs".

lwander: Nice! Thanks

PR #963, opened 2018-06-03
feat(authn/oauth2): Add commands for byo oauth2

awalker125:
This PR implements a BYO oauth config mechanism via halyard.

See

https://github.com/spinnaker/halyard/pull/956

https://github.com/spinnaker/spinnaker/issues/2856

https://github.com/spinnaker/spinnaker/issues/2664

Example byo comand

```
./hal config security authn oauth2 edit \
--client-id YYYYYYYYYYYYYYYYY \
--client-secret XXXXXX \
--provider byo \
--access-token-uri https://example.com/login/oauth/access_token \
--user-authorization-uri https://example.com/login/oauth/authorize \
--user-info-uri https://example.com/api/v3/user \
--scope all \
--client-authentication-scheme query \
--user-info-requirements foo=bar \
--user-info-requirements baz=qux \
--user-info-mapping-username username \
--user-info-mapping-first-name firstname \
--user-info-mapping-last-name lastname \
--user-info-mapping-email email```

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java
@@ -120,6 +120,19 @@ public class OAuth2 extends AuthnMethod {
         newUserInfoMapping.setFirstName("givenName");
         newUserInfoMapping.setLastName("surname");
         break;
+      case OTHER:
+        newClient.setAccessTokenUri(client.getAccessTokenUri());
+        newClient.setUserAuthorizationUri(client.getUserAuthorizationUri());
+        newClient.setScope(client.getScope());
+        newClient.setClientAuthenticationScheme(client.getClientAuthenticationScheme());
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by usrleon** [comment]
Current documentation says I must provide all data in gate-local.yml file, but according to example all data provided in command line. So what is the way?

**Comment by awalker125** [comment]
@usleon currently it must be done in gate-local.yml. If this gets approved then either should work.

File: halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java [hunk 1]
```diff
@@ -168,6 +181,7 @@ public String toString() {
     AZURE("azure"),
     GITHUB("github"),
     ORACLE("oracle"),
+    BYO("byo"),
```
ttomsu: Can we name this "other" or "custom"? Simply "byo" isn't clear that it means "bring your own".

awalker125:
I've changed byo to other

PR #965, opened 2018-06-04
feat(grpc): Update lognet library to enable health checking.

brandonnelson3:


brandonnelson3:
The lognet library now has builtin health checking for all registered services. 

$ grpc_cli call localhost:6565 grpc.health.v1.Health.Check 'service: "spinnaker.halyard.Deployments"':

connecting to localhost:6565
status: SERVING
Rpc succeeded with OK status

PR #966, opened 2018-06-08
feat(provider/kubernetes): caching policy support

lwander:
https://github.com/spinnaker/clouddriver/pull/2697


lwander:
https://github.com/spinnaker/spinnaker/issues/2879

PR #969, opened 2018-06-12
fix(version/bom): Remove version bom validation making sure version does not start with "branch:" or "local:".

joonlim:


joonlim:
@duftler @lwander Please review.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/VersionsService.java
```diff
@@ -68,22 +68,6 @@ public BillOfMaterials getBillOfMaterials(String version) {
...
-    if (Versions.isLocal(version)) {
-      throw new HalException(
-          new ConfigProblemBuilder(FATAL,
-              "Version prefixed with \"local:\" does not have a BOM")
-              .build()
-      );
-    }
-
     try {
       return profileRegistry.readBom(version);
```
lwander: What about putting these checks in the GCS profile registry's implementation of `readBom(..)`?

duftler: Sounds reasonable to me.

joonlim: The original fix was to make sure ```hal version bom``` returned a more helpful error message than '500' when the version is set to a version without a bom (e.g., branch:upstream/master). While we concluded that the root cause of a helpful message not being returned was the inability for Spring to accept an url-encoded forward slash character in a path variable, we decided that it made sense to add checks to make sure that the version did not start with "branch:" or "local:" anyway.

Now that we know that the checks for "branch:" and "local:" conflict with ```hal deploy apply``` I believe it makes sense to just remove it.

I have attempted to add the checks to the GCS profile registry's implementation, but ```hal version bom``` does not (always?) call that method.

PR #971, opened 2018-06-14
fix(install): Fix permissions of files installed by Halyard

ezimanyi:


ezimanyi:
@lwander : I want to make sure I understand why we're doing `chown $HAL_USER:halyard /opt/halyard` which confused me at first.  It looks like this is necessary to allow the `pid` file to be written into `/opt/halyard`...but given that (for typical `setuid`) the permissions on `/opt/halyard` will by 755, the `halyard` group won't actually have permission to create the `pid` file so you have to run Halyard as the `$HAL_USER`.  We could fix this by setting the permissions to `775` but it also makes me wonder if it might be better to have the `pid` file in `/var/run` as is more typical (and then completely avoid needing to `chown` the `/opt/halyard` directory).  This point is somewhat outside the scope of this PR (and I just ignored it for the change here) but it came to mind as I was reading this.

PR #973, opened 2018-06-14
fix(core): Move pidfile to /var/run/halyard

ezimanyi:


ezimanyi:
Also realized that we needed to add the `tar` arguments from https://github.com/spinnaker/halyard/pull/971 to `update-halyard` as well.

After thinking about this, we actually don't need to worry about a migration path---the existing behavior of `update-halyard` is to delete the entire `/opt/halyard` directory (including the pidfile), so the new installation is starting with a clean slate with no pidfile.

ezimanyi:
There's potentially some work that could be done to refactor common things out of `InstallHalyard.sh` and `update-halyard` but I've left that aside for now.

ezimanyi:
Also:
- Decided that it's best to avoid umask issues and explicitly do `chmod a+x` rather than `chmod +x`; similarly removed the `--no-same-permissions` flag I'd added in the last PR as that was causing none of the installed files to be readable/executable if you have a restrictive `umask`.

startup/debian/update-halyard
```diff
@@ -57,19 +57,40 @@ function install_halyard() {
...
   rm -rf $TEMPDIR
 }
 
+function create_pidfile_directory() {
+  # Use systemd-tmpfiles to ensure that the /var/run/halyard directory is created with proper
+  # permissions on each boot.
+  # As Ubuntu 14.04 does not support systemd-tmpfiles, fall back to using an upstart script to
+  # create the directory; this fallback can be removed when we no longer support Ubuntu 14.04
+  if systemd-tmpfiles --version &> /dev/null; then
+    echo "d /var/run/halyard 0755 $HAL_USER halyard" > /usr/lib/tmpfiles.d/halyard.conf
```
lwander: What's the rational for depending on either upstart or systemd?

ezimanyi: The `/run` filesystem (which `/var/run` symlinks to) is generally mounted as a `tmpfs` and is thus cleared on every boot, so we need to have the system create `/var/run/halyard` with the proper permissions each boot.

If `hal` were running as root, it could just make the directory on-the-fly if absent, but as it runs as a normal user we need to have some system process create the directory.  (This is similar to how daemons will often create a `/var/run/` directory during initialization before dropping privileges, except that here we're not actually starting the halyard daemon during startup.)

usrleon: 'hal deploy apply' must be run as root, because it eventually runs .hal/default/install.sh  which clearly says you must be root: "echo_err "$0 must be executed with root permissions; exiting""

ezimanyi: True, there are some configurations (local debian) where it needs to be run as root---but also some other configurations (ex: deploying to kubernetes) where it doesn't---so I'd rather not make this a requirement in all cases.

ezimanyi:
Going to close this for now, but keep the branch around to re-think this.

PR #974, opened 2018-06-14
feat(security): Moves Gate's management endpoints off of the default …

ttomsu:
…port (8084) onto a new port (8085) that is not exposed through the loadbalancer, and turns off security for this port.

This implements the settings described in https://github.com/spinnaker/gate/pull/557, and should fix https://github.com/spinnaker/spinnaker/issues/2765. This PR should not be submitted until that Gate PR is submitted.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/GateBoot154ProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/GateBoot154ProfileFactory.java
@@ -41,6 +41,13 @@ public class GateBoot154ProfileFactory extends GateProfileFactory {
       config.x509 = new X509Config(security);
     }
 
+    if (security.getAuthn().isEnabled()) {
+      config.management = new Management();
+      config.spring = new SpringConfig();
+      config.endpoints = new GateConfig.Endpoints();
+      gate.setHealthPort(config.management.port);
+    }
+
     return config;
   }
 }
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -1645,6 +1645,7 @@ hal config canary aws account add ACCOUNT [parameters]
  * `--access-key-id`: The default access key used to communicate with AWS.
  * `--bucket`: The name of a storage bucket that your specified account has access to. If you specify a globally unique bucket name that doesn't exist yet, Kayenta will create that bucket for you.
  * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
+ * `--endpoint`: The endpoint used to reach the service implementing the AWS api. Typical use is with Minio.
  * `--no-validate`: (*Default*: `false`) Skip validation.
  * `--root-folder`: The root folder in the chosen bucket to place all of the canary service's persistent data in (*Default*: `kayenta`).
  * `--secret-access-key`: (*Sensitive data* - user will be prompted on standard input) The secret key used to communicate with AWS.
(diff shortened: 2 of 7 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/GateBoot154ProfileFactory.java
```diff
@@ -41,6 +41,13 @@ protected GateConfig getGateConfig(ServiceSettings gate, Security security) {
       config.x509 = new X509Config(security);
     }
 
+    if (security.getAuthn().isEnabled()) {
+      config.management = new Management();
+      config.spring = new SpringConfig();
+      config.endpoints = new GateConfig.Endpoints();
+      gate.setHealthPort(config.management.port);
```
lwander: We will likely have to make a GateBoot19Profile factory since this change will only affect spinnaker releases after that point, right?

ttomsu:
This is an internal change as far as I'm concerned, so I don't see any
reason to split it out.

georgeliaw:
Any update on this? Would be helpful to have.

dbyron0:
I'm interested in this too, especially if it simplifies spinnaker-monitoring-daemon access when e.g. TLS + OAuth are enabled.  Potentially related to https://github.com/spinnaker/spinnaker/issues/2016.

lwander:
Wait, can we revisit the assumption that this needs a new profile factory? The corresponding change to gate was merged into 1.10 and this didn't break. This also only generates new config, it doesn't alter any existing config, so it doesn't look like that would break either. Can you confirm @ttomsu?

lwander:
@dbyron0 -- just spoke to @ttomsu offline. This doesn't simplify monitoring daemon access, it's just for the spring management endpoints.

The issue (https://github.com/spinnaker/spinnaker/issues/2016) you pointed out has the fix for monitoring access.

lwander:
I don't think so, but halyard automatically configures the monitoring daemon to use basic auth, while users can continue to use oauth + tls without knowing basic auth is enabled. I don't think this is a very invasive solution, but I could be misunderstanding.

dbyron0:
I just tried enabling basic auth in gate as described in https://github.com/spinnaker/spinnaker/issues/2016#issuecomment-437990487 and it does work.  It opens up basic auth access to the world as far as I can see though.  The port that now accepts basic auth is the same one deck uses, which means it has to be exposed, no?

ttomsu:
@dbyron0 Yes it opens up basic auth to the world, so I would suggest using a strong username/password combo. 

As @lwander mentioned, this PR doesn't address the metrics collection. The Metrics controller is not coded to be part of Spring's "management endoints" infrastructure, which is what this bug is referring to. "Managed endpoints" are things like `/health`, `/info`, `/env`, and `/resolvedEnv` - See https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

ttomsu:
I'm going to close this PR out in favor of a more generalized effort to move Managed Endpoints off the default port for all services: https://github.com/spinnaker/spinnaker/issues/3883

dbyron0:
Fair enough, but it still seems like a hole better not opened in the first place....And yes, I suppose I am commenting on the wrong ticket, but here we are.  Is converting the metrics controller to be a management endpoint a possibility?

ttomsu:
Filed https://github.com/spinnaker/spinnaker/issues/3901. Lets discuss there.

dbyron0:
Thanks.

PR #977, opened 2018-06-15
fix(install): Fix permissions of installed files

ezimanyi:


ezimanyi:
@lwander : I pulled just the permission-fixing changes from https://github.com/spinnaker/halyard/pull/973 so we can get those in and think some more about the move of the pidfile.

PR #979, opened 2018-06-21
fix(canary): Add validation so warning is issued if no CONFIGURATION/…

duftler:
…OBJECT_STORE is configured.


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/2935.

PR #980, opened 2018-06-25
feat(halyard): Added hal shutdown command.

brandonnelson3:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShutdownCommand.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+@Parameters(separators =  "=")
+public class ShutdownCommand extends NestableCommand {
+
+  @Getter(AccessLevel.PUBLIC)
+  private String commandName = "shutdown";
+
+  @Getter(AccessLevel.PUBLIC)
+  private String description = "Shutdown the halyard daemon.";
+
```
lwander: Extra newline?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShutdownCommand.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  protected void executeThis() {
+    Daemon.shutdown();
```
lwander: Should we print the shutdown message if it fails?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShutdownCommand.java
brandonnelson3: Done.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/ShutdownCommand.java
brandonnelson3: It already prints if halyard isn't up and running: 

$ hal shutdown
! ERROR Failed to connect to localhost/0:0:0:0:0:0:0:1:8064
? Is your daemon running?

lwander: Given that you check to see that the response from the daemon is the expected shutdown message, isn't it possible that the daemon sends something else when the shutdown fails?

PR #981, opened 2018-06-29
feat(provider/kubernetes): adding changes for aws eks

anshrma:


anshrma:
Related documentation updates --> https://github.com/spinnaker/spinnaker.github.io/pull/884

Dockerfile
```diff
@@ -22,6 +22,12 @@ RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/${KUBECTL
     chmod +x ./kubectl && \
     mv ./kubectl /usr/local/bin/kubectl
 
+RUN curl -o heptio-authenticator-aws https://amazon-eks.s3-us-west-2.amazonaws.com/1.10.3/2018-06-05/bin/linux/amd64/heptio-authenticator-aws && \
```
lwander: Is there a way to fetch the latest release like we do with kubectl above?

lwander:
I think the clouddriver pod will need this build change as well.

Dockerfile
anshrma: As the authenticator is pinned down to a specific version of kubectl + folder structure in S3 is pinned down to a specific release date, we cannot get the latest, however, good point, will parameterize this so that the authenticator url will not require change for future upgrades

lwander:
Sorry one last thing -- can you make this change here too? https://github.com/spinnaker/halyard/blob/master/Dockerfile.local I think this is the one we build in our nightlies.

lwander:
Thanks! Will merge & rebuild the `:nightly` tag now...

PR #983, opened 2018-07-06
fix(deploy): Gracefully handle non-existent services in localgit

ezimanyi:
Currently haylard will NPE when trying to deploy a localgit deployment of spinnaker that predates kayenta.


ezimanyi:
https://github.com/spinnaker/halyard/pull/918 fixed all the other deployment types, but looks like I missed localgit there.

PR #984, opened 2018-07-09
fix(storage/s3): use assume-role in storage s3 edit command

dbyron0:
See https://spinnakerteam.slack.com/archives/C3YBRLG8H/p1531172663000099.  Is there anything else required to wire this up?


lwander:
This needs a change in https://github.com/spinnaker/front50 (the persistent storage service) for this to work.

Thanks for submitting this!

dbyron0:
Sorry, I could use a hint about where in front50.  Part of me is thinking that this involves adding the logic from https://aws.amazon.com/blogs/aws/delegating-api-access-to-aws-services-using-iam-roles/ to https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/aws/AwsAccountValidator.java, and then changing https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/persistentStorage/S3Validator.java to call it properly, but then that doesn't involve front50.  The assume role code has gotta be _somewhere_ already...but I'm struggling to find it.

ezimanyi:
@dbyron0: I'm cleaning up old PRs...looks like this hasn't had activity in a while.  Mind if I close it, and you can reopen if/when you revisit this?  Also happy to help you move this forward if you're still interested in working on it.

dbyron0:
Yeah, let's close this for now.

PR #996, opened 2018-08-02
chore(install): configure command-completion on install

lwander:
The new install path was not generating the command-completion scripts.

fyi @ttomsu who found this

### DIFF
--- a/install/debian/InstallHalyard.sh
+++ b/install/debian/InstallHalyard.sh
@@ -91,6 +91,10 @@ function get_user() {
   echo $user
 }
 
+function get_home() {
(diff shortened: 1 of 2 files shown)


install/debian/InstallHalyard.sh
```diff
@@ -274,6 +278,43 @@ function install_java() {
   fi
 }
 
+function configure_bash_completion() {
+  local yes
+  echo ""
```
jtk54: Intentional?

lwander: Yup, for newline

PR #997, opened 2018-08-02
feat(spin): Adds new command to install spin CLI.

jtk54:
Installs latest version only. Will introduce commands to install from a BOM and from a spin version soon.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/SpinCommand.java
```diff
@@ -0,0 +1,24 @@
+package com.netflix.spinnaker.halyard.cli.command.v1;
```
lwander: Missing header

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/SpinController.java
```diff
@@ -0,0 +1,42 @@
...
+import com.netflix.spinnaker.halyard.core.tasks.v1.DaemonTaskHandler;
+import com.netflix.spinnaker.halyard.deploy.spin.v1.SpinService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/v1/spin")
+public class SpinController {
```
lwander: Should this be "spin" or "install" controller? Esp. if we added the "install" command to fetch other tools in the future

jtk54: The command structure is `hal spin <>`. We came up with three commands yesterday for things we'd need to support for `spin`: `install` (latest) `install --version` and `uninstall|remove`. Do we want to change that to `hal install|uninstall spin`?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/SpinCommand.java
jtk54: Added.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/SpinController.java
lwander: No I think your approach is better, `hal install...` might confuse people

PR #998, opened 2018-08-03
feat(docs): Clarify that clean doesnt delete hal config

sbwsg:
fixes https://github.com/spinnaker/spinnaker/issues/3110 though doesn't mention anything kubernetes-specific since this command is for all providers


lwander:
I think you need to rerun "make" in the halyard-cli directly since this creates a docs change as well

sbwsg:
:+1: fixed

PR #1000, opened 2018-08-03
feat(spin): Adds support for installing specific spin version.

jtk54:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
```diff
@@ -772,4 +772,7 @@
 
   @GET("/v1/spin/install/latest")
   DaemonTask<Halconfig, Object> installSpin();
+
+  @GET("/v1/spin/install/{version}")
```
lwander: Does this cause ambiguity in the response mapping? Is it even worth keeping the /latest endpoint?

lwander:
#1000 !!!

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
jtk54: This is a good point, `version=latest` is a real version. I can just version default to latest and simplify the controller, etc.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/spin/v1/InstallSpinCommand.java
```diff
@@ -45,8 +43,8 @@
   String version = "latest";
 
   @Override
-  protected OperationHandler<RemoteAction> getRemoteAction() {
-    return new OperationHandler<RemoteAction>()
+  protected OperationHandler<String> getRemoteAction() {
+    return new OperationHandler<String>()
```
lwander: Is this really desirable? This makes automation either copy-paste or piping into bash

jtk54: The argument here is that not every `spin` user will necessarily have `hal` installed since they won't necessarily be an admin. It seems unintuitive to need `hal` to acquire `spin`. We can go back to the `RemoteAction` plan, but then we'll need to support another path for users not using `hal` which also is annoying from a maintenance perspective.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spin/v1/SpinService.java
```diff
@@ -25,19 +25,12 @@
...
-        + "curl -LO https://storage.googleapis.com/spinnaker-artifacts/spin/$(curl -s https://storage.googleapis.com/spinnaker-artifacts/spin/latest)/linux/amd64/spin\n"
-        + "chmod +x spin\n"
-        + "sudo mv spin /usr/local/bin/spin";
-    result.setScript(script);
-    result.setScriptDescription("The generated script will install the latest version of spin CLI");
-    result.setAutoRun(true);
-    result.commitScript(halconfigDirectoryStructure.getSpinInstallScriptPath());
-    return result;
+  public String install(String version) {
+    String curlInstall = String.format("curl -LO https://storage.googleapis.com/spinnaker-artifacts/spin/%s/linux/amd64/spin\n",
```
lwander: We should have the script check if we're on darwin or not

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/spin/v1/InstallSpinCommand.java
lwander: That's true, but isn't that why we have https://www.spinnaker.io/guides/spin/cli/? 

Also, if we do want this, rather than introduce different behavior by default, can we put this behind `--print-script-only`?

jtk54:
We're opting to forego a surface in Halyard in favor or simplifying spin's distribution.

PR #1003, opened 2018-08-06
fix(authn/oauth) copy/paste bug in setting username in oauth user info mapping

got-root:
Fixing a small copy/paste error on setting the username field mapping in oauth. The expected behavior is that it sets the username field as entered in the CLI but due to a bug it instead sets it to the field specified for the last-name (if set).

Here's the command for reference:
https://www.spinnaker.io/reference/halyard/commands/#hal-config-security-authn-oauth2-edit


got-root:
Thanks @lwander, just updated the commit message.

PR #1006, opened 2018-08-16
feat(pubsub): Adds GCB as first-class message format for PubSub artif…

ttomsu:
…acts.

Do not merge until @lwander adds the jinja template to echo (link to PR to be added later).


ttomsu:
This is good to merge now?

PR #1008, opened 2018-08-17
feat(kubernetes): correct orca shutdown behavior

lwander:
This allows orca to deregister itself & wait while tasks are drained
during graceful termination


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
```diff
@@ -75,6 +84,14 @@ default boolean isEnabled(DeploymentConfiguration deploymentConfiguration) {
     return Arrays.asList("wget", "--spider", "-q", settings.getScheme() + "://localhost:" + settings.getPort() + settings.getHealthEndpoint());
   }
 
+  default boolean hasPrestopCommand() {
```
duftler: Shouldn't `Prestop` be `PreStop` everywhere?

PR #1019, opened 2018-08-27
fix(deploy/ha): Move settings of profiles from super to child classes for SpringServiceSettings to fix env initialization.

joonlim:
Fixing an issue I introduced in my last commit.

env is always initialized by sub-classes, so the setProfiles method only works when called directly from the subclass.


joonlim:
@lwander Fixing a bug I introduced.

PR #1020, opened 2018-08-27
fix(install): Fix syntax error in install script

ezimanyi:


ezimanyi:
Introduced in https://github.com/spinnaker/halyard/pull/1017
Fixes https://github.com/spinnaker/spinnaker/issues/3250

PR #1024, opened 2018-08-29
feat(deploy): Add default AWS settings to deck

mizzy:
Current halconfig/settings.js has fixed values for aws settings.
It should pick up settings configured by halyard.

This works with https://github.com/spinnaker/deck/pull/5681 .

Ref: https://github.com/spinnaker/spinnaker/issues/3271


mizzy:
Thanks for merging!

PR #1025, opened 2018-08-29
fix(deploy): Fix overrideBaseUrl field by favoring the field over baseUrl.

joonlim:


joonlim:
@lwander Fix for issue bburnett from slack is seeing.

lwander:
Rebuilding & publishing Halyard now

PR #1031, opened 2018-09-06
fix(#3268) Adds missing ldap config parameters managerDn managerPassw…

senare:
Why didn't it get linked to ?

https://github.com/spinnaker/spinnaker/issues/3268
 
Oki, I need to enable Ldap so trying to fix whats missing here

First time trying to contrib so some hints and tips be welcome, thanx in advance !

I think this should be easy and straight forward but I am still trying to figure out how to build and test all to verify this, but I am opening the PR early/now anyway to gather some feedback and help

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -9286,6 +9286,9 @@ hal config security authn ldap edit [parameters]
 
 #### Parameters
  * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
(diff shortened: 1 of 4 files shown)


senare:
Also related  https://github.com/spinnaker/spinnaker/issues/2570 ?

ajordens:
Can we get this resolved and merged down @stewchen?

PR #1032, opened 2018-09-06
fix(deploy): remove illegal version characters from label

lwander:
https://github.com/spinnaker/spinnaker/issues/3297


ethanfrogers:
@lwander @danielpeach have you all validated this yet? i'm attempting to do the same thing as #3297 and getting the same error with this fix. 

for some odd reason, `value.replaceAll("/[^A-Za-z0-9-_.]/", "")` doesn't replace any `:`.

PR #1033, opened 2018-09-07
fix(cli/accounts): Allows specifying environments for accounts

polarbizzle:
The `credentials.challengeDestructiveActionsEnvironments` setting in
clouddriver controls a few features in deck. This setting is based off
the configured environment for an account. Adding environment to the
add and edit account commands allows use of halyard to take advantage of
those deck features.

Moving the environment flag up from the Openstack subclasses makes it
available to all account types.


polarbizzle:
We're using a clouddriver-local.yml profile to handle setting challengeDestructiveActionsEnvironments.  Perhaps someday that should make it into the cli as well, but this setup is working well for us now

polarbizzle:
No problem -- done

PR #1034, opened 2018-09-10
feat(viewAllDeployments): Lists all deployments Halyard can see.

bpowell:


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ListCommand.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Getter(AccessLevel.PUBLIC)
+    private String description = "Lists all deployments";
+
+    @Override
+    protected void executeThis() {
+        new OperationHandler<List<DeploymentConfiguration>>()
+            .setFailureMesssage("Failed to get all deployments.")
+            .setSuccessMessage("Retrieved all deployments.")
+            .setFormat(AnsiFormatUtils.Format.JSON)
```
lwander: The default for other commands has been either YAML or a special string -- any reason to choose JSON?

bpowell: Yeah, YAML was giving me trouble displaying the deployments. Was getting this error:
```
! ERROR Cannot deserialize instance of `java.util.LinkedHashMap`
  out of START_ARRAY token
 at [Source: UNKNOWN; line: -1, column: -1]
```

lwander: Ohh we might be assuming that the object being deseralized is always a map, and in this case it's a list. Could you add a check in the deserializer to handle that correctly?

PR #1035, opened 2018-09-11
fix(config): unsupported class

bpowell:


bpowell:
@lwander this fixes a problem that I introduced with https://github.com/spinnaker/halyard/pull/1034 

When running `hal config` it throws an error due to only checking for `List`s and `Map`s.

PR #1037, opened 2018-09-14
fix(deploy): fix regex for illegal char replace

ethanfrogers:
the regex used in `replaceAll` in `makeValidLabel` was incorrect not
replacing chararacters. the intent was to remove all characters that are
not `A-Za-z0-9-_.` but using a leading and trailing `/` broke it. by
removing the `/`, we get the desired effect.

@lwander PTAL :)


lwander:
Thanks!

PR #1038, opened 2018-09-17
fix(ha): Rename echo-slave to echo-replica.

joonlim:


joonlim:
@jtk54 @lwander Renaming 'echo-slave' to 'echo-replica'

PR #1039, opened 2018-09-18
fix(install): Halyard startup should use a login shell

ezimanyi:
On Ubuntu 18.04 on GCE, gcloud is installed by snap in /snap/bin. There is an entry to add /snap/bin to the path in /etc/profile.d but this only runs for login shells, so if halyard runs in a non-login shell it is unable to find gcloud. Rather than address the specific case, we should start the halyard daemon in a login shell.

### DIFF
--- a/install/debian/InstallHalyard.sh
+++ b/install/debian/InstallHalyard.sh
@@ -368,6 +368,6 @@ configure_defaults
 install_java
 install_halyard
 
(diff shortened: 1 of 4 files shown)


ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/3349

lwander:
Nice fix

PR #1042, opened 2018-09-27
 fix(localgit): Fix SSL certificates for localgit deployment

ezimanyi:


ezimanyi:
First commit is just updating the autogenerated `commands.md` file as it looks like some recent changes didn't update/commit that file.

Second commit is the actual change.  For webpack, it appears you need to pass in the CA certificate to the config, so added a parameter for that and set it as an environment variable.  These variables are read here:
https://github.com/spinnaker/deck/blob/2125de975c149537e5eaeb8763016e6ce731f9ef/webpack.config.js#L177

PR #1046, opened 2018-10-01
fix(authz): allow anonymous LDAP bind

ncknt:
This PR is to allow anonymous LDAP search (managerDn and manager password both empty). Fiat supports it but Halyard doesn't allow that configuration.


lwander:
Thanks @ncknt!

PR #1047, opened 2018-10-02
fix(config/pubsub): Set pubsub global enabled based on child config.

jtk54:
@ezimanyi FYI.


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Pubsubs.java
```diff
@@ -42,6 +43,26 @@ public NodeIterator getChildren() {
...
+  }
+
+  public void setEnabled() {
+    NodeIterator pubsubNodes = getChildren();
+    Pubsub pubsub;
+    Boolean allEnabled = true;
+    while ((pubsub = (Pubsub) pubsubNodes.getNext()) != null) {
+      allEnabled &= pubsub.isEnabled();
+    }
+    enabled = allEnabled;
```
lwander: Why not just call `getEnabled()`?

PR #1049, opened 2018-10-03
 config(Datadog): Add custom Datadog tags option

shividhar:
Currently Spinnaker supports the use of custom Datadog tags set by the following [PR](https://github.com/spinnaker/spinnaker-monitoring/pull/71). This PR seeks to allow users to set custom Datadog tags with Halyard so they can be consumed by Spinnaker. 

I also re-rendered the docs for Halyard and fixed a mistake in the docs.


shividhar:
cc @lwander @ttomsu

shividhar:
cc @ezimanyi

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/metricStores/datadog/DatadogStore.java
```diff
@@ -42,5 +45,5 @@ public String getNodeName() {
   private String appKey;
 
   @JsonProperty("tags")
-  private String tags;
+  private List<String> tags = new ArrayList<>();
```
lwander: Does halyard complain during an upgrade that the schema changed?

shividhar: Don't understand what you mean by upgrade exactly, but I rebuilt the Halyard Binary and tested the changes locally to see if the Halconfig was correctly generated. I didn't see any errors or warnings with the changes.

lwander: If your halconfig has `tags: <string>`, and you upgrade to include this patch, does halyard complain about deserialization failing because of a mismatched type?

shividhar: The Halyard master  branch doesn't have `tags` as a [member variable](https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/metricStores/datadog/DatadogStore.java#L43). The Github UI seems to have messed that up 

If you click on the `Files Changed` tab then you won't see the redaction of the `tags` member.

lwander: Ah got it, looks good to me!

lwander:
Thanks @shividhar

PR #1051, opened 2018-10-03
fix(kubernetes/health): Fix Gate health check with SSL enabled

ezimanyi:
The gate health check is currently broken when SSL is enabled. The request is to the localhost domain and does not pass in any additional trusted CA's; the request will thus fail unless gate presents a certificate for localhost issued by a root CA (which I believe will never be validly issued). As the request is within the pod to localhost, validating the certificate does not have high value, so just pass a flag to wget to skip certificate validation.


ezimanyi:
Ran into this myself, but I believe it closes https://github.com/spinnaker/spinnaker/issues/2566

PR #1052, opened 2018-10-03
chore(ha): Add comments indicating that redis url must contain protocol and port.

joonlim:


joonlim:
@jtk54 @lwander

joonlim:
@lwander Updated to point to an online standard describing the Redis URI schema in detail: https://www.iana.org/assignments/uri-schemes/prov/redis.

The standard describes some cases that I haven't considered. For example, a Redis URI may start with "rediss://", which indicates that the Redis uses SSL.

joonlim:
@lwander ping

PR #1059, opened 2018-10-05
fix(kubernetes): Fix container artifact ids

ezimanyi:


ezimanyi:
Closes https://github.com/spinnaker/spinnaker/issues/3427

PR #1061, opened 2018-10-05
fix(igor/travis): Set artifact.decorator.enabled: true when Travis CI is enabled

mrampton:
@gardleopard @lwander 

This is a fix for Igor finding itself in a non-starting state once Travis is enabled. The manual workaround is detailed here: https://github.com/spinnaker/spinnaker/issues/2918

Let me know if you would prefer it be fixed in another manner (hardcoding the configuration in igor.yml, for example).


lwander:
If we merge this, and want to make this (artifact decoration) configurable in the future without breaking users halconfigs, this will get a little tricky. I think the "more correct" fix is to allow igor to start with only travis enabled but without the artifact decoration enabled.

mrampton:
Thanks for the input Lars. @gardleopard just put in a fix to Igor yesterday that made the decorators optional for Travis. Closing this PR as a result.

PR #1062, opened 2018-10-05
chore(dependencies): Bump spinnaker-dependencies

ezimanyi:


ezimanyi:
Needed to update the GCS validator to handle a change in the constructor due to:
* Retry logic (added in https://github.com/spinnaker/front50/pull/191)
* Scheduled retries (added in https://github.com/spinnaker/front50/pull/191)

I just used the default retry parameters, as given this is Halyard which generates the config files, we don't have a config file to read (as we do in front50).

ezimanyi:
This fixes the test that broke in the auto-update done in https://github.com/spinnaker/halyard/pull/1060

ezimanyi:
Mind taking a look @lwander?

PR #1066, opened 2018-10-12
fix(provider/kubernetes): v1 - pass account service account to kubecfg parser

sbwsg:
Making a PR for discussion on Monday.

Issue here: https://github.com/spinnaker/spinnaker/issues/3435#issuecomment-429375735

This user of the spinnaker helm chart is attempting to deploy a v1 account but seeing an error because the kubecfg is null. It looks like the kubecfg parser can deal with this by using the provided service account instead but needs to be instructed to do so.

I'm having some trouble compiling this to test against.  Getting the following gradle error:

```
 (master) λ ./gradlew clean
Inferred project: halyard, version: 0.37.0-SNAPSHOT

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/sbws/dev/spinnaker/halyard/build.gradle' line: 36

* What went wrong:
A problem occurred evaluating root project 'halyard'.
> Failed to apply plugin [class 'com.netflix.spinnaker.gradle.project.SpinnakerProjectConventionsPlugin']
   > index is out of range 0..-1 (index = 0)
```


lwander:
Nice, thanks!

PR #1076, opened 2018-10-18
fix(deploy): fix deploy w/o aws region specified

ablundgren:
The previous patch didn't get the bug, it is on an empty list, not an empty name.  This checks both.

https://github.com/spinnaker/spinnaker/issues/3280


sbwsg:
Thanks!

PR #1081, opened 2018-10-22
fix(run): Pass argument escaping through startup script

danopia:
The startup script is breaking apart arguments with spaces in them. Spaces in arguments become important when configuring docker-registry with a `--password-command`. The lack of argument escaping was causing errors as actual hal tried to parse the password-command arguments as its own.

Brief explanation of the specific shell syntax: https://stackoverflow.com/a/905910

### DIFF
--- a/startup/debian/hal
+++ b/startup/debian/hal
@@ -63,4 +63,4 @@ else
     fi
 fi
 
-$HAL $@
+$HAL "$@"
--- a/startup/macos/hal
+++ b/startup/macos/hal
@@ -63,4 +63,4 @@ else
     fi
 fi
 
-$HAL $@
+$HAL "$@"


danopia:
Here's a command I've been using to repro:

```
# wrapper script in $PATH, breaks up argument
$ /usr/local/bin/hal config deploy edit --type "distributed af"
Was passed main parameter 'af' but no main parameter was defined

# actual binary, handles argument correctly
$ /opt/halyard/bin/hal config deploy edit --type "distributed af"
DeploymentType "distributed af" is not a valid choice. The options are: [Distributed, LocalDebian, LocalGit, BakeDebian]
```

avram:
@edwinavalos I believe you last touched this bit; does this make sense for you?

PR #1090, opened 2018-10-25
feat(halyard): Add support for deprecating flags

sbwsg:
Twin of https://github.com/spinnaker/deck/pull/5894

Appengine Container URL deploys will now always be available in deck.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
@@ -47,19 +47,19 @@ public class KubernetesAccount extends ContainerAccount implements Cloneable {
   String context;
   String cluster;
   String user;
-  @ValidForSpinnakerVersion(lowerBound = "1.5.0", message = "Spinnaker does not support configuring this behavior before that version.")
+  @ValidForSpinnakerVersion(lowerBound = "1.5.0", tooLowMessage = "Spinnaker does not support configuring this behavior before that version.")
   Boolean configureImagePullSecrets;
   Boolean serviceAccount;
   int cacheThreads = 1;
   List<String> namespaces = new ArrayList<>();
   List<String> omitNamespaces = new ArrayList<>();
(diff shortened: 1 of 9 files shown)


sbwsg:
Does Hal have a flag deprecation process?  If a consumer is using this feature flag in a script then I suppose that script will now be broken.

lwander:
Not yet for flags, but yet for commands: https://github.com/spinnaker/halyard/blob/master/docs/deprecating-commands.md

What I would do is look at how @danielpeach implemented the ValidForSpinnakerVersion annotation, and add "upperBound" support: https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java#L62

sbwsg:
@lwander @danielpeach I've extended the ValidForSpinnakerVersion annotation to add a `upperBound` and `tooHighMessage`.

It manifests as follows:

screen shot 2018-10-25 at 3 48 01 pm

PTAL

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccount.java
```diff
@@ -47,19 +47,19 @@
   String context;
   String cluster;
   String user;
-  @ValidForSpinnakerVersion(lowerBound = "1.5.0", message = "Spinnaker does not support configuring this behavior before that version.")
+  @ValidForSpinnakerVersion(lowerBound = "1.5.0", upperBound = "", tooLowMessage = "Spinnaker does not support configuring this behavior before that version.", tooHighMessage = "")
```
lwander: I had assumed there was a way to keep parameters default/non-required -- do we need this `upperBound = ""` for each use of the annotation?

sbwsg: I likewise assumed I could declare some `default` methods (ala `getUpperBound()`, `setUpperBound(String)`) on the interface but wasn't able to get the syntax into a format that intellij accepted.

sbwsg: Ach, got it, @danielpeach pointed me towards how to declare defaults in annotation.

PR #1099, opened 2018-11-06
Add aws cli to container build

jhohertz:
Per discussion in spinnaker/spinnaker#2394, adding awscli to container
such that the features added in spinnaker/spinnaker#2941 to enable
an external passwordCommand for a docker registry can be utilized by
AWS users wishing to add their ECR-based docker registries.

Signed-off-by: Joe Hohertz <joe@viafoura.com>


lwander:
LGTM

PR #1100, opened 2018-11-07
fix(kayenta/google): Do not unnecessarily require compute-ro scope.

duftler:


duftler:
Fixes https://github.com/spinnaker/kayenta/issues/352.

ttomsu:
I don't see anything in the code that would lead me to believe those changes had anything to do with scopes, let alone the `compute-ro` scope.

PR #1101, opened 2018-11-12
fix(deploy): null-safe rosco status check

lwander:


lwander:
Not a full fix for: https://github.com/spinnaker/spinnaker/issues/3577, but at least addresses a symptom.

PR #1102, opened 2018-11-12
feat(appengine): add support for filtering appengine services and versions

sbwsg:
Configuration for https://github.com/spinnaker/spinnaker/issues/3539

Clouddriver implementation: https://github.com/spinnaker/clouddriver/pull/3150

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/appengine/AppengineCommandProperties.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/appengine/AppengineCommandProperties.java
@@ -37,4 +37,12 @@ public class AppengineCommandProperties {
             + " against a known_hosts file.";
     static final String GCLOUD_RELEASE_TRACK = "The gcloud release track (ALPHA, BETA, or STABLE) that Spinnaker"
             + " will use when deploying to App Engine.";
+    static final String SERVICES = "A list of regular expressions. Any service matching one of these regexes " +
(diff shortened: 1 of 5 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/appengine/AppengineCommandProperties.java
```diff
@@ -37,4 +37,8 @@
             + " against a known_hosts file.";
     static final String GCLOUD_RELEASE_TRACK = "The gcloud release track (ALPHA, BETA, or STABLE) that Spinnaker"
             + " will use when deploying to App Engine.";
+    static final String SERVICES = "The list of services to which Spinnaker should limit its indexing.";
```
lwander: I realize we don't do this yet, but is it possible to use these to prevent a user from modifying/deploying a certain service or version in an account? We do the same for namespaces & kinds in k8s for both the indexing & deploying.

sbwsg:
todo: add minimum spinnaker version annotations

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/appengine/AppengineCommandProperties.java
sbwsg: Interesting idea. Quickly looking at DeployAppengineAtomicOperation this appears to be doable. I've created an issue to track this: https://github.com/spinnaker/spinnaker/issues/3590

PR #1111, opened 2018-11-29
feat(deploy): Handle spinnaker-monitoring-local.yml

ewiseblatt:
@lwander 
It isnt clear to me how the interfaces work so I'm just guessing here.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
@@ -75,6 +75,10 @@ abstract public class SpinnakerMonitoringDaemonService extends SpinnakerService<
     return "spinnaker-monitoring.yml";
   }
 
+  private static String monitoringLocalProfileName() {
+    return "spinnaker-monitoring-local.yml";
+  }
+
   private static String defaultFilterProfileName() {
     return "monitoring-daemon/filters/default.yml";
   }
@@ -97,6 +101,7 @@ abstract public class SpinnakerMonitoringDaemonService extends SpinnakerService<
     String profilePath = Paths.get(CONFIG_OUTPUT_PATH, profileName).toString();
 
     results.add(spinnakerMonitoringDaemonProfileFactory.getProfile(profileName, profilePath, deploymentConfiguration, endpoints));
+
     return results;
   }
 
@@ -111,6 +116,11 @@ abstract public class SpinnakerMonitoringDaemonService extends SpinnakerService<
 
     profile = monitoringProfiles.get(monitoringProfileName());
     result.add(profile);
+
+    profile = monitoringProfiles.get(monitoringLocalProfileName());
+    if (profile != null) {
+      result.add(profile);
+    }
 
     profile = monitoringProfiles.get(defaultFilterProfileName());
     if (profile != null) {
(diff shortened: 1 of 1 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
```diff
@@ -97,6 +101,11 @@ private static String defaultFilterProfileName() {
     String profilePath = Paths.get(CONFIG_OUTPUT_PATH, profileName).toString();
 
     results.add(spinnakerMonitoringDaemonProfileFactory.getProfile(profileName, profilePath, deploymentConfiguration, endpoints));
+
+    profileName = monitoringLocalProfileName();
+    profilePath = Paths.get(CONFIG_OUTPUT_PATH, profileName).toString();
+    results.add(spinnakerMonitoringDaemonProfileFactory.getProfile(profileName, profilePath, deploymentConfiguration, endpoints));
```
lwander: You're generating the same profile here as the non-local profile. Is this intentional?

ewiseblatt:
No. I wanted to deposit the hand curated -local.yml file if there was one.
I guess I was right about not figuring out the interface!

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
```diff
@@ -112,6 +121,9 @@ private static String defaultFilterProfileName() {
     profile = monitoringProfiles.get(monitoringProfileName());
     result.add(profile);
 
+    profile = monitoringProfiles.get(monitoringLocalProfileName());
+    result.add(profile);
```
lwander: this might be null, if the user didn't supply it.

lwander:
Ok, just to make sure I understand: you don't want to generate a `-local` profile, only pick it up if a user has put it into ~/.hal/default/profiles/spinnaker-monitoring-local.yml?

If that's the case, you need to do the same thing that @brandonnelson3 did here to support custom filter paths: https://github.com/spinnaker/halyard/pull/1111/files#diff-18741187809b5ae5f6973f266036be07R155

In effect, you're having the service decide which files in that profiles directory are relevant to it, and if they are return Optional.of(<path>).

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
ewiseblatt: Removed.

ewiseblatt:
@ping

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerMonitoringDaemonService.java
```diff
@@ -75,6 +75,10 @@ private static String monitoringProfileName() {
     return "spinnaker-monitoring.yml";
   }
 
+  private static String monitoringLocalProfileName() {
+    return "spinnaker-monitoring.yml";
```
lwander: Shouldn't this be `spinnaker-monitoring-local.yml`?

ewiseblatt: good catch

PR #1115, opened 2018-12-03
Add nodeSelector in .hal/config file kubernetes v1

jtdawso:
Config will now allow nodeSelectors to be added to the deployed spinnaker nodes. (Tested in kubernetes v1)

```
  deploymentEnvironment:
    ...
    nodeSelectors:
       service_type: api

```


ethanfrogers:
@jtdawso any chance you could add this for V2 as well? Seems like it'd be useful.

jtdawso:
@ethanfrogers  I haven't done anything with v2 kubernetes, and don't have a way to test it, so it would be close to a stab in the dark for me to try it.

jtdawso:
The force push was to fix the commit message to follow convention, I'm not sure if I edited the comment the recommend way. But let me know if there's a better way

lwander:
thanks!

PR #1117, opened 2018-12-04
fix(config/orca): set global.spinnaker.timezone for orca

maggieneterval:
- Fixes bug where Restrict Execution During Time Window stages always calculate whether the current time is within range using PST rather than a user's configured timezone


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
```diff
@@ -45,5 +45,7 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
     profile.appendContents("pipelineTemplates.enabled: " + pipelineTemplates);
     // For backward compatibility
     profile.appendContents("pipelineTemplate.enabled: " + pipelineTemplates);
+
+    profile.appendContents("global.spinnaker.timezone: " + deploymentConfiguration.getTimezone());
```
jtk54: Should this go in Orca's config or should this be inherited from the top level spinnaker.yml?

maggieneterval: Good call, moved to `SpinnakerProfileFactory`!

PR #1118, opened 2018-12-05
 refactor(webhooks): A few minor refactors to the webhook controllers

ezimanyi:
A few minor refactors to the webhook controllers; probably easier to read each commit separately as they are primarily doing the same trivial change to each of the controllers.  This is to make the diff in an upcoming PR to clean up duplicated code cleaner.


halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/AccountController.java
```diff
@@ -97,43 +88,40 @@
...
   }
 
   @RequestMapping(value = "/account/{accountName:.+}/options", method = RequestMethod.PUT)
-  DaemonTask<Halconfig, List<String>> existingAccountOptions(
-      @PathVariable String deploymentName,
+  DaemonTask<Halconfig, List<String>> existingAccountOptions(@PathVariable String deploymentName,
       @PathVariable String providerName,
       @PathVariable String accountName,
-      @RequestParam(required = false, defaultValue = DefaultValidationSettings.severity) Severity severity,
+      @ModelAttribute ValidationSettings validationSettings,
```
lwander: How are ModelAttributes passed to the controller from an HTTP perspective?

ezimanyi: They are initialized in a function annotated as `@ModelAttribute` in a class annotated as `@ControllerAdvice`; in this case [here](https://github.com/spinnaker/halyard/blob/fb75291ed92206a0cec4ce7a07ed902c4b94569a/halyard-web/src/main/java/com/netflix/spinnaker/halyard/models/v1/ValidationSettingsHandler.java#L25)

ezimanyi: Yes, the `@ModelAttribute`-annotated method runs before the handler methods to populate model attributes, which can then be referenced in the handler method.  In this case, it's to encapsulate the common parameters for all handlers into an object rather than have them separately in all handlers.

ezimanyi: And in this case the `@ModelAttribute` method is just reading from the request params, so the interface is the same for the client.

lwander: Nice :+1:

PR #1129, opened 2018-12-12
feat(kubernetes/v2): Wait for containers to be ready before returning from hal deploy apply for kubernetes v2 provider.

joonlim:
@lwander


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Utils.java
```diff
@@ -105,6 +105,61 @@ static private boolean exists(KubernetesAccount account, String namespace, Strin
...
+    }
+
+    if (status.getState() != JobStatus.State.COMPLETED) {
+      throw new HalException(Problem.Severity.FATAL, String.join("\n",
+          "Unterminated readiness check for " + service + " in " + namespace,
+          status.getStdErr(),
+          status.getStdOut()));
+    }
+
+    if (status.getResult() == JobStatus.Result.SUCCESS) {
```
lwander: I think this can fail if no containers could be found -- which might happen immediately after a deployment.

joonlim: Upon experimentation I found that if no containers exist the status result will still be SUCCESS, with stdout containing two single quotes ('').

lwander:
https://github.com/spinnaker/spinnaker/issues/2288

PR #1130, opened 2018-12-13
Allow specifying of serviceAccountName

abergmeier:
At least in OpenShift, we need to run certain Services under a
serviceAccountName, in order to be able to run containers under
anyuid.


abergmeier:
There are at least some document changes necessary. Anyone care to guide what is missing still?

PR #1144, opened 2019-01-08
fix(cli): add password flag to password fields

KathrynLewis:
A few fields in the CLI seem to be missing the "password = true" flag.


KathrynLewis:
Not a high priority, especially if it will cause issues. I just noticed those fields that should probably have that flag.

PR #1149, opened 2019-01-09
feat(deploy/kubernetes): add serviceType to kubernetes service settings

jgramoll:
kubernetes:
  serviceType: NodePort
  nodePort: 30085

### DIFF
--- a/halyard-deploy/src/main/resources/kubernetes/manifests/service.yml
+++ b/halyard-deploy/src/main/resources/kubernetes/manifests/service.yml
@@ -10,7 +10,11 @@ spec:
   selector:
     app: spin
     cluster: spin-{{ name }}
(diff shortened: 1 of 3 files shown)


halyard-deploy/src/main/resources/kubernetes/manifests/service.yml
```diff
@@ -10,7 +10,11 @@ spec:
   selector:
     app: spin
     cluster: spin-{{ name }}
+  type: {{ type }}
   ports:
   - protocol: TCP
     port: {{ port }}
     targetPort: {{ port }}
+  {% if nodePort != null %}
```
Z3r0Sum: you probably also want to check to see if the type is `NodePort` and only set this if both those characteristics are true.

PR #1150, opened 2019-01-11
feat(canary/prometheus): Add basic auth support.

duftler:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/canary/prometheus/PrometheusCanaryAccountValidator.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2018 Google, Inc.
```
lwander: Date on copyrights is wrong

PR #1151, opened 2019-01-14
chore(deploy): Extract processing of Volumes

abergmeier:
New KubernetesV2Service.getVolumes


abergmeier:
There seems to be currently no test associated. Not sure how one should test the provider the best. Maybe spawn Minikube?

abergmeier:
Added a unit test so there is at least bare minimum testing.

PR #1152, opened 2019-01-15
fix(redis): For Kubernetes V2 provider, do not attempt to redeploy Redis deployment if it already exists.

joonlim:


joonlim:
@lwander

PR #1154, opened 2019-01-15
fix(versioning): Ensure releases (nightly and stable) include the Neb…

ttomsu:
…ula module that injects the version metadata.

Fixes https://github.com/spinnaker/spinnaker/issues/3720


ttomsu:
@spinnakerbot cherry-pick 1.13

PR #1157, opened 2019-01-16
feat(config/secrets): add SecretSession and SecretSessionManager

KathrynLewis:
Part of secrets management project. 

Note: these classes will require the korkSecrets dependency in order to build.

[spinnaker/spinnaker#3649](https://github.com/spinnaker/spinnaker/issues/3649)

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/secrets/SecretSession.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/secrets/SecretSession.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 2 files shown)


lwander:
QQ: why put these in halyard rather than kork? As I understand it, these will be used by both the services & halyard where appropriate.

ncknt:
@lwander This adds caching and secret file tracking. The thought was that Halyard will need to decrypt the same thing multiple times whereas services may not. That was before learning about `archaius` which seems to check the config multiple times. So we've added caching to `kork/SecretManager` as well (not in a PR yet). Let us clean that up.

KathrynLewis:
@lwander We've moved the caching into the SecretManager class in kork. See [this PR](https://github.com/spinnaker/kork/pull/225). The update here depends on that PR being merged. Let me know if anything else should be changed here. Thanks!

lwander:
LGTM, you may need to rebase to force a rebuild against the current HEAD

PR #1160, opened 2019-01-17
 feat(secrets): Add object mapper to decrypt secrets

ncknt:
WIP: Depends on https://github.com/spinnaker/halyard/pull/1157 and https://github.com/spinnaker/halyard/pull/1159

Implement spinnaker/spinnaker#3649

Object mapper to be used to send config to services that don't support decryption. When secrets are in separate files, the content is kept in memory and tracked in the Profile.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
@@ -53,12 +53,14 @@ public class DecryptingObjectMapper extends ObjectMapper {
 
     protected Profile profile;
     protected Path decryptedOutputDirectory;
+    protected SecretSessionManager secretSessionManager;
 
     public DecryptingObjectMapper(SecretSessionManager secretSessionManager, Profile profile, Path decryptedOutputDirectory) {
         super();
         this.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, true);
         this.setSerializationInclusion(JsonInclude.Include.NON_NULL);
 
+        this.secretSessionManager = secretSessionManager;
         this.profile = profile;
         this.decryptedOutputDirectory = decryptedOutputDirectory;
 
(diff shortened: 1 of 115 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
```diff
@@ -0,0 +1,151 @@
...
+                }
+                return SecretType.NO_SECRET;
+            }
+        }
+        if (_class.getSuperclass() != null) {
+            return getFieldSecretType(_class.getSuperclass(), fieldName);
+        }
+        return SecretType.NO_SECRET;
+    }
+
```
lwander: extraneous newline

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
```diff
@@ -0,0 +1,151 @@
...
+        this.decryptedOutputDirectory = decryptedOutputDirectory;
+
+        SimpleModule module = new SimpleModule();
+        module.setSerializerModifier(new BeanSerializerModifier() {
+
+            @Override
+            public List<BeanPropertyWriter> changeProperties(SerializationConfig config, BeanDescription beanDesc, List<BeanPropertyWriter> beanProperties) {
+                Class _class = beanDesc.getBeanClass();
+
+                for (Iterator<BeanPropertyWriter> it = beanProperties.iterator(); it.hasNext(); ) {
```
lwander: I think the preferred way to write this is

`for (BeanPropertyWriter it : beanProperties) {`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
```diff
@@ -0,0 +1,151 @@
...
+                    BeanPropertyWriter bpw = it.next();
+
+                    switch (getFieldSecretType(_class, bpw.getName())) {
+                        case SECRET:
+                            // Decrypt the field secret before sending
+                            bpw.assignSerializer(new StdScalarSerializer<Object>(String.class, false) {
```
lwander: To avoid this really deep nesting, could we define/create functions to return these anon class at the top level in this class?

PR #1164, opened 2019-01-17
feat(secrets): Don't stage dependencies with secrets

ncknt:
WIP
Depends on #1157 and #1159
Implementing spinnaker/spinnaker#3649

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.halyard.config.model.v1.node;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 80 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -176,14 +177,27 @@ public void parentify() {
 
     while (clazz != null) {
       res.addAll(Arrays.stream(clazz.getDeclaredFields())
-          .filter(f -> f.getDeclaredAnnotation(LocalFile.class) != null)
+              .filter(f -> f.getDeclaredAnnotation(LocalFile.class) != null && !isSecretFile(f))
```
lwander: If someone doesn't want to use the secret mgmt, does this break anything? We're now omitting a number of files that used to be included.

ncknt: It shouldn't. `isSecretFile` returns true only if the field is annotated with @SecretFile AND the content is a secret (or looks like one, `encrypted!...`)

PR #1165, opened 2019-01-17
feat(config/secrets): decrypt secret fields and files in validators

KathrynLewis:
Decrypting secret fields and files in validators.

Depends on PR: https://github.com/spinnaker/halyard/pull/1157

Part of secrets management: [spinnaker/spinnaker#3649](https://github.com/spinnaker/spinnaker/issues/3649)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidator.java
```diff
@@ -107,9 +110,9 @@ public void validate(ConfigProblemSetBuilder p, DockerRegistryAccount n) {
           .accountName(n.getName())
           .address(n.getAddress())
           .email(n.getEmail())
-          .password(n.getPassword())
+          .password(secretSessionManager.decrypt(n.getPassword()))
```
link108: you should be able to use `resolvedPassword` here, since the password is already decrypted

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidator.java
```diff
@@ -53,9 +56,9 @@ public void validate(ConfigProblemSetBuilder p, DockerRegistryAccount n) {
     }
 
     if (passwordProvided) {
-      resolvedPassword = password;
+      resolvedPassword = secretSessionManager.decrypt(password);
     } else if (passwordFileProvided) {
-      resolvedPassword = ValidatingFileReader.contents(p, passwordFile);
+      resolvedPassword = secretSessionManager.decrypt(ValidatingFileReader.contents(p, passwordFile));
```
link108: The value of `passwordFile` may be an encrypted secret. `secretSessionManager.decryptFile` should be called to get the localPasswordFile (it returns the given value if it's not an encrypted secret, so it's ok to call each time). We can then just call `ValidatingFileReader.contents(p, passwordFile)` with the new value for localPasswordFile to get the resolvedPassword

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidator.java
```diff
@@ -107,9 +110,9 @@ public void validate(ConfigProblemSetBuilder p, DockerRegistryAccount n) {
           .accountName(n.getName())
           .address(n.getAddress())
           .email(n.getEmail())
-          .password(n.getPassword())
+          .password(secretSessionManager.decrypt(n.getPassword()))
           .passwordCommand(n.getPasswordCommand())
-          .passwordFile(n.getPasswordFile())
+          .passwordFile(secretSessionManager.decryptFile(n.getPasswordFile()))
```
link108: this can be the localPasswordFile referenced in an early comment.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/helm/HelmEditArtifactAccountCommand.java
```diff
@@ -39,14 +39,12 @@
       description = "Helm chart repository basic auth username"
   )
   private String username;
-  @Secret
```
lwander: Do we need to remove these? I'm afraid this will break users work flows automating the CLI

KathrynLewis: Those were initially put there in error, when I was still getting to know halyard. The annotation is supposed to be just on the config classes. So in this case, the @Secret is in HelmArtifactAccount. But we don't have to remove them if it will break things.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/secrets/SecretSessionManager.java
```diff
@@ -69,6 +71,14 @@ public String decrypt(String filePathOrEncryptedString) {
     return secretManager.decrypt(filePathOrEncryptedString);
   }
 
+  public String validatingFileDecrypt(ConfigProblemSetBuilder ps, String filePath) {
+    String contents = secretManager.decrypt(filePath);
```
lwander: This reads a little funny to me -- why are we checking to see if the output of decrypting the contents of a filepath matches the filepath?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/artifacts/helm/HelmEditArtifactAccountCommand.java
lwander: sgtm

PR #1167, opened 2019-01-17
fix(deploy): Handle null ServiceSettings

abergmeier:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
```diff
@@ -150,6 +150,9 @@ default String getVolumeYaml(ConfigSource configSource) {
   default String getResourceYaml(AccountDeploymentDetails<KubernetesAccount> details,
       GenerateService.ResolvedConfiguration resolvedConfiguration) {
     ServiceSettings settings = resolvedConfiguration.getServiceSettings(getService());
+    if (settings == null) {
+      throw new NullPointerException("Could not resolve ServiceSettings for " + getService());
```
ethanfrogers: What's the case where `settings` would be null?

abergmeier: Mostly tests - this reliance on non-null makes testing a PITA currently.

ethanfrogers: Gotcha. I wonder if throwing an exception here is actually what we want. How does it react if you do a `hal deploy apply` without anything in your `service-settings` directory?

abergmeier: The exception here is mostly just to detect the problem earlier. Without that check, you currently get a `NullPointerException` later in anyway.

> How does it react if you do a hal deploy apply without anything in your service-settings directory

Seems like you just volunteered to write another test 🥇

robzienert: I'm not sure this is the correct resolution to the problem. Why throw an NPE if the code later will throw an NPE without the presence of this check? Seems like we should be fixing the code itself so it's safer and handles exceptional behavior better.

On a professionalism note, I would say this sort of response (passing off improvement to someone taking the time to review your contribution) is not something we want to encourage within the community. There's a lot of things in Spinnaker to improve and the only way that will happen is if all contributors make the effort to cleanup technical debt themselves. If this comment was a joke, then fair play, an emoji helping lead us to that conclusion would help in the future. 😄

abergmeier: > Why throw an NPE if the code later will throw an NPE without the presence of this check?

So what this change does is change the precondition (there must be a ServiceSetting present) from implicit to explicit. Question is whether semantically the code was intended this way or not. I am new to Spinnaker and can not answer that.

> Seems like we should be fixing the code itself so it's safer and handles exceptional behavior better.

As a Java non-expert, I take any guidance I can get there. Since Java does not have the concept of non-null references, null pointers might not be a generally applicable to be handled (AFAIK there are ways like asserts, NotNull, Option, ... to do that but Java programmers seem to shy away from these).

PR #1176, opened 2019-01-23
fix(kayenta/setorder): Make Kayenta canary account supported types ordered.

stewchen:
This ensures that the generated kayenta.yaml keeps the same order for
supported types.


duftler:
Looks like the travis job failed to build.

PR #1183, opened 2019-01-24
feat(rosco/trim): Write only ImageProviders in rosco profile

stewchen:
This removes the unnecessary config from non-ImageProviders from the
config, also prevents unrelated config changes from bouncing rosco on hal deploy update.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/RoscoProfileFactory.java
```diff
@@ -101,8 +105,20 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
 
     augmentProvidersBaseImages(providers, otherProviders);
 
+    Map imageProviders = new TreeMap();
+
+    NodeIterator iterator = providers.getChildren();
+    Provider child = (Provider) iterator.getNext();
+    while (child != null) {
+      if (child instanceof HasImageProvider) {
```
ttomsu: Please also remove any image provider that is disabled

PR #1196, opened 2019-01-30
chore(*): Pin spring-boot-starter-test to springBoot version

dibyom:


build.gradle
```diff
@@ -57,7 +57,7 @@ subprojects { project ->
   dependencies {
     spinnaker.group('test')
     testCompile spinnaker.dependency('groovy')
-    testCompile 'org.springframework.boot:spring-boot-starter-test'
+    testCompile("org.springframework.boot:spring-boot-starter-test:${spinnaker.version('springBoot')}")
```
ezimanyi: I think this could also be `testCompile spinnaker.dependency('bootTest')`

dibyom: That looks much better! I'll make the change

dibyom:
@spinnakerbot cherry-pick 1.12

PR #1199, opened 2019-01-31
Revert "fix(rosco/profile): Set required files from enabled image pro…

stewchen:
…viders only. (#1198)"

This reverts commit bdfb2830d5d2bfc48c2e453ddb9f7ee9279a457c.


ezimanyi:
I think spinnakerbot will complain about the commit name...

PR #1205, opened 2019-02-03
fix(kubernetes): move kube specific setting to kubernetes config

paulczar:
When updating the docs for #1171 I realized that there's a kubernetes specific config, so this moves the `useExecHealthCheck` setting to `kubernetes.useExecHealthCheck`.

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>


paulczar:
@lwander

PR #1208, opened 2019-02-07
feat(docker): give a better error message during auth issues

kevinawoo:
Produces an error message like

and in the daemon logs


kevinawoo:
@spinnaker/reviewers :pray:

duftler:
Nice.

PR #1219, opened 2019-02-19
feat(deployments): add command for component sizing on kubernetes distributed deployments

costimuraru:
fixes https://github.com/spinnaker/spinnaker/issues/3981

```
spinnaker@halyard:$ hal config deploy component-sizing orca edit --replicas 3
+ Get current deployment
  Success
+ Get the deployment environment
  Success
+ Edit the deployment environment
  Success
+ Successfully edited component sizing for service orca.

spinnaker@halyard:$ hal config deploy component-sizing orca
+ Get current deployment
  Success
+ Get the deployment environment
  Success
+ Successfully got component sizing for service orca.
{replicas=3}
```

`$ cat ~/.hal/config`
```yaml
currentDeployment: default
deploymentConfigurations:
  deploymentEnvironment:
    size: SMALL
    type: Distributed
    ....
    customSizing:
      spin-orca:
        replicas: 3
```

```sh
$HAL_COMMAND config deploy component-sizing clouddriver edit --pod-requests-cpu 250m --pod-requests-memory 512Mi

$HAL_COMMAND config deploy component-sizing clouddriver edit --pod-limits-cpu 1 --pod-limits-memory 1Gi
```

```yaml
    customSizing:
      spin-clouddriver:
        replicas: 2
        requests:
          cpu: 250m
          memory: 512Mi
        limits:
          cpu: 1
          memory: 1Gi
```

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
@@ -25,8 +25,10 @@ import com.netflix.spinnaker.halyard.cli.command.v1.config.AbstractConfigCommand
 import com.netflix.spinnaker.halyard.cli.services.v1.Daemon;
 import com.netflix.spinnaker.halyard.cli.services.v1.OperationHandler;
 import com.netflix.spinnaker.halyard.cli.ui.v1.AnsiUi;
+import com.netflix.spinnaker.halyard.config.model.v1.ha.HaServices;
 import com.netflix.spinnaker.halyard.config.model.v1.node.CustomSizing;
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentEnvironment;
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.SpinnakerService;
 import lombok.AccessLevel;
 import lombok.Getter;
 
@@ -34,83 +36,59 @@ import java.util.HashMap;
 import java.util.Map;
 
 @Parameters(separators = "=")
-public class ComponentSizingEditCommand extends AbstractConfigCommand {
-    @Getter(AccessLevel.PROTECTED)
-    private Map<String, NestableCommand> subcommands = new HashMap<>();
-
-    @Getter(AccessLevel.PUBLIC)
-    private String commandName = "edit";
-
-    @Getter(AccessLevel.PUBLIC)
-    private String serviceName;
+public class ComponentSizingEditCommand extends AbstractComponentSizingUpdateCommand {
 
     @Parameter(
             names = "--replicas",
             description = "Set the number of replicas (pods) to be created for this service."
     )
-    private Integer replicas;
+    private Integer replicas = 1;
 
     @Parameter(
             names = "--pod-requests-cpu",
-            description = "Sets the cpu request for the container running the spinnaker service, as well as any sidecar containers (e.g. the monitoring daemon). Example: 250m."
+            description = "Sets the cpu request for the container running the spinnaker service, as well as any " +
+                    "sidecar containers (e.g. the monitoring daemon). Example: 250m."
     )
     private String requestsCpu;
 
     @Parameter(
             names = "--pod-requests-memory",
-            description = "Sets the memory request for the container running the spinnaker service, as well as any sidecar containers (e.g. the monitoring daemon). Example: 512Mi."
+            description = "Sets the memory request for the container running the spinnaker service, as well as any " +
+                    "sidecar containers (e.g. the monitoring daemon). Example: 512Mi."
     )
     private String requestsMemory;
 
     @Parameter(
             names = "--pod-limits-cpu",
-            description = "Sets the cpu limit for the container running the spinnaker service, as well as any sidecar containers (e.g. the monitoring daemon). Example: 1."
+            description = "Sets the cpu limit for the container running the spinnaker service, as well as any " +
+                    "sidecar containers (e.g. the monitoring daemon). Example: 1."
     )
     private String limitsCpu;
 
     @Parameter(
             names = "--pod-limits-memory",
-            description = "Sets the memory limit for the container running the spinnaker service, as well as any sidecar containers (e.g. the monitoring daemon). Example: 1Gi."
+            description = "Sets the memory limit for the container running the spinnaker service, as well as any " +
+                    "sidecar containers (e.g. the monitoring daemon). Example: 1Gi."
     )
     private String limitsMemory;
 
-    public ComponentSizingEditCommand(String serviceName) {
-        this.serviceName = serviceName;
+    public ComponentSizingEditCommand(SpinnakerService.Type spinnakerService) {
+        super(spinnakerService, "edit");
     }
 
     @Override
     protected String getShortDescription() {
-        return "Edit the component sizing for service " + getServiceName();
+        return "Edit the component sizing for service " + spinnakerService.getCanonicalName() +
+                ", such as the number of replicas and the resources limits.";
     }
 
     @Override
-    protected void executeThis() {
-        String currentDeployment = getCurrentDeployment();
-        String serviceName = getServiceName();
-        DeploymentEnvironment deploymentEnvironment = new OperationHandler<DeploymentEnvironment>()
-                .setFailureMesssage("Failed to get component sizing for service " + serviceName + ".")
-                .setOperation(Daemon.getDeploymentEnvironment(currentDeployment, false))
-                .get();
-
-        CustomSizing customSizing = deploymentEnvironment.getCustomSizing();
-        int originalHash = customSizing.hashCode();
-
-        customSizing = edit(customSizing);
-
-        if (originalHash == customSizing.hashCode()) {
-            AnsiUi.failure("No changes supplied.");
-            return;
-        }
-
-        new OperationHandler<Void>()
-                .setFailureMesssage("Failed to edit component sizing for " + serviceName + ".")
-                .setSuccessMessage("Successfully edited component sizing for service " + serviceName + ".")
-                .setOperation(Daemon.setDeploymentEnvironment(currentDeployment, !noValidate, deploymentEnvironment))
-                .get();
+    protected CustomSizing update(CustomSizing customSizing) {
+        return edit(customSizing);
     }
 
     private CustomSizing edit(CustomSizing customSizing) {
-        Map serviceSizing = customSizing.computeIfAbsent(spinService(), (k) -> new HashMap<>());
+        Map serviceSizing = customSizing.computeIfAbsent(spinnakerService.getServiceName(), (k) -> new HashMap<>());
         edit(serviceSizing);
         return customSizing;
     }
@@ -131,9 +109,5 @@ public class ComponentSizingEditCommand extends AbstractConfigCommand {
         if (value != null) {
             map.put(key, value);
         }
-    }
-
-    private String spinService() {
-        return "spin-" + serviceName;
     }
 }
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1.config.deploy.sizing;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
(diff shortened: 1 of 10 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
```diff
@@ -0,0 +1,139 @@
...
+    private String commandName = "edit";
+
+    @Getter(AccessLevel.PUBLIC)
+    private String serviceName;
+
+    @Parameter(
+            names = "--replicas",
+            description = "Set the number of replicas (pods) to be created for this service."
+    )
+    private Integer replicas;
```
kevinawoo: does this default to 1?

bpowell: In Java the default value for `Integer` is 0

kevinawoo: Oh I meant if it's not set will halyard still deploy 1 pod?

bpowell: I think it might do 0 pods, because it will be set to the default value of 0 and then `putIfNotNull` will be ran and the value is not null, so 0 will be placed for replicas. But I will test this out to see

kevinawoo:
I don't think this accounts for different HA sizing shown here: https://www.spinnaker.io/reference/halyard/high-availability/

Did you want to add it to this PR or do a different one?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
kevinawoo: I think we're good, if this field isn't provided in the normal deployment, it'll default to 1.

Defaultly, it doesn't write it to `~/.hal/config`

costimuraru: The Java default for `int` is 0, but the default for `Integer` is null [1]. Given we're using `Integer`, we should be fine because of the `putIfNotNull` call. For more clarity though, I could just put a default `1` here. What do you think?


https://stackoverflow.com/questions/2437603/why-does-using-a-default-valued-java-integer-result-in-a-nullpointerexception

costimuraru:
@kevinawoo, are you saying that if HA is enabled, giving birth to more services (eg. clouddriver-caching, clouddriver-rw, echo-scheduler, echo-worker), we should be able to control the no of replicas and requests/limits for them as well?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
bpowell: @costimuraru I think it's better to be explicit then implicit on this. I would default it to 1

kevinawoo:
> @kevinawoo, are you saying that if HA is enabled, giving birth to more services (eg. clouddriver-caching, clouddriver-rw, echo-scheduler, echo-worker), we should be able to control the no of replicas and requests/limits for them as well?

Yep, we should be able to control those too from the CLI.

costimuraru:
Good point, @kevinawoo.
I made the change to also include the HA components. It looks like it's working!

```sh
$HAL_COMMAND config deploy ha clouddriver enable
$HAL_COMMAND config deploy component-sizing clouddriver-caching edit --replicas 3
$HAL_COMMAND config deploy ha echo enable
$HAL_COMMAND config deploy component-sizing echo-scheduler edit --replicas 4
$HAL_COMMAND config deploy component-sizing echo-worker edit --replicas 5
```

```
spin-clouddriver-caching-6574dd98cd-9pbt4   0/2     Pending             0          7s
spin-clouddriver-caching-6574dd98cd-j8p9l   0/2     ContainerCreating   0          7s
spin-clouddriver-caching-6574dd98cd-sppdd   0/2     ContainerCreating   0          7s
spin-clouddriver-ro-765bc758cb-266sm        0/2     ContainerCreating   0          7s
spin-clouddriver-ro-deck-65cfc68dcb-ctdlb   0/2     Pending             0          6s
spin-clouddriver-rw-654496b884-sdnwc        0/2     ContainerCreating   0          7s
spin-echo-scheduler-6949f76997-brk2s        0/2     ContainerCreating   0          7s
spin-echo-scheduler-6949f76997-fw4g9        0/2     ContainerCreating   0          7s
spin-echo-scheduler-6949f76997-qww4t        0/2     ContainerCreating   0          7s
spin-echo-scheduler-6949f76997-shz4q        0/2     ContainerCreating   0          7s
spin-echo-worker-578fd5c86b-4z8vd           0/2     ContainerCreating   0          7s
spin-echo-worker-578fd5c86b-jkxms           0/2     ContainerCreating   0          7s
spin-echo-worker-578fd5c86b-kpmdr           0/2     ContainerCreating   0          7s
spin-echo-worker-578fd5c86b-qqp92           0/2     ContainerCreating   0          7s
spin-echo-worker-578fd5c86b-x7wbk           0/2     ContainerCreating   0          7s
```

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
costimuraru: Made the change to default it to 1, @bpowell

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
```diff
@@ -0,0 +1,139 @@
...
+        putIfNotNull(requests, "memory", requestsMemory);
+    }
+
+    private void putIfNotNull(Map map, String key, Object value) {
+        if (value != null) {
+            map.put(key, value);
+        }
+    }
+
+    private String spinService() {
```
milesmatthias: At some point this should be a utility method somewhere, as it's done elsewhere too: https://github.com/spinnaker/halyard/blob/master/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/SpinnakerService.java#L191

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/NamedComponentSizingCommand.java
```diff
@@ -0,0 +1,71 @@
...
+  @Override
+  protected void executeThis() {
+    String currentDeployment = getCurrentDeployment();
+    String serviceName = getServiceName();
+    new OperationHandler<Map>()
+        .setFailureMesssage("Failed to get component sizing for service " + serviceName + ".")
+        .setSuccessMessage("Successfully got component sizing for service " + serviceName + ".")
+        .setFormat(STRING)
+        .setUserFormatted(true)
+        .setOperation(() -> Daemon.getDeploymentEnvironment(currentDeployment, !noValidate).get().getCustomSizing().get("spin-" + serviceName))
```
milesmatthias: Could use a utility method to get pod name "spin-" + serviceName

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
```diff
@@ -0,0 +1,139 @@
...
+
+        customSizing = edit(customSizing);
+
+        if (originalHash == customSizing.hashCode()) {
+            AnsiUi.failure("No changes supplied.");
+            return;
+        }
+
+        new OperationHandler<Void>()
+                .setFailureMesssage("Failed to edit component sizing for " + serviceName + ".")
```
milesmatthias: There are 4 lines of output messages (failure/success x get/edit) you could template with a private method.

milesmatthias:
Just added a few thoughts/questions if it's helpful.

Nice job supporting the HA services too! I assume it works with or without the HA services, and if you transition from customSizing with non-HA services to HA services, everything still works?

i.e., this workflow doesn't produce errors:

1. Without HA, use this command to scale `clouddriver`.
2. Enable HA, which splits `clouddriver` into multiple services.
3. Scale one of the new `clouddriver` component services with this command.

I guess you can test this now though, without this command, because really Halyard should ignore customSizing for `clouddriver` if HA is enabled.

costimuraru:
Yes, right now you can enable HA and set the number of replicas for
clouddriver to X, in which case this will be ignored (because clouddriver
is split into the sub-services). This might be a bit confusing though. Do
you think we should update the code to throw an error if you try to update
clouddriver/echo when HA enabled, and show a message to instruct the user
to update clouddriver-rw, clouddriver-rw etc individually?

Same goes with updating clouddriver-rw when HA disabled.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
```diff
@@ -0,0 +1,139 @@
...
+        Map limits = (Map) serviceSizing.computeIfAbsent("limits", (k) -> new HashMap<>());
+        putIfNotNull(limits, "cpu", limitsCpu);
+        putIfNotNull(limits, "memory", limitsMemory);
+
+        Map requests = (Map) serviceSizing.computeIfAbsent("requests", (k) -> new HashMap<>());
+        putIfNotNull(requests, "cpu", requestsCpu);
+        putIfNotNull(requests, "memory", requestsMemory);
+    }
+
+    private void putIfNotNull(Map map, String key, Object value) {
```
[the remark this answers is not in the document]
costimuraru: computeIfAbsent refers to the key being absent. putIfNotNull checks that the value is null, so they are quite distinct. 

If we use computeIfAbsent and can imagine a scenario where you want to increase the number of replicas from 2 to 3:
computeIfAbsent(replicas, 2)
computeIfAbsent(replicas, 3)

The second call is ignored, because the replicas key exists and the value will continue to be 2, which is not what we want.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
costimuraru: Good point, I think I will actually use that enum in this class and get rid of this method together.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
costimuraru: Not sure I'm understanding this one. Could you please help me with more details?

kevinawoo: Refactor out `Failed to ... ` and `Success to... ` to a common method, then call that in each 4 places. 

However I rather be it be verbose for readability sake. But I'm not strong either way.

costimuraru:
Based on your feedback I restructured the code a little bit:
- to catch custom sizings mismatches for services with high availability enabled, I added validators for ha echo and ha clouddriver
- in order to clear the custom sizings for a given service (useful if you move from HA to non-HA and vice-versa) I've added a `component-size delete` command


```sh
$ hal config deploy ha echo enable
+ Edit the echo high availability service configuration
  Success

$ hal config deploy component-sizing echo edit --replicas 4
Problems in default.deploymentEnvironment:
- WARNING High Availability (HA) is enabled for echo, but found
  custom sizing for the main service (this setting will be ignored). With HA
  enabled, the service is split into multiple sub-services (echo-scheduler,
  echo-scheduler). You need to update the component sizing for each sub-service,
  individually.

$ hal config deploy component-sizing echo delete
+ Successfully deleted component sizings for service echo.
```
and 
```
$ hal config deploy ha echo disable
+ Successfully disabled echo

hal config deploy component-sizing echo-worker edit --replicas 4
Problems in default.deploymentEnvironment:
- WARNING Discovered custom sizing for HA echo subcomponent, but
  High Availability (HA) is not enabled. Please enable HA or edit the echo main
  service directly.
```

Let me know what you think.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/NamedComponentSizingCommand.java
costimuraru: This has been fixed. We now use `getCustomSizing().get(spinnakerService.getServiceName())`

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
costimuraru: Agree, I also feel we'd lose on readability.

milesmatthias:
Yes, epic! 👏

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/deploy/sizing/ComponentSizingEditCommand.java
milesmatthias: Doesn't matter to me either way, just thought I'd call it out. 👍

PR #1222, opened 2019-02-21
fix(secrets): fix gce credential loading

KathrynLewis:
The original change was adding support for storing credentials in an encrypted s3 bucket instead of a plain text file. This is part of the secrets management project here: https://github.com/spinnaker/spinnaker/issues/3649

The problem:
We're seeing an NPE from `GoogleAccount.getNamedAccountCredentials` because we're trying to autowire the `SecretSessionManager` in a class that's not being managed by Spring. The NPE occurs when we try to call `secretSessionManager.validatingFileDecrypt(p, getJsonPath())`.

The fix:
Removing the autowired `SecretSessionManager` in both `GoogleAccount` and `GoogleCanaryAccount` and instead passing it in from the upstream callers where supported. Namely these upstream validators: `GoogleAccountValidator`, `GoogleProviderValidator`,  `GoogleCanaryAccountValidator`.

The flow that isn't supported is the call from `GoogleProviderUtils`. In this case, we're passing in null and the credentials file will be read in the standard way, using `ValidatingFileReader.contents(p, getJsonPath())`.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/canary/google/GoogleCanaryAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/canary/google/GoogleCanaryAccount.java
@@ -30,7 +30,6 @@ import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.StringUtils;
-import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.SortedSet;
 import java.util.TreeSet;
@@ -47,11 +46,8 @@ public class GoogleCanaryAccount extends AbstractCanaryAccount implements Clonea
   private String rootFolder = "kayenta";
   private SortedSet<AbstractCanaryServiceIntegration.SupportedTypes> supportedTypes = new TreeSet<>();
 
-  @Autowired
-  private SecretSessionManager secretSessionManager;
-
   @JsonIgnore
-  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, ConfigProblemSetBuilder p) {
+  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, SecretSessionManager secretSessionManager, ConfigProblemSetBuilder p) {
     String jsonKey = null;
     if (!StringUtils.isEmpty(getJsonPath())) {
       jsonKey = secretSessionManager.validatingFileDecrypt(p, getJsonPath());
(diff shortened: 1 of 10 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/canary/google/GoogleCanaryAccount.java
```diff
@@ -47,11 +46,8 @@
   private String rootFolder = "kayenta";
   private SortedSet<AbstractCanaryServiceIntegration.SupportedTypes> supportedTypes = new TreeSet<>();
 
-  @Autowired
-  private SecretSessionManager secretSessionManager;
-
   @JsonIgnore
-  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, ConfigProblemSetBuilder p) {
+  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, SecretSessionManager secretSessionManager, ConfigProblemSetBuilder p) {
```
jtk54: Should we null guard `secretSessionManager` as in `GoogleAccount`?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/google/GoogleAccount.java
```diff
@@ -47,14 +47,15 @@
...
-  @Autowired
-  private SecretSessionManager secretSessionManager;
-
   @JsonIgnore
-  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, ConfigProblemSetBuilder p) {
+  public GoogleNamedAccountCredentials getNamedAccountCredentials(String version, SecretSessionManager secretSessionManager, ConfigProblemSetBuilder p) {
     String jsonKey = null;
     if (!StringUtils.isEmpty(getJsonPath())) {
-      jsonKey = secretSessionManager.validatingFileDecrypt(p, getJsonPath());
+      if (secretSessionManager != null) {
```
jtk54: In what cases will this case occur?

jtk54:
Can we summarize what's broken and why in the PR/commit message to add more context to the change?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/canary/google/GoogleCanaryAccount.java
KathrynLewis: The only place this method is getting called is in `GoogleCanaryAccountValidator` which has a (correctly) autowired `SecretSessionManager`. But if we want to be safe, I can certainly add that in.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/google/GoogleAccount.java
KathrynLewis: This method is called from two places:
* `GoogleAccountValidator` - this class has a functioning `SecretSessionManager` so it won't be null
* `GoogleProviderUtils` - doesn't have `SecretSessionManager` and secret decryption isn't supported in this flow, so we're passing in null. The credentials will be read in the way they were before, using `ValidatingFileReader.contents(p, getJsonPath())`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/canary/google/GoogleCanaryAccount.java
jtk54: SG as is. Probably better to fail explicitly than default to subtle backup behavior.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/providers/google/GoogleAccount.java
jtk54: Sounds good.

PR #1225, opened 2019-02-25
feat(core): Use gradle to generate docs

ezimanyi:
The Halyard docs are currently generated via a makefile. Move this to gradle so that it can be incorporated into the build process rather than a separate shell command.

In particular, the check task should fail if the docs do not generate properly or if the docs committed to git do not match what are generated from source. This will cause our Travis build to fail and block merging when a PR either forgets to update the docs or breaks doc generation.


ezimanyi:
Travis will currently fail because the docs are currently in a bad state in the repo (which was my motivation for this PR in the first place); it will pass again once I rebase on top of a good state for the docs.

ezimanyi:
As a separate PR, will look into removing this from the makefile entirely, but we have some scripts that currently use the makefile that need to be changed first.

PR #1227, opened 2019-02-25
refactor(docs): Refactor descriptions to use static type checking

ezimanyi:
The docs generation currently has description, shortDescription, and longDescription, and commands must implement a subset of these that is checked at runtime.  Simplify this by just using shortDescription (required) and longDescription (optional), and enforcing this using static type checking.


ezimanyi:
This should provide even quicker feedback to prevent breaking docs generation; the important point is that the generated docs are identical before and after this PR (and as of #1225 Travis would fail if this had changed the docs, and I didn't check in the change).

PR #1229, opened 2019-02-26
fix debian noninteractive install

milesmatthias:
I ran into this installing Halyard from a debian 9 dockerfile.

All of the other `apt install` commands run non-interactively other than installing Java. Maybe adding ca-certificates is something you want someone to acknowledge, but if that's the case, then when someone does

```
InstallHalyard.sh -y
```

*everything* should be non-interactive.

I also had to have my dockerfile manually add `stretch-backports` to `/etc/apt/sources.list.d/stretch-backports.list`

Edit: fixes https://github.com/spinnaker/spinnaker/issues/4045


milesmatthias:
Got rid of the commit that adds the repo to the sources list per @ezimanyi.

PR #1230, opened 2019-03-01
fix(clouddriver/profile): Enable custom UDF files

ncknt:
This is to enable UDF files to be staged and deployed. Fixes https://github.com/spinnaker/spinnaker/issues/4058


ncknt:
@ezimanyi @duftler Could you PTAL?

PR #1235, opened 2019-03-05
feat(appengine): add flag to configure caching agent interval

sbwsg:
Allows a user to configure Spinnaker to hit the AppEngine Admin API less often, thereby reducing the risk of hitting quotas.


ezimanyi:
Whoops, I think I accidentally hit "merge" on this....meant to only approve.  But approved!

PR #1236, opened 2019-03-06
feat(kubernetes): add live-manifest-calls flag to cli

maggieneterval:
There's been some discussion in Slack that the `liveManifestCalls` feature works wonderfully, but should not have to be manually added to one's halconfig.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesEditAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesEditAccountCommand.java
@@ -227,6 +227,13 @@ public class KubernetesEditAccountCommand extends AbstractEditAccountCommand<Kub
   )
   public Boolean checkPermissionsOnStartup;
 
+  @Parameter(
+      names = "--live-manifest-calls",
+      arity = 1,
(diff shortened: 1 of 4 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesEditAccountCommand.java
```diff
@@ -296,6 +303,7 @@ protected Account editAccount(KubernetesAccount account) {
     account.setSkin(isSet(skin) ? skin : account.getSkin());
     account.setOnlySpinnakerManaged(isSet(onlySpinnakerManaged) ? onlySpinnakerManaged : account.getOnlySpinnakerManaged());
     account.setCheckPermissionsOnStartup(isSet(checkPermissionsOnStartup) ? checkPermissionsOnStartup : Boolean.TRUE);
+    account.setLiveManifestCalls(isSet(liveManifestCalls) ? liveManifestCalls : Boolean.FALSE);
```
ezimanyi: I think the false branch of this should be `account.getLiveManifestCalls()` (with a similar pre-existing bug existing in the above line).  Otherwise editing any other field will turn off `liveManifestCalls`.

maggieneterval: Oh good catch! Fixed the bug in the line above as well.

PR #1247, opened 2019-03-12
fix(secrets): fix saml secret decryption

KathrynLewis:
Normally, secret annotations are placed on secret fields in the config classes so that the `DecryptingObjectMapper` knows which fields to decrypt when serializing. In the case of `Saml`, the `GateProfileFactory` creates a new `SamlConfig` object instead of passing the `Saml` object itself to the `DecryptingObjectMapper`. Therefore, the `SamlConfig` class needs annotations on secret fields in order to decrypt when printing the service profiles.


KathrynLewis:
@dibyom PTAL?

PR #1248, opened 2019-03-13
feat(secrets): clear secret caches at end of each session

KathrynLewis:
We want to make sure secrets and more importantly the secret temp files are cleaned up at the end of each session. So we're doing that `DaemonTask` in the `cleanupResources` method.  In order to avoid circular dependencies, we're moving `SecretSessionManager` from halyard-config to halyard-core. To simplify the caching, we've moved `SecretSession` out of halyard and handling all caching in kork instead (see https://github.com/spinnaker/kork/pull/247).

(this won't build until spinnakerbot autobumps dependencies)

@dibyom PTAL?


brandonnelson3:
Added a commit bumping the spinnaker dependencies. Expecting Travis to pass this time.

PR #1249, opened 2019-03-14
feat(secrets): set minimum version for decryption in services

KathrynLewis:
When creating service profiles, check for the min version required for each service in order to be able to decrypt secrets. If unsupported in that version, decrypt secrets on the way out, otherwise leave them as encrypted strings.


KathrynLewis:
@dibyom PTAL?

PR #1251, opened 2019-03-19
chore(install): Add Arch Linux installation

coredump:
Not a lot of changes, since bash is bash anywhere, but this is a list:

 * Uses `who` instead of `who -m` that doesn't behave correctly on arch,
 I guess something different about how loginctl works, but couldn't
 find anything specific.
 * Uses pacman to install the needed java/ca packages instead of dpkg.
 * Removed the migration check, that is specific to past debian
 packages.
 * Uses a `sysusers` file to create system groups instead of using
 `groupadd` as suggested by the packaging guidelines.

Still mentions debian for the download location/archives since there's no 
generic named file to download.

 Tested on Arch and Manjaro, and worked fine. There may be some
 imcompatibility with other arch based distros but should cover most of
 the options around.


install/arch/InstallHalyard.sh
```diff
@@ -0,0 +1,343 @@
...
+        echo "ERROR: Unknown argument '$key'"
+        exit -1
+    esac
+  done
+}
+
+function get_user() {
+  local user
+
+  user=$(who | awk '{print $1;}')
```
ezimanyi: Won't moving from `who -m` to `who` mean we'll print out all of the users currently logged in and store them in `user`?  I wonder if this can just be replaced by `user=$(whoami)` (which should also work for Debian, reducing the differences between the install scripts).

coredump: whoami works, you are right. I will do some work on this this week.

coredump:
Modified the PR to make it a single script, as @ezimanyi suggested.

dansimau:
This breaks the docs at:
https://www.spinnaker.io/setup/install/halyard/

Filed https://github.com/spinnaker/spinnaker.github.io/issues/1320

ezimanyi:
Whoops, I didn't notice in the PR diff that the file was moved as well as changed.  I think we should leave this at the original path...while it would be technically more correct to have it in a path called `linux` rather than `debian` this change has the potential to break automation that people have set up and I'm not sure being a bit more correct is worth that breakage.

ezimanyi:
Moving the script back in #1277.

PR #1252, opened 2019-03-19
feat(provider/kubernetes): Pods security context in service-settings

german-muzquiz:
This is for being able to specify the `securityContext` under which spinnaker's pods will run. Example `service-settings/clouddriver.yml`:
```
kubernetes:
  securityContext:
    runAsUser: 99
    fsGroup: 99
```
Results in:
```
bash-4.4$ whoami
whoami: unknown uid 99
bash-4.4$
bash-4.4$ ls -la /opt/spinnaker/
total 0
drwxr-xr-x    3 root     root            20 Mar 19 17:21 .
drwxr-xr-x    1 root     root            23 Mar 19 17:21 ..
drwxrwsrwt    3 root     99             120 Mar 19 17:20 config
bash-4.4$
```


german-muzquiz:
cc @dibyom

PR #1257, opened 2019-03-25
feat(kubernetes): add initialDelaySeconds to livenessProbe

maggieneterval:
- Adds three-minute initial delay to livenessProbe to prevent repeated restarting of the container on startup

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v1/KubernetesV1DistributedService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v1/KubernetesV1DistributedService.java
@@ -420,6 +420,7 @@ public interface KubernetesV1DistributedService<T> extends DistributedService<T,
     livenessHandlerAction.setPort(port);
     livenessHandler.setTcpSocketAction(livenessHandlerAction);
     livenessProbe.setHandler(livenessHandler);
+    livenessProbe.setInitialDelaySeconds(180);
     container.setLivenessProbe(livenessProbe);
 
     applyCustomSize(container, deploymentEnvironment, name, description);
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
@@ -341,6 +341,7 @@ public interface KubernetesV2Service<T> extends HasServiceSettings<T> {
 
     TemplatedResource livenessProbe = new JinjaJarResource("/kubernetes/manifests/tcpSocketLivenessProbe.yml");
     livenessProbe.addBinding("port", settings.getPort());
+    livenessProbe.addBinding("initialDelaySeconds", 180);
 
     String lifecycle = "{}";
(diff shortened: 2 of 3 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v1/KubernetesV1DistributedService.java
```diff
@@ -420,6 +420,7 @@ default KubernetesContainerDescription buildContainer(String name, ServiceSettin
     livenessHandlerAction.setPort(port);
     livenessHandler.setTcpSocketAction(livenessHandlerAction);
     livenessProbe.setHandler(livenessHandler);
+    livenessProbe.setInitialDelaySeconds(180);
```
dibyom: Is there a way to make this configurable? Clouddriver might take a really long time to start up if it is caching a large number of accounts or k8s cluster

ewiseblatt: second making it configurable.

maggieneterval: Makes sense! Not very familiar with Halyard, but I'm guessing we'd configure this by adding a `livenessProbeInitialDelaySeconds` field to [`ServiceSettings`](https://github.com/spinnaker/halyard/blob/master/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/ServiceSettings.java#L45)?

ezimanyi: One concern is that if we think that people will run into this limit and will need to override the default value, they will only find out when `clouddriver` fails to start...and then we'll be asking them to change another parameter (which is not a great experience).  I'm wondering if we should think of some way that is more general (and doesn't require more user-overridable values)...not that something comes to mind right away.

maggieneterval: Great point. What if we set a very high default `initialDelaySeconds`?

maggieneterval: Just chatted offline with @ezimanyi and we are going to revert the commit that added the `livenessProbe` until we can land on a proper configuration. We'll get some feedback from the Spinnaker k8s SIG and circle back on this.

ewiseblatt: you could possibly go the other way for now if you wished something were in there.
Have it configurable with the default being disabled. Then you can turn it on with a value if you want it. That would satisfy short term demand if this was motivated by requests.

maggieneterval: That's a good option, thanks @ewiseblatt! Another option floated by @brandonnelson3 is to have a second server in each service's container and do a TCP check on that for the `livenessProbe` -- any thoughts on that? For now, I've reverted the original change until we land on the best solution.

PR #1258, opened 2019-03-25
Revert "feat(kubernetes): add livenessProbe"

maggieneterval:
Reverts spinnaker/halyard#1253

Adding a `livenessProbe` with no configured `initialDelaySeconds` causes services with long startup times to crash and restart indefinitely. However, we are hesitant to introduce another user-configured value that can only be discovered after an initial failed deployment experience, with no obvious default. Reverting this change until we land on the best way to configure a `livenessProbe` that will work for most users out of the box.


maggieneterval:
cc @davidxia -- so sorry that we ended up having to revert this for now! We'll try to land on a good default configuration soon and can add this back in.

davidxia:
No worries. Thanks for the heads up.

PR #1261, opened 2019-03-28
feat(gremlin): Allow Gremlin to get configured via Halyard

mattrjacobs:
The idea here is to allow `hal config feature edit` to accept `--gremlin` as an argument.

Then, that setting will appear as `features.gremlin` to `deck`, `orca`, and `gate`.


mattrjacobs:
I see this failure in the CI logs: ```* What went wrong:
Execution failed for task ':halyard-cli:checkDocs'.
> The docs for halyard-cli are out of date. Regenerate them by running the :halyard-cli:updateDocs gradle task.```

I don't think that has anything to do with my changes.  Any thoughts on how I can resolve?

PR #1268, opened 2019-04-03
fix(halyard): Ignore the user profile if it doesn't exist.

plumpy:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/CustomProfileFactory.java
```diff
@@ -17,14 +17,15 @@
 
 package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile;
 
+import com.google.common.io.Files;
```
ezimanyi: We tend to use apache commons over guava for I/O (and actually I don't think guava is pulled into most of our microservices) so it might make sense to stick with apache commons.  (I think that `guava` is actually only transitively included in this module.)

plumpy: Yeah okay, fair enough. (But also: sad.)

PR #1269, opened 2019-04-03
Master mac note

milesmatthias:
To match https://www.spinnaker.io/setup/install/halyard/


milesmatthias:
blargh accidentally included too many commits.

PR #1254, opened 2019-03-21
update(kubectl): Update dockerfile to use latest kubectl available

ricklagerweij:
Update kubectl to latest current version available.

Issue #699, opened 2017-09-18
hal deploy connect in docker binds to localhost


neilhwatson:
Using the docker Halyard image ```hal deploy connect``` will forward remote ports to localhost in the container. These ports are not reachable from the container's host.

lwander:
Blocked on https://github.com/kubernetes/kubernetes/issues/43962

neilhwatson:
I don't think this is k8s related. I'm speaking of the shipped docker image for the halyard client. Another option would be to support a native client on multiple Linux distros (e.g. Debian and RH).

lwander:
The connect command for k8s relies on kubectl, which is what that bug depends on.

spinnaker-release:
Migrated to https://github.com/spinnaker/spinnaker/issues/2409

PR #1270, opened 2019-04-09
chore: update kubectl to version 1.14.1

jfillo:


jfillo:
@ezimanyi this is the PR to replace https://github.com/spinnaker/halyard/pull/1254

jemc:
When will this update be reflected in `gcr.io/spinnaker-marketplace/halyard:stable`?
I'm still seeing the old version of `kubectl` in that image, even though this PR has the `target-release/1.19` label, and GCR shows the latest stable image version as being `1.20`.
Screenshot from 2019-06-20 08-23-41

Note that this fix is needed as a workaround for #699.

ezimanyi:
@jemc: This was actually partially rolled back in #1278, which pulled us to 1.12.7 (with the reasons for that listed on that PR).  Given that it's 2 months later now and kubernetes 1.15 is out, I think it would be reasonable to upgrade to kubectl 1.13.x at this point.

PR #1272, opened 2019-04-12
fix(halyard): Use "kubectl replace --force" for Kubernetes Secret creation

legal90:
This PR fixes the issue of deploying Spinnaker in Kubernetes when the halconfig is quite big (~35 accounts and more):
https://community.spinnaker.io/t/halyard-with-v2-metadata-annotations-too-long/894

Unlikely, `kubectl apply`, the `kubectl replace --force` command does not preserves
current data to the annotation. That allows to work around the issue when the halyard config is quite big and could not fit into the annotation.


maggieneterval:
@ezimanyi seems like an okay workaround to me, as long as there's no `apply`-specific functionality we're relying on for secret creation?

ezimanyi:
I dug around a bit, and didn't see any `apply`-specific things we're relying on for secrets.

PR #1273, opened 2019-04-12
feat(kube/v2) - add ability to specify Node, Pod, and Anti Affinities

perek:
Background
We need to allow more control over where the Halyard kubernetes provider schedules the services pods. Adding Affinity allows us to tell kubernetes to not schedule certain pods together, and to schedule certain pods on certain nodes.

Change
Add affinity support to Halyard

Test
Added extensive unit tests to confirm the format of the affinity


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/AffinityConfig.java
```diff
@@ -0,0 +1,101 @@
...
+package com.netflix.spinnaker.halyard.config.model.v1.node;
+
+import lombok.Data;
+
+import java.util.List;
+import java.util.Map;
+
+@Data
+public class AffinityConfig {
+    public PodAffinity  podAntiAffinity;
```
ezimanyi: We generally just leave a single space between the type and variable name rather than align things like this.  (I realize style things like this aren't really well documented anywhere.)

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/AffinityConfig.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.node;
+
+import lombok.Data;
+
+import java.util.List;
+import java.util.Map;
+
+@Data
```
ezimanyi: This class (and nested classes in this file) is both `@Data` and has all fields `public` (or in some cases package private).

Given that `@Data` will generate all the accessors for us, it probably makes sense to just make the fields private.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/AffinityConfig.java
```diff
@@ -0,0 +1,101 @@
...
+        NodeSelectorTerm preference;
+    }
+
+    @Data
+    public static class NodeSelectorRequirement {
+        String   key;
+        Operator operator;
+        String[] values;
+
+        enum Operator
```
ezimanyi: Style nit: opening brace should be on same line as `Operator`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/AffinityConfig.java
```diff
@@ -0,0 +1,101 @@
...
+        List<LabelSelectorRequirement> matchExpressions;
+    }
+
+    @Data
+    public static class LabelSelectorRequirement {
+        String   key;
+        Operator operator;
+        String[] values;
+
+        enum Operator
```
ezimanyi: Same style nit as above: opening brace should be on same line as `Operator`

PR #1274, opened 2019-04-13
fix(secrets): Always decrypt monitoring secrets

ncknt:
Monitoring secrets (Datadog, Stackdriver) won't have Kork decryption and need to be mounted decrypted. This is done by adding a `alwaysDecrypt` on `@Secret` and `@SecretFile` and simplifying the object mapper annotation check.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Secret.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Secret.java
@@ -11,5 +11,9 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 public @interface Secret {
+    /**
+     * @return true to force a secret to be decrypted even when the service it is attached to supports decryption.
+     * This is true for properties that cannot be decrypted at runtime (e.g. spinnaker-monitoring settings).
+     */
     boolean alwaysDecrypt() default false;
 }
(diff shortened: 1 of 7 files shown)


ncknt:
@dibyom could you PTAL?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Secret.java
```diff
@@ -11,4 +11,5 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 public @interface Secret {
+    boolean alwaysDecrypt() default false;
```
dibyom: Nit. Could we add a comment description for `alwaysDecrypt` and when it should be set to true?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/SecretFile.java
```diff
@@ -12,4 +12,5 @@
 @Target(ElementType.FIELD)
 public @interface SecretFile {
   String prefix() default "";
+  boolean alwaysDecrypt() default false;
```
dibyom: Same as above

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Secret.java
ncknt: Good point. Added.

dibyom:

PR #1276, opened 2019-04-15
fix(secrets): fix file prefix for saml keystore

KathrynLewis:
The saml keystore requires a "file:" prefix in the `gate.yml`, which is currently added in the `secretFileSerializer`.  The bug: when secret decryption is supported in the services AND the keystore file is not an encrypted secret, the code skips the custom secretFileSerializer, which means "file:" doesn't get prepended to the path.  This fixes the logic so that an encrypted secret that should stay encrypted doesn't get the prefix, but all other cases do.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
@@ -35,8 +35,6 @@ import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.Profile;
 import org.apache.commons.lang.RandomStringUtils;
 
 import java.io.IOException;
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Field;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/config/v1/secrets/DecryptingObjectMapper.java
```diff
@@ -107,21 +106,21 @@ public void serialize(Object value, JsonGenerator gen, SerializerProvider provid
...
-                    if (EncryptedSecret.isEncryptedSecret(sValue)) {
+                    if (!EncryptedSecret.isEncryptedSecret(sValue)) {
+                        sValue = annotation.prefix() + sValue;
+                    } else if (shouldDecrypt) {
                         // Decrypt the content of the file and store on the profile under a random
                         // generated file name
                         String name = newRandomFilePath(beanPropertyWriter.getName());
                         byte[] bytes = secretSessionManager.decryptAsBytes(sValue);
                         profile.getDecryptedFiles().put(name, bytes);
                         sValue = getCompleteFilePath(name);
```
dibyom: Can this simply be `sValue = annotation.prefix() + getCompleteFilePath(name);` ?

KathrynLewis: yes, will change!

dibyom:
Fixes https://github.com/spinnaker/spinnaker/issues/4294

dibyom:
@spinnakerbot  cherry-pick 1.19

dibyom:

PR #1278, opened 2019-04-16
fix(build): Fix aws-iam-authenticator in docker container

ezimanyi:
Fixes spinnaker/spinnaker#4306

Updating kubectl to 1.14 broke the link to download the aws-iam-authenticator.  The highest currently-published version is for kubectl 1.12.7.

Let's downgrade kubectl to 1.12.7 (which is still an upgrade from the 1.10.3 it was in the prior version of halyard). In general as I think of it more, we should probably try to keep this as N-1 or N-2 of the newest kubectl version to try to have the most overlap with where users are.


ezimanyi:

jfillo:
@ezimanyi you should include a comment in this docker file indicating reiterating what you stated in issue#1278 . So whomever comes behind us later knows the reasoning the kubectl version isn't always kept at the latest

> as I thought of this more, we don't necessarily always want the newest kubectl version here. As kubernetes allows a version skew of +/-1 between kubectl and the cluster, we will maximize overlap with our users by staying with at most N - 1 where N is the most recent version.

PR #1289, opened 2019-04-24
fix(cli): Use "--no-validate" parameter on "hal config provider * account get"

dragon3:
## WHAT

This pull request fixes an issue that `hal config provider * account get` doesn't use `--no-validate` parameter.

## WHY

For example, the commands document says that `hal config provider docker-registry account get` supports the `--no-validate` option: 
https://www.spinnaker.io/reference/halyard/commands/#hal-config-provider-docker-registry-account-get

But actually it doesn't work properly because the `validate` option is hard-coded in the code.
( Currently always skip validation )

This pull request fixes the issue.

This issue was introduced in https://github.com/spinnaker/halyard/commit/668ad922ac854a1caede3d49fdfe45e39fa114be#diff-b7acb0a85fc298aec1a196698cbc8fdeR45


dragon3:
Thanks!

dragon3:
Just 1 question:
is this going to be released as 1.19.x? or I need to wait for 1.20?

ezimanyi:
We rarely do patch release unless there's a significant regression, so this will be in 1.20; I'd expect it to be released in the next few weeks.

dragon3:
Thank you for the reply!
I got it. I'm looking forward to 1.20!

PR #1291, opened 2019-04-25
chore(openstack): remove openstack provider

seanpeters86:
Removing OpenStack provider docs as per spinnaker/spinnaker#4316


emjburns:
LGTM

PR #1293, opened 2019-04-26
fix(gremlin): Fix the Halyard configuration of Gremlin

mattrjacobs:
Now programatically generate the config yaml for "integrations" section, based on Features object


mattrjacobs:
I previously misunderstood how properties were getting applied.  I followed the Kayenta example in this PR.  Now we're reading the `Features` object from `DeploymentConfiguration`, reading out the falg for if Gremlin should be enabled, and writing out YAML.  Because this applies to both `gate` and `orca`, I encapsulated this common logic in `IntegrationsConfigWrapper`

bhavith:
This change introduced duplicate entries in the generated yaml. The integrations session was getting printed twice and it was causing deployment failure - parsing error.

PR #1296, opened 2019-05-01
refactor(ci): Remove reflection from Cis class

ezimanyi:
* refactor(ci): Move CI service and controller their own package 

* refactor(ci): Combine CI+Master for Controller and Service 

  In the next commit we'll be subclassing these for each type of CI; it makes more sense to split the files along the lines of CI type than to have separate classes for the CI and the Master.

* refactor(ci): Make base Controller + Service abstract 

  In order to support static typing of CI classes, make the base controller and service abstract, and have each type of CI integration extend that class and supply the particular type of CI.

* refactor(ci): Remove reflection from Cis class 

  Now that the services and controllers use static typing, we can remove the reflection to dynamically determine the class to build.

* refactor(ci): Rename base class from Master -> CIAccount 

  The term 'Master' is very Jenkins-specific and does not apply to more general build systems. We'll leave the existing subclasses as is, but will rename the base class to CIAccount to better reflect that this represents a general account rather than a particular server.


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ci/CiService.java
```diff
@@ -18,53 +18,53 @@
...
 
 /**
  * This service is meant to be autowired into any service or controller that needs to inspect the current halconfigs
  * cis.
  */
-@Component
-public class CiService {
-  @Autowired
-  private LookupService lookupService;
+@RequiredArgsConstructor
```
jtk54: If all the sub-classes are meant to be spring beans, I think you can leave the `@Component` here.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ci/ConcourseService.java
```diff
@@ -0,0 +1,46 @@
...
+@Component
+public class ConcourseService extends CiService<ConcourseMaster, ConcourseCi> {
+    public ConcourseService(LookupService lookupService, ValidateService validateService) {
+        super(lookupService, validateService);
+    }
+
+    public String ciName() {
+        return "concourse";
+    }
+
```
jtk54: Extra line.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ci/CiController.java
```diff
@@ -31,129 +30,118 @@
...
+import org.springframework.stereotype.Component;
 import org.springframework.web.bind.annotation.*;
 
 import java.util.List;
 
-@RestController
-@RequiredArgsConstructor
-@RequestMapping("/v1/config/deployments/{deploymentName:.+}/ci")
-public class CiController {
+public abstract class CiController<T extends Master, U extends Ci<T>> {
```
jtk54: Little confused about what's going on here with the API root (e.g. `/v1/config/deployments/...`). Can you clarify?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ci/ConcourseController.java
```diff
@@ -0,0 +1,42 @@
...
+package com.netflix.spinnaker.halyard.controllers.v1.ci;
+
+import com.netflix.spinnaker.halyard.config.model.v1.ci.concourse.ConcourseCi;
+import com.netflix.spinnaker.halyard.config.model.v1.ci.concourse.ConcourseMaster;
+import com.netflix.spinnaker.halyard.config.services.v1.ci.ConcourseService;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/v1/config/deployments/{deploymentName:.+}/ci/concourse")
```
jtk54: OK nevermind --  I get it now.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/services/v1/ci/CiService.java
ezimanyi: I just tried this now, and it didn't seem like Spring was creating beans for the subclasses.

jtk54: I stand corrected, must have misremembered that.

halyard-web/src/main/java/com/netflix/spinnaker/halyard/controllers/v1/ci/CiController.java
ezimanyi: Each sub-controller is declaring a specific path for that CI system.

PR #1297, opened 2019-05-01
refactor(ci): Rename interface method getAccounts

ezimanyi:
* refactor(ci): Remove unused code 

  Some of the CI methods are unused, as we don't actually have any halyard commands or endpoints to list all CI accounts regardless of type.

* style(ci): Fix indentation 

  Some of the files I added in the last PR ended up with 4-space indents; fix this as a separate commit to reduce the diff on later commits.

* refactor(ci): Move more things to base classes 

  We can make the classes I added in the last PR a bit simpler by pushing some things to the base class.

* refactor(ci): Rename interface method getAccounts 

  We'll allow implementations of CIAccount to call their accounts field whatever they want (masters, accounts, etc.) but will have the top-level interface expose getAccounts().

  Push the 'masters' field down into the existing classes; we can't rename this field in those classes as it would break existing hal configs.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/master/AbstractListMastersCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/master/AbstractListMastersCommand.java
@@ -49,7 +49,7 @@ abstract class AbstractListMastersCommand extends AbstractCiCommand {
   @Override
   protected void executeThis() {
     Ci ci = getCi();
-    List<CIAccount> account = ci.getMasters();
+    List<CIAccount> account = ci.getAccounts();
     if (account.isEmpty()) {
       AnsiUi.success("No configured masters for " + getCiName() + ".");
     } else {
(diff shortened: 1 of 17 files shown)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/ci/master/AbstractListMastersCommand.java
```diff
@@ -49,7 +49,7 @@ private Ci getCi() {
   @Override
   protected void executeThis() {
     Ci ci = getCi();
-    List<CIAccount> account = ci.getMasters();
+    List<CIAccount> account = ci.getAccounts();
```
jtk54: May be worth a comment to clarify that 'masters' in the command is a veneer to make it more understandable.

ezimanyi: Good point; I'm likely going to need to make a slight change to this abstract command when I implement it for GCB so I'll drop a comment in that PR.

PR #1298, opened 2019-05-01
feat(gcb): Add GCB configuration to Halyard

ezimanyi:
This PR does not yet add commands for manipulating the config; these will be in the next PR.

* feat(gcb): Add GCB configuration to Halyard 

  The configuration should be added to both igor and echo. This commit only adds functionality for parsing these fields in the halconfig and generating the correct entries in the igor/echo config. A future commit will add 'hal' commands to edit this block in the halconfig.

* feat(gcb): Add service and controller for GCB config

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/gcb/GoogleCloudBuildAccount.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/gcb/GoogleCloudBuildAccount.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1.ci.gcb;
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.CIAccount;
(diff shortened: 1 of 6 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/EchoProfileFactory.java
```diff
@@ -68,6 +66,15 @@ protected void setProfile(Profile profile, DeploymentConfiguration deploymentCon
       profile.appendContents(yamlToString(deploymentConfiguration.getName(), profile, new ArtifactWrapper(artifacts)));
     }
 
+    Cis cis = deploymentConfiguration.getCi();
```
jtk54: The plurality of the function call doesn't match the return type, is that intentional?

ezimanyi: Yeah, unfortunately it's correct like this (though confusing).  It's an artifact of the fact that the deployment configuration field called `ci` is actually of type `Cis` (https://github.com/spinnaker/halyard/blob/e07f6c92b479000b3b7854fda9b5bd8a5eebdafc/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentConfiguration.java#L88)

jtk54: Got it, bummer.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/gcb/GoogleCloudBuildAccount.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class GoogleCloudBuildAccount extends CIAccount {
+  private String name;
+  private String project;
+  private String subscriptionName;
+  @LocalFile @SecretFile private String jsonKey;
```
Derenge: Hi! I was wondering why this was `jsonKey` instead of `jsonPath`? It would seem all of the other google accounts prefer `jsonPath`.

PR #1303, opened 2019-05-06
feat(kubernetes): Allow adding custom pod and service labels

bigkraig:
This allows adding customized labels to the pods and services created by Halyard for Spinnaker services.


ezimanyi:
@bigkraig : I need to revert this as it's causing deployment errors due to a jinja syntax error.  (There should not be two closing braces after the `% endfor %`.)  I missed that when reviewing.

bigkraig:
@ezimanyi oof. I just pushed up a fix

PR #1308, opened 2019-05-15
feat(s3): Add --path-style-access option to config s3 bucket access style (#4400)

tiggerlee2:
There are two styles to access s3 bucket: `virtual hosted-style` and `path-style`. Users can config which one they want to use.

I make this option default false which means using `virtual hosted-style` in considering `path-style` access is already deprecated by AWS([Amazon S3 Path Deprecation Plan – The Rest of the Story](https://aws.amazon.com/cn/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/)).


tiggerlee2:
spinnaker/spinnaker#4400

PR #1316, opened 2019-05-16
fix(*): Don't fail validation if openstack is present

ezimanyi:
We removed support for OpenStack, but this is causing Halyard to fail validation if there is an openstack block in the config. This is particularly bad as Halyard adds an empty block by default for each provider, so even users who have never used openstack are seeing validation failures after the release of 1.20.0.

Explicitly ignore the "openstack" field on deserialization so we don't cause validation errors.


ezimanyi:

PR #1318, opened 2019-05-17
fix(deploy): Add local profile to deployment configuration

ezimanyi:
The change to update clouddriver to use Spring Cloud Config is causing it to fail to startup when we don't specify 'local' as a profile. Let's add the 'local' profile here so that it discovers the cloud configuration.


ezimanyi:
This change adds the `local` profile to the bootstrap clouddriver deployment that is used when halyard spins up Spinnaker.  In order to prevent a regression where we now use undesired properties from people's existing `local` profile (which they intended for their actual clouddriver, not the bootstrap clouddriver), remove it from the pod entirely.  (This profile was mounted in the pod, but not being used up to now.)

Without this change, Halyard can no longer deploy to Kubernetes using the V1 provider as clouddriver fails to start up when the `local` profile is absent.

ezimanyi:
This was fixed in spinnaker/clouddriver#3683, which is a better way of fixing it than this PR.

PR #1320, opened 2019-05-20
fix(gcb): Fix gcb command parameters

ezimanyi:
The GCB command parameters are camel case, which is inconsistent with all other halyard commands. Fix this now before the feature is live in Spinnaker and users are depending on these.


ezimanyi:

PR #1322, opened 2019-05-21
fix(aws): AWS Credentials file not being created if keys are not in halconfig

estebangarcia:
Issue Summary:

If an AWS service is configured, such as S3 Persistent Store, but credentials are not set on halconfig the DefaultAWSCredentialsProviderChain (https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html) is used to get the secret and access keys and be able to validate the configuration.
But after successful validation the AWS credentials file is not being created using the values from the chain making the deployment of services depending on these configured AWS services fail.

Proposal:
Generate AWS credentials file with values from the DefaultAWSCredentialsProviderChain if AWS Keys are not present in halconfig

Issue Tracker: https://github.com/spinnaker/spinnaker/issues/4436

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/AwsCredentialsProfileFactoryBuilder.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/AwsCredentialsProfileFactoryBuilder.java
@@ -18,7 +18,11 @@
 
 package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile;
 
+import com.amazonaws.SdkClientException;
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentConfiguration;
+import com.netflix.spinnaker.halyard.core.error.v1.HalException;
+import com.netflix.spinnaker.halyard.core.problem.v1.Problem;
 import com.netflix.spinnaker.halyard.core.secrets.v1.SecretSessionManager;
 import com.netflix.spinnaker.halyard.deploy.services.v1.ArtifactService;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
@@ -26,9 +30,11 @@ import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSetting
 import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
+import lombok.AccessLevel;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.Setter;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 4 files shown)


ezimanyi:
@estebangarcia : Thanks for the PR!  This will need to be manually rebased on `master` as we've run a code formatter since this was opened that created a few conflicts.

If I understand what's happening here, it looks like this change will make `halyard` grab the credential ID and secret available to it via the `DefaultAWSCredentialsProviderChain` and write those values out to the config files that it generates for microservices.

The GCE provider has similar functionality to fall back to "Application Default Credentials" when no specific credentials are specified, but in that case what happens is that we simply write out the microservice config files with the credentials absent, and rely on each microservice to use its own default credentials to talk to the cloud provider.

I'm not too familiar with `DefaultAWSCredentialsProviderChain` but my expectation would be that it would behave in a similar way...if you don't explicitly specify credentials, Halyard should just generate config files with those credentials absent and then let each service discover its own credentials.  (If that's somehow not working that may be a separate bug to fix.)  It may be surprising to users to have Halyard write out the credentials in its environment to a file and pass it along to the other services.

CC a few people that are much more familiar with AWS than I am as they might have better ideas about this than I do: @clareliguori @robzienert @imosquera

estebangarcia:
Hey @ezimanyi, thanks for taking the time to review the PR. I've rebased the branch.

Yes, what you describe is what I'm proposing.  I'm not familiar myself with the way the GCE provider looks for the "default credentials" but I guess is similar to `DefaultAWSCredentialsProviderChain` which follows this order:

- Environment Variables - AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
- Java System Properties - aws.accessKeyId and aws.secretKey
- Credential profiles file at the default location (~/.aws/credentials)
- Credentials delivered through the Amazon EC2 container service
- Instance profile credentials delivered through the Amazon EC2 metadata service

This here might not be 100% the perfect solution, as it might surprise some users that are aware of halyard not writing the credentials file when keys are not present in halconfig and are hosting spinnaker on an EC2 environment using an attached IAM role. But I wanted to at least open the conversation on this matter and find the best solution for it.

It was surprising for me to see the `S3Validator` use the `DefaultAWSCredentialsProviderChain` but `KayentaService` and `ClouddriverService` don't follow that same pattern when generating the credentials file.

On a side note the usage help from halyard-cli when configuring the S3 as storage makes it a bit confusing in my opinion:

```
  --access-key-id
    Your AWS Access Key ID. If not provided, Halyard/Spinnaker will try to find AWS
    credentials as described at
    http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html#credentials-default
```

This could all just be lack of knowledge from my part on how to properly handle this :)

ezimanyi:
@estebangarcia : Thanks for the detailed explanation!  Let's wait to see what the people I tagged above who are more familiar with AWS suggest for how to handle this case.

clareliguori:
If I understand correctly, this PR is intended to cover the scenario when halyard is running somewhere that has access to AWS credentials through the default chain (like manually configured IAM user creds on a laptop?).  But then front50 is deployed somewhere that doesn't have access to creds through the default chain (for example?).

I believe that if creds are not provided in the halconfig, it is assumed that front50 will be deployed to something like an EC2 instance where it can get its creds through the instance role, and never have any static credentials.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/AwsCredentialsProfileFactoryBuilder.java
```diff
@@ -40,12 +46,44 @@
...
+  @Setter(AccessLevel.NONE)
+  private String secretAccessKey;
 
   @Setter private SpinnakerArtifact artifact;
 
+  public AwsCredentialsProfileFactoryBuilder setAccessKeyId(String accessKeyId) {
+    if (StringUtils.isEmpty(accessKeyId)) {
+      try {
+        this.accessKeyId =
+            DefaultAWSCredentialsProviderChain.getInstance().getCredentials().getAWSAccessKeyId();
```
clareliguori: This assumes that the creds are not temporary, for example not AssumeRole creds which would have a session token.  If I am running hal on an EC2 instance, it can pick up creds from the instance role.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/AwsCredentialsProfileFactoryBuilder.java
```diff
@@ -40,12 +46,44 @@
...
+      this.accessKeyId = accessKeyId;
+    }
+    return this;
+  }
+
+  public AwsCredentialsProfileFactoryBuilder setSecretAccessKey(String secretAccessKey) {
+    if (StringUtils.isEmpty(secretAccessKey)) {
+      try {
+        this.secretAccessKey =
+            DefaultAWSCredentialsProviderChain.getInstance().getCredentials().getAWSSecretKey();
```
clareliguori: By separating out the access key and secret key setters, there is a race condition here that the getCredentials result may have changed.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/Front50Service.java
```diff
@@ -82,9 +81,7 @@ public Type getType() {
     PersistentStore.PersistentStoreType type =
         deploymentConfiguration.getPersistentStorage().getPersistentStoreType();
     S3PersistentStore store = deploymentConfiguration.getPersistentStorage().getS3();
-    if (type == PersistentStore.PersistentStoreType.S3
-        && !StringUtils.isEmpty(store.getAccessKeyId())
-        && !StringUtils.isEmpty(store.getSecretAccessKey())) {
+    if (type == PersistentStore.PersistentStoreType.S3) {
```
clareliguori: I find it dangerous to assume that the user is OK with propagating their hal credentials to the deployed services.  For example, if hal retrieves my personal IAM user creds, then it's not safe to assume I also want front50 running as 'me'.  How can I disable this to let front50 run with EC2 instance creds instead of my own creds?  It seems like there should be a configuration flag to enable this new behavior.

robzienert:
Closing: It sounds like there's no resolution to the issues raised in the PR.

PR #1326, opened 2019-05-23
feat(kubernetes): Support cacheThreads and implement allNamespaces

ncknt:
This exposes `cacheThreads` in the CLI. Also added `allNamespaces` which was not implemented (I took it as clear the namespaces currently whitelisted).


ncknt:
@ezimanyi Could you PTAL?

PR #1327, opened 2019-05-27
fix(ldap): add managerDn and managerPassword to gate's auth configuration

estebangarcia:
Is possible to set managerDn and managerPassword for auth using halyard-cli but the values are not reflected in gate.yml


staciegravy:

PR #1328, opened 2019-05-28
Halyard to boot2

pdelagrave:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/SecurityConfig.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.netflix.spinnaker.halyard.config.config.v1;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+@Configuration
+public class SecurityConfig extends WebSecurityConfigurerAdapter {
```
ezimanyi: Was there an implicit permit all configuration in boot1 that now needs to be explicit?

halyard-web/src/main/java/com/netflix/spinnaker/halyard/Main.java
```diff
@@ -18,16 +18,34 @@
...
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
 @Configuration
-@ComponentScan({"com.netflix.spinnaker.config", "com.netflix.spinnaker.halyard",})
+@ComponentScan(
+    value = {
+      "com.netflix.spinnaker.config",
```
ezimanyi: Do you know what we're depending on from `com.netflix.spinnaker.config`?  I wonder if we actually even want to more explicitly pull in the few things we need from there, as we may end up breaking halyard if we add further things to clouddriver or front50.

build.gradle
```diff
@@ -51,22 +52,34 @@ allprojects {
...
+
+    annotationProcessor platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion")
+    annotationProcessor "org.projectlombok:lombok"
+    testAnnotationProcessor platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion")
+    testAnnotationProcessor "org.projectlombok:lombok"
+  }
+
+  configurations.all {
+    resolutionStrategy {
+      eachDependency { DependencyResolveDetails details ->
```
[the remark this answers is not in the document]
pdelagrave: Good point, it's right there in the docs: https://docs.gradle.org/current/userguide/managing_transitive_dependencies.html#sec:enforcing_dependency_version

Fixed in ff76cce

halyard-config/halyard-config.gradle
```diff
@@ -1,15 +1,40 @@
...
+  implementation 'com.amazonaws:aws-java-sdk-s3:1.11.534'
+  implementation 'com.google.apis:google-api-services-compute'
+  implementation 'com.google.apis:google-api-services-appengine:v1-rev92-1.25.0'
+  implementation 'com.microsoft.azure:adal4j:1.6.3'
+  implementation 'com.microsoft.azure:azure:1.19.0'
+  implementation 'commons-collections:commons-collections:3.2.1'
+  implementation 'commons-lang:commons-lang:2.6'
+  implementation 'commons-io:commons-io:2.6'
+  implementation 'io.fabric8:kubernetes-client'
+  implementation 'org.codehaus.groovy:groovy'
```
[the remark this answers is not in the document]
pdelagrave: Fixed in e2fabc9

halyard-backup/halyard-backup.gradle
```diff
@@ -1,12 +1,19 @@
...
-  compile spinnaker.dependency('bootWeb')
-  compileOnly spinnaker.dependency("lombok")
-  annotationProcessor spinnaker.dependency("lombok")
-  compile spinnaker.dependency('fabric8Api')
+  implementation 'org.springframework.boot:spring-boot-starter-actuator'
+  implementation 'org.springframework.boot:spring-boot-starter-web'
+  // TODO(lwander) Move to spinnaker-dependencies
+  implementation 'com.google.apis:google-api-services-cloudkms:v1-rev8-1.22.0'
+  implementation 'com.google.apis:google-api-services-storage'
+  implementation 'commons-lang:commons-lang:2.6'
```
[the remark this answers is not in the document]
pdelagrave: Fixed in 1798b36

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/SecurityConfig.java
pdelagrave: It wasn't implicit in boot1:
* https://github.com/spinnaker/halyard/blob/master/halyard-web/config/halyard.yml#L42
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#actuator-security

Now by default the actuator endpoints in boot2 would be protected by basic auth with the randomly generated password (using boot's autoconfigured user details services), unless at least one user-provided WebSecurityConfigurerAdapter is detected:
* https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#security

This is our case: `ActuatorEndpointsConfiguration` and `MetricsEndpointConfiguration`. They both permit requests to some endpoints but specify nothing else, leaving everything open. Which is fine but is kinda implicit and almost happening accidently as we include these security configs component by scanning the kork classes.

Hence I proposed to make the configuration explicit by excluding those beans:
* https://github.com/pdelagrave/halyard/blob/boot2/halyard-web/src/main/java/com/netflix/spinnaker/halyard/Main.java#L42

and by adding this `SecurityConfig` class.

ezimanyi: Sounds good, that makes perfect sense---thanks for the explanation!

halyard-web/src/main/java/com/netflix/spinnaker/halyard/Main.java
pdelagrave: I agree that it'd be cleaner to have a whitelist instead of a blacklist in that case, we probably don't use many and it'd make the bean dependency explicit.

@ComponentScan has some flexibility with include and exclude filters but for what we want to do that's bending it too much:
Scanning for everything in `com.netflix.spinnaker.halyard` but only selecting a few classes in `com.netflix.spinnaker.config`. This doesn't seem supported.

The kork and clouddriver @Component beans (anything under `com.netflix.spinnaker.config` really) could be refactored into @Configuration with @Conditionals so it'd be easier to integrate the optionally generated Beans when needed. That seems like the right approach for a library exposing spring beans.

ezimanyi: That makes sense, I figured there probably wouldn't be a way to do this without refactoring the beans in clouddriver and kork.  (I've in the past been surprised by things that kork brings in.)  But certainly out of scope for this boot2 upgrade.

ezimanyi:
@pdelagrave : I was just doing some smoke testing of this myself, and am seeing 403 errors when the CLI is trying to send requests to the daemon process.  (As one example, I tried to run `./hal config generate`.)

As one example:
```
curl -X POST http://localhost:8064/v1/config/deployments/default/generate/?validate=true
```
should be starting a task to generate the config file for the default deployment, but is returning a 403 after these changes.

But getting the config with
```
curl -v http://localhost:8064/v1/config/currentDeployment
```
works, so perhaps this is only affecting POST requests?

I'm wondering if something in the change to the `SecurityConfig` is the cause of this.

pdelagrave:
> @pdelagrave : I was just doing some smoke testing of this myself, and am seeing 403 errors when the CLI is trying to send requests to the daemon process. (As one example, I tried to run `./hal config generate`.)
> 
> As one example:
> 
> ```
> curl -X POST http://localhost:8064/v1/config/deployments/default/generate/?validate=true
> ```
> 
> should be starting a task to generate the config file for the default deployment, but is returning a 403 after these changes.
> 
> But getting the config with
> 
> ```
> curl -v http://localhost:8064/v1/config/currentDeployment
> ```
> 
> works, so perhaps this is only affecting POST requests?
> 
> I'm wondering if something in the change to the `SecurityConfig` is the cause of this.

thanks for reporting
csrf was still enabled, fixed in 7c86b75

halyard-web/src/main/java/com/netflix/spinnaker/halyard/Main.java
pdelagrave: Just committed what I think is a pretty satisfying compromise in a8ae5d1, let me know what you think.

Thanks to @scottfrederick and @jkschneider!

ezimanyi: That looks great, thanks!

PR #1330, opened 2019-05-31
fix(core): Fix halyard cli

ezimanyi:
The boot2 upgrade removed halyard-cli as a dependency of halyard-web. This was reasonable to do, as halyard-web does not directly depend on any functionality of halyard-cli. Unfortunately, the halyard install
process does not separately install halyard-cli; it only installs halyard-web into /opt/halyard but the CLI nonetheless expects to find its jar file there.

This means that currently the halyard-cli won't run when installed using the Halyard installer. While there's almost certainly a better way of implementing this, just re-add the dependency for now to unbreak halyard.

### DIFF
--- a/halyard-web/halyard-web.gradle
+++ b/halyard-web/halyard-web.gradle
@@ -34,6 +34,10 @@ dependencies {
   implementation 'org.codehaus.groovy:groovy'
 
   implementation project(':halyard-backup')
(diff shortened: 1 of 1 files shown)


ezimanyi:
CC @pdelagrave

ezimanyi:
Going to merge this now as Halyard is completely broken in the current state, and it's just reverting a small part of the prior commit.

PR #1332, opened 2019-06-03
feat(cf): Cloudfoundry account's API endpoint server certificate validated by default, skipping as an option.

pdelagrave:
The feature can be controlled with `--skip-ssl-certificate=true|false` on `add` and `edit` cloudfoundry account.


pdelagrave:
https://github.com/pdelagrave/halyard/blob/cf-api-skip-ssl/build.gradle#L20 will have to be updated to the clouddriver version containing this patch: spinnaker/clouddriver#3737

PR #1343, opened 2019-06-10
fix: Updated the httpGet readiness probe for spin-gate to support the HTTPS scheme (#4479)

rakewill:
The default spin-gate readiness probe that executes the external wget command was exposing a bug in the wget command in the busybox image. Every execution of the wget command leaves behind a default ssl_client process when used with a URL that uses the https scheme. We resolved this by setting kubernetes.useExecHealthCheck: false in the custom service settings file for gate. The httpGet probe is being used as expected after this change. However, Spinnaker is missing support for specifying that the HTTPS protocol be used by the readiness probe when the endpoints use TLS. This pull request adds the necessary support.


ezimanyi:
@rakewill: Can you please open a PR against the `master` branch?  Generally we accept PRs into that branch; in cases where it's warranted, we'll then cherry-pick a change into a release branch.

rakewill:
@ezimanyi: I created PR #1344 for this.

PR #1350, opened 2019-06-19
fix(deps): upgrade fabric8 to 4.1.1 in response to okhttp upgrade to …

maggieneterval:
…address connection leaks

- On Monday, Halyard pulled in an updated version of kork, which [updated its okhttp version](https://github.com/spinnaker/kork/pull/317). The new okhttp version throws errors on connection leaks. This caused the nightly integration tests to fail because Halyard's current fabric8 version has issues with connection leaks.
- Upgrades fabric8 library to version 4.1.1, which [addresses connection leaks](https://github.com/fabric8io/kubernetes-client/releases?after=kubernetes-client-1.4.14.redhat-000036).
- To do: upgrade Clouddriver's fabric8 version and address breaking HPA API change for the Kubernetes v1 provider.


maggieneterval:
cc @plumpy

PR #1355, opened 2019-06-21
feat(travis): Support travis property numberOfJobs

jervi:
Support for the new Travis property `numberOfJobs` introduced in spinnaker/igor#467. Also removes the old property `numberOfRepositories`. Should not be merged before spinnaker/igor#467 is merged.


jervi:
This is now ready to be merged.

plumpy:
I don't know if it's actually safe to merge this since it's totally possible that someone is using Halyard 1.27 with Spinnaker 1.16 (which still expects `numberOfRepositories`).

Which is unfortunate, since I don't think Halyard has a great story for that kind of conflict... You can say Spinnaker 1.17 requires Halyard >= 1.27, but I don't think there's a way to say Halyard 1.27 is incompatible with Spinnaker <= 1.17.

jervi:
That's a valid point. Should I add back the old property and just deprecate it?

plumpy:
I suppose arguably the right thing to do is to have Halyard support both properties, but it
a) won't write out `numberOfRepositories` if `version` >= 1.17
b) won't write out `numberOfJobs` if `version` < 1.17

Is that feasible?

jervi:
Closing in favour of #1717.

PR #1356, opened 2019-06-24
fix(core): Re-enable resolvedEnv endpoint

ezimanyi:
The resolvedEnv endpoint was inadvertently disabled as part of the boot2 upgrade because it moved to a package that's not in the classpath scan.  Re-enable it.


ezimanyi:
@spinnakerbot cherry-pick 1.21

PR #1362, opened 2019-06-30
list pods instead of namespaces during validation

cfedersp:
In a large org with structured permissions, Kubernetes namespace ops may not be available. This change lists pods during validation instead of namespaces. Since most or all critical Spinnaker ops are within a particular namespace, listing pods should be sufficient validation of the K8 connection. Even if there are no pods, thats OK since the purpose of this method is to ensure no exception is thrown.

### Issue 4582:
https://github.com/spinnaker/spinnaker/issues/4582


kevinawoo:
Hey all, this is a pretty old PR, so I'm going play janitor 🧹 and just close it for now. 

Feel free to create a new PR and we'll take a look again with a new set of eyes 👀 !

PR #1364, opened 2019-07-03
feat(secrets): Handle secrets in service settings

mtweten:
While experimenting with [secret support in halyard](https://www.spinnaker.io/reference/halyard/secrets/), I found that custom profiles and halconfig are supported, but secrets in [custom service settings](https://www.spinnaker.io/reference/halyard/custom/#custom-service-settings) are not.

There isn't a lot in service settings that might be considered secret except `password`, but in practice we also use environment variables to inject some secrets (for example, we have Spinnaker image builds with the new relic agent present, and the license key can be injected via environment variables). 

This change allows using the same secret syntax in custom service settings, but is only scoped to the password and environment fields. Is this a change people would be interested in? If so, would there be any reason to scope it further to any arbitrary fields in the service settings?


mtweten:
This was working previously but it seems like I've missed a dependency when preparing this for PR - looking into that.

PR #1367, opened 2019-07-09
feat(kubernetes): add opt-in livenessProbe with configurable initialD…

maggieneterval:
…elaySeconds

Closes https://github.com/spinnaker/spinnaker/issues/4023

- Adds two new sub-commands to the `hal config deploy edit` command, modeled off the code for the `vault` sub-commands : `--liveness-probe-enabled` and `--liveness-probe-initial-delay-seconds`.
- Adds liveness probe configuration to Kubernetes v1 and v2 providers, abstracting logic shared with readiness probes into its own method in each of the three affected classes.
- Adds `initialDelaySeconds` to manifest probe templates, and removes "readiness" from the name of each since these are now reused by liveness probes.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditDeploymentEnvironmentCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditDeploymentEnvironmentCommand.java
@@ -182,12 +182,12 @@ public class EditDeploymentEnvironmentCommand extends AbstractConfigCommand {
     vault.setEnabled(isSet(vaultEnabled) ? vaultEnabled : vault.isEnabled());
     deploymentEnvironment.setVault(vault);
 
-    livenessProbeConfig.setEnabled(
-        isSet(livenessProbesEnabled) ? livenessProbesEnabled : livenessProbeConfig.isEnabled());
-    livenessProbeConfig.setInitialDelaySeconds(
-        isSet(livenessProbeInitialDelaySeconds)
-            ? livenessProbeInitialDelaySeconds
-            : livenessProbeConfig.getInitialDelaySeconds());
+    if (isSet(livenessProbesEnabled)) {
+      livenessProbeConfig.setEnabled(livenessProbesEnabled);
+    }
+    if (isSet(livenessProbeInitialDelaySeconds)) {
+      livenessProbeConfig.setInitialDelaySeconds(livenessProbeInitialDelaySeconds);
+    }
     deploymentEnvironment.setLivenessProbeConfig(livenessProbeConfig);
 
     deploymentEnvironment.setLocation(
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditDeploymentEnvironmentCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditDeploymentEnvironmentCommand.java
@@ -114,6 +114,20 @@ public class EditDeploymentEnvironmentCommand extends AbstractConfigCommand {
       description = "This is the git user your github fork exists under.")
   private String gitOriginUser;
 
+  @Parameter(
+      names = "--liveness-probe-enabled",
+      arity = 1,
+      description =
+          "When true, enable Kubernetes liveness probes on Spinnaker services deployed in a Distributed installation. See docs for more information: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/")
+  private Boolean livenessProbesEnabled;
+
+  @Parameter(
+      names = "--liveness-probe-initial-delay-seconds",
+      arity = 1,
+      description =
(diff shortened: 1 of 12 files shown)


maggieneterval:
@plumpy if we could hold the next halyard release until this is merged I would be forever grateful :pray:

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v1/KubernetesV1DistributedService.java
```diff
@@ -417,24 +417,8 @@ default void resizeVersion(
...
-    int port = settings.getPort();
-    String scheme = settings.getScheme();
-    if (StringUtils.isNotEmpty(scheme)) {
-      scheme = scheme.toUpperCase();
-    } else {
-      scheme = null;
-    }
-
-    String healthEndpoint = settings.getHealthEndpoint();
     if (healthEndpoint != null) {
```
maggieneterval: This looks like a bigger diff than it is, really I just plopped all the probe handler stuff into a separate method

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/DeploymentEnvironmentValidator.java
```diff
@@ -112,4 +114,14 @@ private void validateDistributedDeployment(ConfigProblemSetBuilder p, Deployment
               + " is not in a provider that supports distributed installation of Spinnaker yet");
     }
   }
+
+  private void validateLivenessProbeConfig(ConfigProblemSetBuilder p, DeploymentEnvironment n) {
```
maggieneterval: I was thinking this might be confusing to people if they expect to be able to enable the probe in one command and add the delay seconds in a followup command, but hopefully if we document this well people will just copy/paste a functioning command from the docs (which would enable and set initialDelaySeconds in a single command)

maggieneterval: Also we can remove this error / set a reasonable default initialDelaySeconds in the future when service startup times are more consistent

ezimanyi: I think this is reasonable as is!  And agree that adding a default eventually would be good (once we are in a place where we can pick one).

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditDeploymentEnvironmentCommand.java
```diff
@@ -162,6 +182,14 @@ protected void executeThis() {
     vault.setEnabled(isSet(vaultEnabled) ? vaultEnabled : vault.isEnabled());
     deploymentEnvironment.setVault(vault);
 
+    livenessProbeConfig.setEnabled(
```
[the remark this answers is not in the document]
maggieneterval: Ah I agree, much clearer -- updated!

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/DeploymentEnvironment.java
```diff
@@ -134,4 +135,10 @@ public Boolean getUpdateVersions() {
     String upstreamUser = "spinnaker";
     String originUser;
   }
+
+  @Data
+  public static class LivenessProbeConfig {
+    boolean enabled;
```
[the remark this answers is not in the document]
maggieneterval: I thought the same thing but left it as lowercase `boolean` because the classes above (`Consul`, `Vault`) use lowercase `boolean enabled` -- I think the `isSet` check in `EditDeploymentEnvironmentCommand` is preventing it from being reset to null if only the delay seconds are passed in. I just tested to make sure it works, so I think we're okay :woman_shrugging:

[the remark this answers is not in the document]
ezimanyi: Sounds good, I stand corrected!

davidxia:
@maggieneterval is there a way to tell what Spinnaker version here this will be released as? https://www.spinnaker.io/community/releases/versions/

maggieneterval:
@davidxia Halyard is actually released separately from Spinnaker, but I expect @plumpy will release Halyard 1.22.0 with this change next week? :pray:

davidxia:
Ah, that's right. Thanks for explaining. Keep me posted on the release. Is there a page I can see the releases or changelog for halyard?

maggieneterval:
I will definitely let you know and also comment on the original issue so anyone following knows as well! As far as I know we are not as good about documenting the halyard releases on spinnaker.io, but at least the last commit in each release is documented here: https://github.com/spinnaker/halyard/releases
Generally, I think the #halyard channel in slack is currently the best place to be notified of halyard releases specifically since we don't post about them in #spinnaker-releases.

PR #1368, opened 2019-07-09
feat(core): Support external configuration using Spring Cloud Config

scottfrederick:
This PR adds the following features related to external configuration using Spring Cloud Config Server (see the [Spinnaker Config Secrets design doc](https://docs.google.com/document/d/1Y7SIgobc8_CCFmFZLRUGl-76sfHWlprFYGy992S5U4A/edit#heading=h.x4tnoxo2zzyn))

* When the CLI detects that an account configuration file contains the prefix `configserver:`, it doesn't convert to an absolute local path
* On validation, if a field that is required for live validation of a connection to an account contains a property placeholder (e.g. `${some.value}`) that should be resolved via Config Server at service runtime, then live account validation is bypassed
* On validation, if a field that is used as an account configuration file contains the prefix `configserver:`, then the contents of the file is not validated and live account validation is bypassed
* On deployment, if the default kubernetes account used for deployment is configured with a `kubeconfigFile` containing the prefix `configserver:`, then the contents of the file will be retrieved from a config server backend and used for the deployment

This PR contains the following related refactorings: 
* `Validator` subclasses all use the `SecretSessionManager` that already existed in the base class instead of defining their own `SecretSessionManager`
* All attempts to load account configuration files use `ValidatingFileReader` (via `Validator#validatingFileDecrypt` where possible)
* `ValidatingFileReader` was moved to eliminate a package cycle

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/canary/prometheus/PrometheusCanaryAccountValidator.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/canary/prometheus/PrometheusCanaryAccountValidator.java
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.halyard.config.validate.v1.canary.prometheus;
 
 import static com.netflix.spinnaker.halyard.core.problem.v1.Problem.Severity.ERROR;
 
-import com.google.common.base.Strings;
 import com.netflix.spinnaker.halyard.config.model.v1.canary.AbstractCanaryAccount;
 import com.netflix.spinnaker.halyard.config.model.v1.canary.prometheus.PrometheusCanaryAccount;
 import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
@@ -55,7 +54,7 @@ public class PrometheusCanaryAccountValidator extends CanaryAccountValidator {
     if (StringUtils.isNotEmpty(usernamePasswordFile)) {
       String usernamePassword = validatingFileDecrypt(p, usernamePasswordFile);
 
-      if (Strings.isNullOrEmpty(usernamePassword)) {
+      if (usernamePassword != null && StringUtils.isEmpty(usernamePassword)) {
         p.addProblem(ERROR, "The supplied username password file does not exist or is empty.")
             .setRemediation("Supply a valid username password file.");
       } else if (!usernamePassword.contains(":")) {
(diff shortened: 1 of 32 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/canary/prometheus/PrometheusCanaryAccountValidator.java
```diff
@@ -55,7 +54,7 @@ public void validate(ConfigProblemSetBuilder p, AbstractCanaryAccount n) {
     if (StringUtils.isNotEmpty(usernamePasswordFile)) {
       String usernamePassword = validatingFileDecrypt(p, usernamePasswordFile);
 
-      if (Strings.isNullOrEmpty(usernamePassword)) {
+      if (usernamePassword != null && StringUtils.isEmpty(usernamePassword)) {
```
ezimanyi: Shouldn't this `&&` be an `||` to keep the current behavior (or was the change deliberate)?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dcos/DCOSClusterValidator.java
```diff
@@ -7,15 +7,12 @@
 import com.netflix.spinnaker.halyard.config.model.v1.node.Validator;
 import com.netflix.spinnaker.halyard.config.model.v1.providers.dcos.DCOSCluster;
 import com.netflix.spinnaker.halyard.config.problem.v1.ConfigProblemSetBuilder;
-import com.netflix.spinnaker.halyard.core.secrets.v1.SecretSessionManager;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.apache.commons.lang.StringUtils;
```
ezimanyi: We generally use `org.apache.commons.lang3` (something must be pulling the `lang` version onto the classpath though).

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/canary/prometheus/PrometheusCanaryAccountValidator.java
claymccoy: It is hard to guess intent, but here goes...
I'd guess that it should not  be && because that leaves the 'else if' condition the chance of throwing an NPE when usernamePassword is null. Making it || is redundant since StringUtils.isEmpty checks for null and "". Given the context, the only thing that makes sense is for it to just be StringUtils.isEmpty(usernamePassword) unless the surrounding code is further modified. I'm going to make that change.

claymccoy: fixed

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dcos/DCOSClusterValidator.java
claymccoy: fixed

PR #1370, opened 2019-07-11
feat:(provider/Kubernetes) Add ability to set deployment strategy for service settings

clanesf:
This expands the existing kubernetes custom service settings to allow the end user to choose and configure a Kubernetes deployment strategy in the service settings yml. 

Right now this supports both options offered by Kubernetes out of the box, `Recreate` and `RollingUpdate` with the ability to tweak the underlying `RollingUpdate` settings of `maxSurge` and `maxUnavabile` with providing extensibility if additional settings are added later.

To make use of this, add the following block to to your service settings yaml, for example if you wanted to replace all Clouddriver pods when a new configuration is rolled out and are comfortable with double the deployment, you can use a `clouddriver.yml` in `~/.hal/<profile>/service-settings` that contains the following
```
kubernetes:
  deploymentStrategy: 
    type: rollingUpdate
    strategySettings:
      maxSurge: 100%
      maxUnavailable: 50%
```

If on the other hand you wanted to use the `Recreate` deployment strategy, then you can use the following `clouddriver.yml` in your `~/.hal/<profile>/service-settings/` directory

```
kubernetes:
  deploymentStrategy: 
    type: recreate
```

I've tested this using both of these configurations and have successfully changed a service's deployment. 

In particular I would like this upstreamed because we are trying to automate Spinnaker deployment to new environments and being able to tweak this setting is key to getting clouddriver pods to all recycle at once when our automation adds a new account.


clanesf:
@ethanfrogers  - As discussed a couple days ago, here is a PR to add strategy to the service-settings for halyard

halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
```diff
@@ -19,6 +19,14 @@ spec:
     matchLabels:
       app: spin
       cluster: spin-{{ name }}
+  {% if deploymentStrategy != null %}
+  strategy:
+    {{ deploymentStrategy.type }}:
```
ethanfrogers: It might be good to be a bit more guarded here and only inject this block if `type == rollingUpdate`. If you switch from `rollingUpdate` to `recreate` but leave the `strategySettings` in place it could result in an invalid manifest.

ethanfrogers:
cc @ezimanyi for an extra set of eyes!

halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
ezimanyi: +1

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeploymentStrategy.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Data;
+
+@Data
```
ezimanyi: I'd suggest either using `@Data` to generate accessors or leaving the fields public, but not both.

halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
clanesf: Sounds good. I thought about this too, but the only reason I didn't limit it to rollingUpdate explicitly was to try to give room to support other update types in the future, but then it would have to change anyways because of the spelling/casing difference below. 
I'll make this change and push it up.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeploymentStrategy.java
clanesf: Showing my lack of experience with Java/lombok here :) I'll get this fixed. Thanks!

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeploymentStrategy.java
```diff
@@ -0,0 +1,16 @@
+package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service;
```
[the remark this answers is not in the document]
clanesf: Added.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeploymentStrategy.java
clanesf: Fixed.

halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
clanesf: The length check is necessary so that someone changes from recreate back to rollingUpdate and doesn't specify any settings it won't create a bad manifest.

halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
```diff
@@ -19,6 +19,16 @@ spec:
...
+  {% if deploymentStrategy != null %}
+  strategy:
+    {% if deploymentStrategy.type == "rollingUpdate" and deploymentStrategy.strategySettings|length > 0 %}
+    {{ deploymentStrategy.type }}:
+      {% for key, value in deploymentStrategy.strategySettings.items() %}
+      {{ key }} : {{ value }}
+      {% endfor %}
+    {% endif %}
+    type: {% if deploymentStrategy.type == "recreate" %} Recreate {% else %} RollingUpdate {% endif %}
+  {% endif %}
```
clanesf: Minor thing, I changed the order here so that the default is `RollingUpdate` not `Recreate` since the current default for Kubernetes if not specified is `RollingUpdate`

ezimanyi:
Looks good!  I'll let @ethanfrogers take one more look, but all set to merge from my perspective.

ethanfrogers:
Looks great to me @clanesf! Nice work! I'll merge now! Would you care to add some documentation for spinnaker.io? We've gotten into a bad habit of not documenting new Kubernetes service settings.

PR #1376, opened 2019-07-24
fix(config): parse 'oauthScopes' stanzas that were incorrectly written

plumpy:
The old version of Lombok used in Halyard <=1.21 didn't copy
@JsonProperty annotations to the generated methods. This means Jackson
was seeing the field as 'oAuthScopes' and the getter/setter as a
separate 'oauthScopes' property. The same data would be written to both
properties, and during parsing whichever came last in the file would be
persisted into the object.

With new versions of Lombok (>=1.18.8), the @JsonProperty annotation is
copied to the bean methods, so the 'oauthScopes' property disappears and
Jackson can no longer parse those old files.

This commit adds some methods to support parsing the files generated by
previous versions of Halyard, but will no longer write out the
incorrect duplicate data.

### DIFF
--- a/halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccountTest.java
+++ b/halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccountTest.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 Google, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/model/v1/providers/kubernetes/KubernetesAccountTest.java
```diff
@@ -0,0 +1,97 @@
+package com.netflix.spinnaker.halyard.config.model.v1.providers.kubernetes;
```
ezimanyi: Missing header

plumpy: Done, thanks

plumpy:
@spinnakerbot cherry-pick 1.21

plumpy:
@spinnakerbot cherry-pick 1.22

PR #1379, opened 2019-07-25
feat(secret): decrypt secrets before sending to deck

KathrynLewis:
Deck doesn't have secret decryption implemented, but we still want to allow users to use the secret format for SSL certs, etc, in their hal config. The change here makes sure those secrets are decrypted before building the deck profile.

@ezimanyi PTAL whenever you have a chance? (or let me know if I should tag someone else?)
Thanks!

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/ApacheSpinnakerProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/ApacheSpinnakerProfileFactory.java
@@ -31,6 +31,7 @@ import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.SpinnakerServic
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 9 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/ApacheSpinnakerProfileFactory.java
```diff
@@ -77,13 +78,24 @@ protected void setProfile(
 
   @Override
   protected Map<String, Object> getBindings(
-      DeploymentConfiguration deploymentConfiguration, SpinnakerRuntimeSettings endpoints) {
+      DeploymentConfiguration deploymentConfiguration,
+      Profile profile,
+      SpinnakerRuntimeSettings endpoints) {
     TemplatedResource resource = new StringResource(SSL_TEMPLATE);
     Map<String, Object> bindings = new HashMap<>();
+    BindingsSecretDecrypter bsc =
```
plumpy: Creating your service dependencies with `new` is generally frowned upon. Instead, you should have Spring inject your dependencies.

You probably want a `@Component BindingsSecretDecrypterFactory` with a `SecretSessionManager` parameter and a `createDecrypter(Profile profile, String decryptedOutputDirectory)` method. Then inject that factory into an `@Autowired` constructor of this class...

plumpy:
Actually, I see spinnaker/spinnaker#3649 already addresses my concern. "The immediate goal of this proposal is to enable Halyard users to version their configuration", which makes sense. I wasn't thinking of that use case.

I'm still curious about the reason why Deck can't do the decryption, but yeah, I guess that shouldn't hold up this PR.

KathrynLewis:
Thanks so much for the review @plumpy ! I think it's much cleaner now. I updated but didn't get to leaving a comment, so thanks for following up.

PR #1380, opened 2019-07-25
fix(core): Fix reading of external files as binary instead of text

scottfrederick:
Fixes https://github.com/spinnaker/spinnaker/issues/4665

Also further consolidates all external file access into `ValidatingFileReader`.


plumpy:
@spinnakerbot cherry-pick 1.22

PR #1382, opened 2019-08-01
fix(eks): Halyard doesn't work with new EKS kube config

costimuraru:
Fixes https://github.com/spinnaker/spinnaker/issues/4712


costimuraru:
@ezimanyi @maggieneterval can you help with a quick review here, please?

Dockerfile.local
```diff
@@ -28,6 +28,10 @@ RUN curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/
 
 ENV PATH "$PATH:/usr/local/bin/aws-iam-authenticator"
 
+RUN wget -O /tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py && \
```
maggieneterval: Not familiar with pip at all or whether this matters, but it looks like in Clouddriver we are specifying a certain version of `awscli`?
https://github.com/spinnaker/clouddriver/blob/master/Dockerfile.slim#L23

ezimanyi:
This might be something that would be better for @clareliguori to look at, as I don't know anything about EKS or `awscli`.

Dockerfile.local
costimuraru: Good point, @maggieneterval! Probably best to pin this to a specific version to have a more predictable behavior, in cases something breaks with future versions. I made the change.

Dockerfile.local
```diff
@@ -28,6 +28,10 @@ RUN curl -o aws-iam-authenticator https://amazon-eks.s3-us-west-2.amazonaws.com/
...
+    python /tmp/get-pip.py && \
+    pip install --upgrade awscli==1.16.208
```
clareliguori: The `aws eks get-token` command is available in version 1.16.232 or greater of the AWS CLI:
https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html

costimuraru: I believe the AWS docs just links toward the latest versions (1.16.232 was released a few days ago, however the `aws eks get-token` has been around for a few months already 1.16.158): https://pypi.org/project/awscli/#history

I would keep this to 1.16.208 just to be in sync with clouddriver, which suffered from the same issue and was recently patched: https://github.com/spinnaker/clouddriver/pull/3997

What do you tink @clareliguori ?

clareliguori: Hmm I'm just going off what the docs say, I don't work much on the EKS side of things

@micahhausler do you know what the story is here with the eks get-token command and AWS CLI versions?  Is it safe to lock to 1.16.208?

costimuraru:
By the way, we tested this at Adobe and it works as expected:

```
+ Successfully added account ee-prd for provider aws.
+ kubectl exec --namespace spinnaker spinnaker-spinnaker-halyard-0 -- sh -c 'mkdir -p /home/spinnaker/.hal/default/profiles/rosco/packer'
+ kubectl exec --namespace spinnaker spinnaker-spinnaker-halyard-0 -- sh -c 'base64 -d /opt/halyard/additionalConfigMaps/aws-ebs.json.base64 > /home/spinnaker/.hal/default/profiles/rosco/packer/aws-ebs.json'

+ hal --daemon-endpoint http://spinnaker-spinnaker-halyard:8064 config artifact helm enable
+ Get current deployment
  Success
+ Edit the helm provider
  Success
```

PR #1385, opened 2019-08-06
fix(build): remove old usage of korkVersion

KathrynLewis:
Spinnakerbot now uses gradle.properties to autobump the kork version ([example](https://github.com/spinnaker/halyard/commit/06f4375202a66b3d0061f6ce0f9460393617b36e)), so this line can cause conflicts.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,6 @@
 
 buildscript {
   ext {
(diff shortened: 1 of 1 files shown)


plumpy:
Thank you!

cc @maggieneterval do you think you could merge this? Thanks!

PR #1386, opened 2019-08-08
feat(plugins): adding halyard commands for plugins

link108:
This PR is related to https://github.com/spinnaker/spinnaker/issues/4181.

This PR adds commands to Halyard to configure plugins, as well as lay down `plugins.yml` and configuration updates to `orca.yml` on deploy.
These files are used by Kork in order to enable plugins (https://github.com/spinnaker/kork/pull/325).

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
@@ -19,15 +19,21 @@ package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile;
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentConfiguration;
 import com.netflix.spinnaker.halyard.config.model.v1.node.Features;
 import com.netflix.spinnaker.halyard.config.model.v1.node.Webhook;
+import com.netflix.spinnaker.halyard.config.model.v1.plugins.Plugin;
 import com.netflix.spinnaker.halyard.config.model.v1.providers.aws.AwsProvider;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSettings;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.integrations.IntegrationsConfigWrapper;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 import lombok.Data;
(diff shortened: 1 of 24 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
```diff
@@ -73,6 +79,20 @@ protected void setProfile(
...
     profile.appendContents("pipelineTemplate.enabled: " + pipelineTemplates);
+
+    final List<Plugin> plugins = deploymentConfiguration.getPlugins().getPlugins();
+    Map<String, Object> fullyRenderedYaml = new LinkedHashMap<>();
+    Map<String, Object> pluginMetadata =
+        plugins.stream()
+            .filter(p -> p.getEnabled())
+            .filter(p -> !p.getManifestLocation().isEmpty())
+            .map(p -> p.generateManifest())
+            .collect(Collectors.toConcurrentMap(m -> m.getName(), m -> m.getOptions()));
```
ezimanyi: Does this map need to be concurrent?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
```diff
@@ -73,6 +79,20 @@ protected void setProfile(
     profile.appendContents("pipelineTemplates.enabled: " + pipelineTemplates);
     // For backward compatibility
     profile.appendContents("pipelineTemplate.enabled: " + pipelineTemplates);
+
+    final List<Plugin> plugins = deploymentConfiguration.getPlugins().getPlugins();
```
ezimanyi: Why are we directly appending to the `orca.yml` config file?  Aren't we writing out a separate `plugins.yml` file that will contain the plugin config and that each service supporting plugins will look at?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
link108: Nope, was thinking of making this a parallel stream, but I think thats overkill. Moving to `Collectors.toMap` instead

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
link108: We are writing out a `plugins.yml` file, it is for use prior to spring context initialization and contains the plugin names and the locations of jars to download. `orca.yml` will contain the configurations for the plugins, eg `url` for a service.

PR #1393, opened 2019-08-15
feat(plugins): enable/disable downloading plugins

link108:
This PR is related to spinnaker/spinnaker#4181 and enables Spinnaker administrators to use halyard to enable downloading of plugin jars, as seen in https://github.com/spinnaker/kork/pull/358. 

Different environments used to run Spinnaker have varying constraints, in order to ensure plugins can be used in all environments, we added the ability to download plugins via Kork at runtime. By default, downloading of plugin jars will be disabled, but if the Spinnaker administrator trusts the download location, they can enable plugin jar downloading. Spinnaker users (application owners) will not have the ability to add plugins.


ethanfrogers:
cc @ezimanyi!

PR #1398, opened 2019-08-20
feat(artifacts): add feature flag to turn on artifactsRewrite

Jammy-Louie:
add the ability to turn on the artifacts rewrite  through halyard

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -5661,6 +5661,7 @@ hal config features edit [parameters]
 #### Parameters
  * `--appengine-container-image-url-deployments`: Enable appengine deployments using a container image URL from gcr.io.
  * `--artifacts`: Enable artifact support. Read more at [https://spinnaker.io/reference/artifacts/](https://spinnaker.io/reference/artifacts/)
+ * `--artifactsRewrite`: Enable new artifact support. Read more at [https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/](https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/)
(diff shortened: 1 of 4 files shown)


Jammy-Louie:
depends on: https://github.com/spinnaker/deck/pull/7347

docs/commands.md
```diff
@@ -5661,6 +5661,7 @@ hal config features edit [parameters]
 #### Parameters
  * `--appengine-container-image-url-deployments`: Enable appengine deployments using a container image URL from gcr.io.
  * `--artifacts`: Enable artifact support. Read more at [https://spinnaker.io/reference/artifacts/](https://spinnaker.io/reference/artifacts/)
+ * `--artifactsRewrite`: Enable new artifact support. Read more at [https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/](https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/)
```
maggieneterval: I *think* parameters are typically kebab-case

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Features.java
```diff
@@ -49,6 +49,12 @@ public NodeIterator getChildren() {
           "Artifacts are not configurable prior to this release. Will be stable at a later release.")
   private Boolean artifacts;
 
+  @ValidForSpinnakerVersion(
+      lowerBound = "1.5.0",
```
louisjimenez: I could be off here but this seems a little low, the rewrite was available starting in 1.5?

docs/commands.md
Jammy-Louie: 👍

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Features.java
Jammy-Louie: 👍

PR #1405, opened 2019-08-27
feat(slack): Allow configurable slack endpoint

diasjorge:
Allow slack notifications to be sent to compatible implementations.

spinnaker/spinnaker#4634

spinnaker/echo#633


diasjorge:
@spinnaker/reviewers the build seems to be failing due to some missing documentation but I think it's unrelated

PR #1406, opened 2019-08-29
feat(telemetry): adds endpoint and enable settings for stats collection

staciegravy:
Adds new configuration fields for telemetry. 

```
telemetry
  enabled (enabled by default)
  endpoint
  instanceId
```

[spinnaker/spinnaker#4738](https://github.com/spinnaker/spinnaker/issues/4738)

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
@@ -45,7 +45,9 @@ public class TelemetryEditCommand extends AbstractConfigCommand {
             .setFailureMesssage("Failed to load telemetry settings.")
             .get();
 
-    telemetry.setEndpoint(isSet(endpoint) ? endpoint : telemetry.getEndpoint());
+    if (isSet(endpoint)) {
+      telemetry.setEndpoint(endpoint);
+    }
 
     new OperationHandler<Void>()
         .setOperation(Daemon.setTelemetry(currentDeployment, !noValidate, telemetry))
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.halyard.cli.command.v1.config.telemetry;
+
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.Parameters;
+import com.netflix.spinnaker.halyard.cli.command.v1.config.AbstractConfigCommand;
+import com.netflix.spinnaker.halyard.cli.services.v1.Daemon;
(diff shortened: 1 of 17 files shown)


ttomsu:
I think the command being called `stats` instead of `telemetry` would be easier for folks (especially non native English speakers) to understand. Thoughts @ezimanyi @demobox @@imosquera?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
```diff
@@ -0,0 +1,56 @@
...
+  @Override
+  protected void executeThis() {
+    String currentDeployment = getCurrentDeployment();
+    Telemetry telemetry =
+        new OperationHandler<Telemetry>()
+            .setOperation(Daemon.getTelemetry(currentDeployment, false))
+            .setFailureMesssage("Failed to load telemetry settings.")
+            .get();
+
+    telemetry.setEndpoint(isSet(endpoint) ? endpoint : telemetry.getEndpoint());
```
ezimanyi: Might be more clear as:
```
if (isSet(endpoint)) {
  telemetry.setEndpoint(endpoint);
}
```
As it, it's a bit confusing that the `false` branch of the ternary is just the no-op `telemetry.setEndpoint(telemetry.getEndpoint());`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
```diff
@@ -0,0 +1,35 @@
...
+@Data
+@EqualsAndHashCode(callSuper = false)
+public class Telemetry extends Node {
+
+  @Override
+  public String getNodeName() {
+    return "telemetry";
+  }
+
+  private Boolean enabled = true;
```
ezimanyi: Are we sure that we want to enable this by default at the same time we're rolling this out to everyone?  Would this benefit from an opt-in period before moving to opt-out?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
```diff
@@ -0,0 +1,35 @@
...
+  private String endpoint = "https://fix-me";
```
ezimanyi: This needs to be fixed.

demobox:
> I think the command being called stats instead of telemetry would be easier for folks

There may be an implied difference in terms of where the data ends up - "stats" could easily be read as "collect statistics locally", although that may also be true of "telemetry"?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
ezimanyi: I've been changing this everywhere I'm already touching and trying to write new code more clearly.  Unfortunately 90% of halyard is people copy-pasting existing code, so if we don't make an effort to make it better it will only get worse.

I'm not overly concerned about this particular pattern, but I *am* concerned about allowing more and more unclear code in Halyard because there's already so much unclear code, as that will get harder and harder to maintain.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
ezimanyi: You mean opt-*in* until 1.17, right?

staciegravy: Just merge changes that make this opt-in for now.

staciegravy:
I agree with @demobox that both terms could be confusing but I also agree that stats is likely an easier term for non-native English speakers. If we want to change the name, we'd also need to do that in the corresponding Echo PR #640. Does anyone have any strong feelings either way? @ezimanyi @plumpy @imosquera

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
ttomsu: Yes, opt-in, sorry.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/telemetry/TelemetryEditCommand.java
staciegravy: @ezimanyi I've addressed your feedback here. :)

PR #1407, opened 2019-08-29
refactor(deploy): Adapt to changes in kork-config

scottfrederick:
depends on https://github.com/spinnaker/kork/pull/370


claymccoy:
Just moving logic around, but it looks like there are some compilation/import issues. After that it LGTM.

scottfrederick:
> looks like there are some compilation/import issues

Yep, those will be resolved once the dependent kork PR is in a release and the kork version is bumped.

PR #1411, opened 2019-09-03
feat(secrets): Support SAML metadata as secret

KathrynLewis:
Corresponding kork PR: https://github.com/spinnaker/kork/pull/372

There have been some requests to support the SAML metadata.xml file as a secret. This does so on the Halyard side and moves the SAML "file:" prefix logic to the object mapper. `metadataRemote` and `metadataLocal` in the halfconfig both get assigned to `metadataUrl` in the service yamls, so we only add the "file:" prefix if `metadataUrl` is a file path.

@plumpy PTAL?


KathrynLewis:
@plumpy or @ezimanyi are either of you able to take a look at this one?

PR #1416, opened 2019-09-05
feat(halyard/localfiles): Support relative local files in hal config

german-muzquiz:
Usually when referencing local files in main hal config with a relative path, halyard throws this error:
```
Problems in Global:
! ERROR Failed to backup user file: default/files/kubeconfig-main

- Failed to generate config.
```
With this change, relative files will be automatically resolved relative to hal config home.
Usually the above error doesn't happen because when making changes with halyard cli, absolute file paths are automatically added to relative file references in that moment, but if users manually edit `config` file to specify portable, relative file references, the above error shows up.


german-muzquiz:
cc @ezimanyi

german-muzquiz:
By the way, I was having an issue when halyard tried to read a kubeconfig from staging directory, the file contents was not read if it wasn't encrypted or coming from ConfigServer (look at `KubernetesV2ClouddriverProfileFactory.java`, `getKubconfigFileContents`.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -414,6 +418,83 @@ public NodeDiff diff(Node other) {
...
+                        String fPath = (String) f.get(n);
+                        if (StringUtils.isEmpty(fPath)) {
+                          return;
+                        }
+                        if (fPath.startsWith(LocalFile.RELATIVE_PATH_PLACEHOLDER)
+                            || Paths.get(fPath).isAbsolute()) {
+                          return;
+                        }
+
+                        Path absolutePath = Paths.get(prefix + File.separator + fPath).normalize();
```
ncknt: `Paths.get(prefix, fPath).normalize()`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -414,6 +418,83 @@ public NodeDiff diff(Node other) {
...
+
+                        String currentPath = (String) f.get(n);
+                        if (StringUtils.isEmpty(currentPath)) {
+                          return;
+                        }
+
+                        // restore original relative path
+                        String relativePath = relativeFileReferences.get(f);
+                        Path absolutePath =
+                            Paths.get(prefix + File.separator + relativePath).normalize();
```
ncknt: `Paths.get(prefix, relativePath).normalize()`

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -53,6 +54,9 @@
 @Slf4j
 public abstract class Node implements Validatable {
 
+  // @LocalFile fields whose file path is relative
+  @JsonIgnore private Map<Field, String> relativeFileReferences = new ConcurrentHashMap<>();
```
ncknt: We should keep a reference by Node + Field - not just Field, otherwise a Node using more than one relative path won't be tracked/restored.

german-muzquiz: The map is not globally unique, there's a map for each instance of `Node`, and `Field` is already unique in a given node hierarchy. I did a test with two kubernetes accounts pointing to relative kubeconfig files (same `Field` definition, two different node instances), and both were correctly generated in `staging` directory.

ncknt: Ah good point, missed that.

plumpy:
Sorry, I've been neglectful, I'm going to look through this tomorrow morning. Thanks!

plumpy:
I guess all think monkeying around with the data inside `Node` just seems weird. (I realize we do it other places, but that's also weird.)

Instead of prepending `{%halconfig-dir%}` to relative files, why not just fix `backupLocalFiles` to ignore relative paths?

Alternate plan: why not just tell users they have to be explicit and specify `{%halconfig-dir%}` in the config if they want to use a relative path? Obviously the error message would have to say this, so the work would be relatively similar, but something like "`Cannot find file '%s'. If you want to reference files within the '.hal' directory, prepend them with '{%halconfig-dir%}'.`"

ncknt:
I'd also like to keep a single way to do relative path but definitely prefer without `{%halconfig-dir%}`.

german-muzquiz:
`{%halconfig-dir%}` is only used by `hal backup create` and `hal backup restore`. When a backup is created, Halyard copies all referenced files inside the tarball and replaces absolute paths with that token prefix. Upon restore, files are extracted to a directory inside hal home (`.hal/.backup`), the token prefix is removed and all file references are set with absolute paths. Using `{%halconfig-dir%}` even breaks other parts of Halyard outside backup operations, having a config file with that token and doing a `hal deploy apply` produces this error:
```
- Get current deployment
  Failure
Problems in Global:
! ERROR Could not parse your halconfig: while scanning for the next
  token
found character '%' that cannot start any token. (Do not use % for indentation)
in 'reader', line 88, column 26:
kubeconfigFile: {%halconfig-dir%}/.armory/kubecfg ...
^
- Failed to get deployment name.
```
Now that the intent is to support relative file references, I think we can just remove the usage of that token altogether. 
The goal of supporting relative paths is that users want to be able to share configs between machines without worrying to have a specific directory structure. I'll look for how can this be improved.

german-muzquiz:
@plumpy I moved the map that keeps track of file paths outside of `Node` to the parser. Also I removed usages of `{%halconfig-dir%}` to avoid having two different ways of specifying local files relative to hal config home. It's still used though to support restoring of backup files created with that prefix.

german-muzquiz:
@plumpy let me know if you have any feedback

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -68,6 +70,9 @@
 
   private boolean useBackup = false;
 
+  // @LocalFile fields whose file path is relative
+  private Map<Node, Map<Field, String>> originalRelativePaths = new ConcurrentHashMap<>();
```
plumpy: Two broken things here:
1) Adding state here is weird because a `@Component` is a singleton, so this is now essentially a global variable... everyone who injects a `HalconfigParser` now shares this map.
2) this data is specific to a given `Halconfig`, but you don't tie it to any specific `Halconfig`, you just apply it to whatever object is handed to you

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -414,53 +414,101 @@ public NodeDiff diff(Node other) {
...
+  private Path toRelativePath(String rootPath, String absolutePath) {
+    if (!absolutePath.startsWith(rootPath)) {
+      throw new HalException(
+          FATAL,
+          "Local file: " + absolutePath + " has incorrect prefix - must begin with " + rootPath);
+    }
+    if (!rootPath.endsWith(File.separator)) {
+      rootPath += File.separator;
+    }
+    return Paths.get(absolutePath.substring(rootPath.length()));
```
plumpy: `toRelativePath('/foo', '/foo2/bar')` returns `/bar`.

Don't try to do path resolution logic yourself; it's more complicated than you think. Convert these `Strings` to `Path` objects and use the built-in methods like `relativize`.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
german-muzquiz: Would it be better to move this back to `Node`? 
Essentially, paths are tied to specific `Node` instances (for example, a `Halconfig` may have multiple Kubernetes accounts, each one with its own individual `@LocalFile` for the kubeconfig), that's why I'm keying by `Node` instead of `Halconfig`.

plumpy: Sorry for the essay and also I 100% realize most of these problems existed before you so please don't think I'm picking on you.

But anyway: where should this code and this temporary data live?

Not in `Node` or `Halconfig`. These are data objects and by any modern coding standard, they should be immutable. They certainly shouldn't have transient data or methods that mutate them in seemingly arbitrary ways. And anyway, 99% of the users of `Halconfig` don't care one bit about transforming absolute paths, so it probably shouldn't live there.

Not in `HalconfigParser`. It's a singleton, so shouldn't have state, as mentioned. But also, it's right there in the name: it's a parser, there's no reason it should also have code to modify data.

Maybe in `BackupService`. This code and data exists _solely_ for use by backups. So this is the first logical place I see where it might live. Putting the code here definitely would make sense. But it's also a singleton, so shouldn't have state.

The state should ideally live in a `BackupScope` object or something. That's how we typically model something like this. Then you pass the scope into the `BackupService`. But if you do that, you'll find you have nowhere to store the scope between calls to `BackupService#create` and `BackupService#restore`. And that's your first clue why this implementation won't work: you have no guaranteee that someone didn't restart halyard between those two cals and lose your state! That's why `{%halconfig-dir%}` was there.

But I think you just don't need any of this anyway. What this is doing is making it so when you back up `/home/german/.hal/mykey.json`, it gets restored on my box as `/home/plumpy/.hal/mykey.json`. But if you're going to make relative paths work anyway, who cares if we just restore it instead to `mykey.json`. Why bother restoring the absolute path at all?

german-muzquiz: These are valid points. As for what's the purpose of this in-memory state, when we can just go away using relative paths in hal `config` file: there is some code, specially validators, that during the validation process try to load these files from file system with regular `IOUtils.toByteArray(new FileInputStream(path))`. This code doesn't work with relative file paths because it's not resolving relative to hal config home, so here I'm converting all paths to absolute when the file is loaded from disk. After changes are done to `Halconfig` structure and it's written back to disk, I need to restore the original relative paths.

So this flow of converting paths to absolute when `Halconfig` is loaded, and then write them back to disk as relative, is not just for doing `hal config backup/restore`, is for any code that wants to load a file to continue working, for example, any `hal config provider kubernetes` command that wants to load kubeconfig files.

I guess a better solution would be to address the root cause, and when any code tries to read a local file, append hal config home if the path is relative just in that moment. In this way there's no need for in-memory state or for path manipulation when loading or saving `Halconfig`. The downside I see is that this solution may be a bit fragile on maintenance, as anyone that is not aware that there's a helper function for loading files, could easily bypass it by doing plain `new File()` or something else. Anyway, it would be worth implementing.

german-muzquiz:
Apologies for changing again the implementation. The new approach for supporting relative files introduces these changes:
* Now there's no need for keeping an in-memory backup of file paths.
* File manipulation for staging backups (`hal config generate` and friends) was taken out of `Node` and moved to `BackupService`, so that anything related with backups lives in that service.
* There's a new `FileService` responsible for getting file paths and file contents taking into account if the reference is a config server resource, a secret or an absolute/relative local file. The idea is that `FileService` is the central place for getting file related information. I already found at least three different places with similar logic for reading a file reference from secrets, from config server or from local file system, so I unified them in `FileService`. This introduces the ability for any `@LocalFile` field to automatically support config server references. Previously it seemed that only `kubeconfig` and maybe other field supported this.

@plumpy let me know what you think about this approach.

AbdulRahmanAlHamali:
I would be a happy user of this feature. Waiting for it to land!

PR #1423, opened 2019-09-16
fix(stats): Replace UUID with ULID as spinnaker instance idententifier

ttomsu:
Fixes https://github.com/spinnaker/spinnaker/issues/4872

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
@@ -16,7 +16,7 @@
 
 package com.netflix.spinnaker.halyard.config.model.v1.node;
 
-import java.util.UUID;
+import de.huxhorn.sulky.ulid.ULID;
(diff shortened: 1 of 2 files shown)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
```diff
@@ -33,7 +33,7 @@ public String getNodeName() {
 
   private Boolean enabled = false;
   private String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
-  private String instanceId = UUID.randomUUID().toString();
+  private String instanceId = new ULID().toString();
```
ezimanyi: I'm pretty sure this is just serializing the `ULID` generator itself, and you need to to call `nextULID()` on the generator to get an actual ULID.

ttomsu: That's what I get for trying to move too fast. Fixed.

PR #1436, opened 2019-09-24
feat(kubernetes): add flag for Kubernetes custom resources

jwelch92:
Adds flag `--custom-resources` to specify CRDs that should be cached by clouddriver. Defining custom resources here is required for them to be used in patch and delete pipeline stages.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesAddAccountCommand.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesAddAccountCommand.java
@@ -119,6 +119,12 @@ public class KubernetesAddAccountCommand extends AbstractAddAccountCommand {
       description = KubernetesCommandProperties.CACHE_THREADS)
   private int cacheThreads = 1;
 
+  @Parameter(
+      names = "--custom-resources",
+      variableArity = true,
+      description = KubernetesCommandProperties.CUSTOM_RESOURCES)
+  public List<String> customResources = new ArrayList<>();
+
   @Override
   protected Account buildAccount(String accountName) {
     KubernetesAccount account = (KubernetesAccount) new KubernetesAccount().setName(accountName);
@@ -143,6 +149,14 @@ public class KubernetesAddAccountCommand extends AbstractAddAccountCommand {
     account.setCheckPermissionsOnStartup(checkPermissionsOnStartup);
     account.setLiveManifestCalls(liveManifestCalls);
     account.setCacheThreads(cacheThreads);
+    customResources.forEach(
+        customResource ->
+            account
+                .getCustomResources()
+                .add(
+                    new KubernetesAccount.CustomKubernetesResource()
+                        .setKubernetesKind(customResource)
+                        .setVersioned(false)));
     return account;
   }
 
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesCommandProperties.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesCommandProperties.java
@@ -69,4 +69,6 @@ public class KubernetesCommandProperties {
   static final String CACHE_THREADS =
       "Number of caching agents for this kubernetes account. Each agent handles a subset of the namespaces available to this account. "
           + "By default, only 1 agent caches all kinds for all namespaces in the account.";
+  static final String CUSTOM_RESOURCES =
+      "(V2 Only) List of Kubernetes Custom Resources to be cached by clouddriver and available for use in patch and delete pipeline stages.";
 }
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -8507,6 +8507,7 @@ during clouddriver startup when you have many kubernetes accounts configured. Th
  * `--configure-image-pull-secrets`: (*Default*: `true`) (Only applicable to the v1 provider). When true, Spinnaker will create & manage your image pull secrets for you; when false, you will have to create and attach them to your pod specs by hand.
  * `--context`: The kubernetes context to be managed by Spinnaker. See [http://kubernetes.io/docs/user-guide/kubeconfig-file/#context](http://kubernetes.io/docs/user-guide/kubeconfig-file/#context) for more information.
 When no context is configured for an account the 'current-context' in your kubeconfig is assumed.
+ * `--custom-resources`: (*Default*: `[]`) (V2 Only) List of Kubernetes Custom Resources to be cached by clouddriver and available for use in patch and delete pipeline stages.
  * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
  * `--docker-registries`: (*Default*: `[]`) A list of the Spinnaker docker registry account names this Spinnaker account can use as image sources. These docker registry accounts must be registered in your halconfig before you can add them here.
  * `--environment`: The environment name for the account. Many accounts can share the same environment (e.g. dev, test, prod)


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesAddAccountCommand.java
```diff
@@ -143,6 +149,14 @@ protected Account buildAccount(String accountName) {
...
     account.setLiveManifestCalls(liveManifestCalls);
     account.setCacheThreads(cacheThreads);
+    customResources.forEach(
+        customResource ->
+            account
+                .getCustomResources()
+                .add(
+                    new KubernetesAccount.CustomKubernetesResource()
+                        .setKubernetesKind(customResource)
+                        .setVersioned(false)));
```
jwelch92: Was unable to verify this property on the custom resources is actually used anywhere. Need guidance on how to properly setup the command for accepting a list of custom resources _and_ whether they are versioned.

dotdotdotpaul: Hmm, I'm not entirely sure where that's done, either.  But the rest of the code looks fine.  😄

thecodejunkie:
This is awesome! 👍

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/kubernetes/KubernetesAccountValidator.java
```diff
@@ -111,6 +111,17 @@ private void validateKindConfig(ConfigProblemSetBuilder psBuilder, KubernetesAcc
...
     }
 
+    if (CollectionUtils.isNotEmpty(customResources)) {
+      List<String> kubernetesKindNotSet =
+          customResources.stream()
+              .map(KubernetesAccount.CustomKubernetesResource::getKubernetesKind)
+              .filter(cr -> (cr == null || cr.isEmpty()))
+              .collect(Collectors.toList());
+      if (CollectionUtils.isNotEmpty(kubernetesKindNotSet)) {
+        psBuilder.addProblem(ERROR, "Custom resources must have \"kubernetesKind\" set.");
```
german-muzquiz: There's no indication to users on which is the `kubernetesKind` field. For example for a command like this:
```
hal config provider kubernetes account edit ACCOUNT --add-custom-resource "spinnakerKind:instances;versioned:true"
```
A better error message would be something like "missing custom resource name".

ethanfrogers:
@ezimanyi can you take a look at this and let us know what you think? Specifically, this bits about specifying the custom resource via the CLI.

jwelch92:
@ezimanyi Yeah I'm not crazy about the format but like you mentioned it doesn't match other commands and I had trouble figuring out a pattern that would work. 

I'll take a look at that command and see if I can rethink this to work more like that one. How you would envision the command looking when adding a new kubernetes account? Currently both the `add` and `edit` subcommands for account support adding N custom resources at the same time. 

It does appear to support reading these fields from the main halconfig. Our use case is making it easier to automate adding kubernetes accounts with potentially a lot of CRDs defined. Editing the yaml directly works but so far has proved to be more difficult to automate remotely.

ezimanyi:
That's a good question about how it would look for the `add` command.  Would it be an option to require a follow-up `edit` command if you want to add custom resources?

Thanks also for the context around why you're using the CLI vs editing the YAML!

thecodejunkie:
@ezimanyi another good reason for wanting to use the CLI, instead of manually editing the YAML, is if you deploy Halyard using the (official) pod. In that pod you don't have access to an editor and you can't sudo to install one.

jwelch92:
@ezimanyi Yeah I think that's an option, I'll wire that up based on your suggestions and see how it looks. Appreciate the feedback!

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesCommandProperties.java
```diff
@@ -69,4 +69,8 @@
   static final String CACHE_THREADS =
       "Number of caching agents for this kubernetes account. Each agent handles a subset of the namespaces available to this account. "
           + "By default, only 1 agent caches all kinds for all namespaces in the account.";
+  static final String CUSTOM_RESOURCES =
+      "(V2 Only) List of Kubernetes Custom Resources to be cached by clouddriver and available for use in patch and delete pipeline stages. "
```
ezimanyi:
```suggestion
      "(V2 Only) List of Kubernetes Custom Resources to be cached by clouddriver and available for use in patch and delete manifest stages. "
```

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/kubernetes/KubernetesCommandProperties.java
```diff
@@ -69,4 +69,8 @@
...
+          + "Fields besides the Kubernetes Kind can be set using the following "
```
ezimanyi: I think these next two lines need to be updated now that we have `--spinnaker-kind` and `--versioned` flags

jwelch92:
Thanks for all the feedback and getting it merged @ezimanyi !

shubhamc183:
@jwelch92 can we add a list(array) of CRD at once?

--add-custom-resource=A,B,C or --add-custom-resource=[A,B,C]

jwelch92:
> @jwelch92 can we add a list(array) of CRD at once?
> 
> --add-custom-resource=A,B,C or --add-custom-resource=[A,B,C]

It's been awhile but after a quick look I'm pretty sure it can only handle one at a time. You can alternatively just manually edit your halconfig file to add the CRDs.

shubhamc183:
@jwelch92 thanks a lot for the reply.

Do you plan to implement array-based addition any time in the future?

jwelch92:
@shubhamc183 Nope, never going to work on Spinnaker code again. I wrote this for work while at my previous company which no longer uses Spinnaker and I've since moved on to a new role.  

The maintainers are pretty good about reviews so I would encourage you to contribute that feature if you can. Best of luck!

Kyslik:
@jwelch92 

> It's been awhile but after a quick look I'm pretty sure it can only handle one at a time. You can alternatively just manually edit your halconfig file to add the CRDs.

I find it really confusing; can a real life example be added _anywhere_ in the documentation or even here?
What is `SpinnakerKind`?

```
customResources:
  - kubernetesKind: "Kind" # serviceentries.networking.istio.io
    spinnakerKind: "SpinnakerKind" # no clue what SpinnakerKind is there is no documentation on it anywhere 🤷🏻 
    versioned: false
```

Thank you for this feature :)

german-muzquiz:
Adding here for reference what's the meaning of `SpinnakerKind`:

Spinnaker has concepts like `ServerGroups` or `Instances` that are common to all cloud providers. These concepts are described here: https://spinnaker.io/concepts/#application

In the case of Kubernetes, core types are mapped to Spinnaker concepts in the code, so for example a Kubernetes `Ingress` is a Spinnaker `LoadBalancer`. This is used in places like the UI where you can see the Load Balancers.

In the case of CRDs Spinnaker has no way to know its corresponding Spinnaker kind, so by default all CRDs are assigned the generic Spinnaker kind `Unclassified`. When you configure `customResources`, you're essentially telling Spinnaker what spinnaker kind corresponds to the CRD. The list of valid Spinnaker kinds can be seen here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/SpinnakerKind.java#L23

PR #1437, opened 2019-09-26
Generalized init container

link108:
In order to provide a UI component for plugins, we need to place the required resources such that Deck has access to them. In a Kubernetes environment, this means the pod needs access to the necessary assets. Ideally those assets are available locally and Deck itself would not have to download them. Some Spinnaker operators may choose to build images with the plugins included, so the deployment of plugin resources should be decoupled with the running service itself. 

To solve this for Kubernetes environments, I am using init containers to download the required 
resources to a volume mount thats shared with the pod, in this case deck. 

I had previously solved this issue for Java services by having Kork download the resources upon startup, however I would like to move to this version of deploying plugin resources to decouple the deployment of resources from the loading of plugins. This also simplifies the deployment of plugin resources by providing a single method of deployment for both Java services and Deck.

This does change how a plugin’s manifest.yml is formatted, changing from:

```
    ...
    jars:
      - resourceA.jar
      - resourceB.jar
```
to

```
    ...
    resources:
      orca:
        - resourceA.jar
      deck:
        - resourceB.js
        - resourceC.png
```
Since the plugin manifest format was changed to allow for specifying which service each resources should be laid down upon, PluginProfileFactory was also updated to allow for specifying which service the plugins.yml file is needed for. 

There will be an accompanying Deck PR that updates settings.js to read the plugin binding. I’ve added another format for replacing keys in settings.js so we can replace values with objects. Previously, we were only able to replace keys as strings, for instance given a variable `foo="bar"` `'{%foo%}'` would be replaced with `'bar'`. The single quotes are necessary for having a valid JSON, but it ensures only strings are templatizable. This PR adds the ability to create objects for consumption by Deck by adding another templating format (double curly braces + percent sign), `'{{%key%}}'`. In this case, the single quotes are still required, however we replace them as well, which allows us to create valid JSON objects. 

I’ve also added manifests as a field to Plugin.java, since a manifest.yml belongs to a particular plugin. This also allows for only downloading the manifest once, rather than each time it is required.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/plugins/Plugin.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/plugins/Plugin.java
@@ -27,8 +27,10 @@ import java.net.URL;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import lombok.AccessLevel;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.representer.Representer;
@@ -36,23 +38,29 @@ import org.yaml.snakeyaml.representer.Representer;
 @Data
 @EqualsAndHashCode(callSuper = true)
 public class Plugin extends Node {
-  public String name;
-  public Boolean enabled;
-  public String manifestLocation;
-  public Map<String, Object> options = new HashMap<>();
+  private String name;
+  private Boolean enabled;
+  private String manifestLocation;
+  private Map<String, Object> options = new HashMap<>();
+
+  @Getter(AccessLevel.NONE)
+  private Manifest manifest;
 
   @Override
   public String getNodeName() {
     return name;
   }
 
-  public Manifest generateManifest() {
+  public Manifest getManifest() {
+    if (manifest != null) {
+      return manifest;
+    }
     Representer representer = new Representer();
     representer.getPropertyUtils().setSkipMissingProperties(true);
     Yaml yaml = new Yaml(new Constructor(Manifest.class), representer);
 
     InputStream manifestContents = downloadManifest();
-    Manifest manifest = yaml.load(manifestContents);
+    manifest = yaml.load(manifestContents);
     manifest.validate();
     return manifest;
   }
@@ -113,6 +121,6 @@ public class Plugin extends Node {
   }
 
   public Map<String, Object> getCombinedOptions() {
-    return Plugin.merge(generateManifest().getOptions(), options);
+    return Plugin.merge(getManifest().getOptions(), options);
   }
 }
--- a/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/resource/v1/ObjectReplaceTemplatedResource.java
+++ b/halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/resource/v1/ObjectReplaceTemplatedResource.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 Armory, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ *
+ */
+
+package com.netflix.spinnaker.halyard.core.resource.v1;
+
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ObjectReplaceTemplatedResource extends TemplatedResource {
+  public static Logger log = LoggerFactory.getLogger(ObjectReplaceTemplatedResource.class);
+
+  protected String formatKey(String key) {
+    return "'{{%" + key + "%}}'";
+  }
+
+  protected String getRegexSafeFormatKey() {
+    return "(?s).*\\'\\{\\{%.*%\\}\\}\\'.*";
+  }
+
+  @Override
+  public String toString() {
+    String contents = getContents();
+    for (Map.Entry<String, Object> binding : bindings.entrySet()) {
+      Object value = binding.getValue();
+      contents =
+          contents.replace(formatKey(binding.getKey()), value != null ? value.toString() : "");
+    }
+    if (contents.matches(getRegexSafeFormatKey())) {
+      log.warn(
+          "Found part of template that still contains a format key, likely a missing template value for a key, template: "
+              + contents);
+    }
+    return contents;
+  }
+}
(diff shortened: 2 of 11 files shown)


link108:
cc @ezimanyi @robzienert

robzienert:
@christopherthielan

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/plugins/Plugin.java
```diff
@@ -37,18 +39,24 @@
...
+  @Getter(AccessLevel.NONE)
+  public Manifest manifest;
+
   @Override
   public String getNodeName() {
     return name;
   }
 
-  public Manifest generateManifest() {
+  public Manifest getManifest() {
```
ezimanyi: It seems dangerous that we're caching the result of `getManifest` on the object when the object isn't immutable.  What if `manifestLocation` is updated, then we call `getManifest()`---we'll get an incorrect result.  Can the properties that determine the manifest be made immutable?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/plugins/Plugin.java
```diff
@@ -37,18 +39,24 @@
   public Boolean enabled;
   public String manifestLocation;
 
+  @Getter(AccessLevel.NONE)
+  public Manifest manifest;
```
ezimanyi: Why is this `public` (and why are we generating an accessor with `AccessLevel.NONE`)?

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/resource/v1/ObjectReplaceTemplatedResource.java
```diff
@@ -0,0 +1,51 @@
...
+package com.netflix.spinnaker.halyard.core.resource.v1;
+
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ObjectReplaceTemplatedResource extends TemplatedResource {
+  public static Logger log = LoggerFactory.getLogger(ObjectReplaceTemplatedResource.class);
+
+  protected String formatKey(String key) {
```
ezimanyi: I think this can be `private`

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/resource/v1/ObjectReplaceTemplatedResource.java
```diff
@@ -0,0 +1,51 @@
...
+    return "'{{%" + key + "%}}'";
+  }
+
+  protected String getRegexSafeFormatKey() {
```
ezimanyi: I think this can be `private`

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/resource/v1/ObjectResource.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+
+public class ObjectResource extends ObjectReplaceTemplatedResource {
+  public ObjectResource(String contents) {
+    this.contents = contents;
+  }
+
+  @Getter(AccessLevel.PROTECTED)
+  private String contents;
```
ezimanyi: Can this be `final`?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
```diff
@@ -232,8 +234,35 @@ protected void setProfile(
       bindings.put("canary.showAllCanaryConfigs", canary.isShowAllConfigsEnabled());
     }
 
+    // Configure Plugins
+    List<String> pluginBinding = new ArrayList<>();
+    List<Plugin> plugins = deploymentConfiguration.getPlugins().getPlugins();
+    Map<String, List<String>> pluginMetadata =
+        plugins.stream()
+            .filter(p -> p.getEnabled())
```
ezimanyi: Can be replaced with just `.filter(Plugin::getEnabled)`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
```diff
@@ -232,8 +234,35 @@ protected void setProfile(
...
+            .filter(p -> !p.getManifestLocation().isEmpty())
+            .map(p -> p.getManifest())
```
ezimanyi: Can be replaced with just `.filter(Plugin::getManifest)`

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
```diff
@@ -232,8 +234,35 @@ protected void setProfile(
...
+            .filter(m -> m.getResources().containsKey(SpinnakerArtifact.DECK.getName()))
+            .collect(
+                Collectors.toMap(
+                    m -> m.getName(), m -> m.getResources().get(SpinnakerArtifact.DECK.getName())));
```
ezimanyi: `m -> m.getName()` can be replaced with just `Manifest::getName`

link108:
@ezimanyi I don't have a design doc, but we do have a PR: https://github.com/spinnaker/deck/pull/7474. The description explains how plugins are loaded by deck. With this PR, Halyard lays down resources prior to the deck container starting, using an init container and a shared volume mount. After the init container downloads the resources, the volume mount is is then mounted to /opt/deck/html/plugins/, which deck is able to serve. I'd like to move away from having services download the required resources and using halyard to manage that, as laying down resources is a deployment concern. I'd like to do the same for Spinnaker java services as well, the plugins.yml that is currently laid down and contains URLs for plugin resources will be updated to contain local paths only. 

The key we're adding to the deck profile is for replacing the new template format I'm adding here. The PR I linked above shows where this is being used on the deck side: https://github.com/spinnaker/deck/pull/7474/files#diff-7a96434a7dcdc316a33ad4a5b0323f64. The format that it expects is here: https://github.com/spinnaker/deck/pull/7474/files#diff-03cccc89a134725d201bda499c22928bR135 (list of objects containing plugin name and its location on deck). 

In DeckProfileFactory, the profile.getBaseContents() function downloads the settings.js template, which is then hydrated. The new key format, {{%key%}}, is replaced with the object key, prior to {%key%} being replaced by the value of key. This templating is done in the toString function in both StringResource and ObjectResource. 

RFC is here: https://docs.google.com/document/d/1vry9JvkjdxhqfvD1CNnET-j3cA6a4gNYKEPFOLNCO4k/edit?usp=sharing
Do I need to create a Spinnaker issue as well?

ezimanyi:
Both this PR as well as spinnaker/deck#7474 are making some pretty significant architectural decisions that I think should be discussed more broadly.  The [initial RFC](https://github.com/spinnaker/spinnaker/issues/4181) only has a very quick sentence on how UI support for plugins will work.  If we're at the point where we're ready to implement that UI support, I'd like to see either an update to the RFC or a new RFC to explain how that will work and get broader visibility/approval.

One thing I'd like to see in that RFC is alternatives to extending the custom templating that Halyard does in `settings.js`.  I don't think that having Halyard write its own custom templating language to generate `settings.js` was a good architectural decision, and I'm wary of further extending it now to be able to print out arbitrary javascript objects.  What alternatives were considered here?

link108:
Closing this PR until we figure out how plugins will be deployed. Front50 seems like it will be the way to go, so will most likely not be re-opening this. Thank you for all the feedback!

PR #1439, opened 2019-10-01
feat(plugins): enable plugin config overrides

link108:
Spinnaker operators should be able to override default plugin config values. This PR enables Spinnaker operators to configure plugin overrides via the hal config file. 

A hal config could contain the following in order to override the options for a particular plugin:
```
  plugins:
    plugins:
    - name: test
      enabled: true
      manifestLocation: https://github.com/armory/stage-plugin/blob/master/manifest.yml
      options:
        example:
          url: blah
          port: 2000
          nested:
            list:
            - last
            - first
```

I am also correcting the copy+paste error for Manifest.java's copyright.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
@@ -24,6 +24,7 @@ import com.netflix.spinnaker.halyard.config.model.v1.providers.aws.AwsProvider;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSettings;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.integrations.IntegrationsConfigWrapper;
+import java.util.AbstractMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


link108:
@spinnakerbot assign-issue @ezimanyi

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
```diff
@@ -80,14 +81,18 @@ protected void setProfile(
     // For backward compatibility
     profile.appendContents("pipelineTemplate.enabled: " + pipelineTemplates);
 
+    // Plugins
```
ezimanyi: Comments like this generally make me think the logic that follows them should be extracted to a function.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/OrcaProfileFactory.java
```diff
@@ -80,14 +81,18 @@ protected void setProfile(
...
     final List<Plugin> plugins = deploymentConfiguration.getPlugins().getPlugins();
     Map<String, Object> fullyRenderedYaml = new LinkedHashMap<>();
     Map<String, Object> pluginMetadata =
         plugins.stream()
             .filter(p -> p.getEnabled())
             .filter(p -> !p.getManifestLocation().isEmpty())
-            .map(p -> p.generateManifest())
-            .collect(Collectors.toMap(m -> m.getName(), m -> m.getOptions()));
+            .map(p -> new AbstractMap.SimpleEntry<>(p, p.generateManifest()))
```
ezimanyi: This construct is somewhat confusing to me---we're creating a temporary `AbstractMap.SimpleEntry` in the stream, but then are collecting to a map that doesn't actually use these entries.  (The indirection also makes it hard to see what we're calling `getOptions()` on below as it's not immediately clear what the value and key are.)

Given that the `Plugin` class has access to all the data it needs to compute the merged options, it might be more clear to have it expose a function to return these merged options (maybe even just doing this in `generateManifest()`) instead of requiring callers to send back data through a static method.

link108: ah thats a good call, moving 👍

PR #1443, opened 2019-10-02
fix(halyard): Change deployment to support new Kubernetes API (…

justinrlee:
…1beta2 has been deprecated and replaced with apps/v1)

As of Kubernetes 1.16, the apps/v1beta2 deployment object has been deprecated.

I'm assuming the commit will be squashed, so I'm ignoring the typo in the commit message.

### DIFF
--- a/halyard-deploy/src/main/resources/kubernetes/manifests/clouddriver.yml
+++ b/halyard-deploy/src/main/resources/kubernetes/manifests/clouddriver.yml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1beta2 # for versions before 1.9.0 use apps/v1beta2
+apiVersion: apps/v1 # As of Kubernetes 1.16, apps/v1beta2 has been deprecated
 kind: Deployment
 metadata:
   name: spin-clouddriver
--- a/halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
+++ b/halyard-deploy/src/main/resources/kubernetes/manifests/deployment.yml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1beta2 # for versions before 1.9.0 use apps/v1beta2
+apiVersion: apps/v1 # As of Kubernetes 1.16, apps/v1beta2 has been deprecated
 kind: Deployment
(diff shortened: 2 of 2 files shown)


justinrlee:
Addresses https://github.com/spinnaker/spinnaker/issues/4995.

I think.

ezimanyi:
@justinrlee The changes look good, but I'd like to dig a bit into the `I think.` part above---when you tested these changes, were you able to deploy successfully to a Kubernetes 1.16 cluster.

Also, what will this mean for users on Kubernetes <=1.9.  I presume this will break deploying to those clusters?  If so, that's probably reasonable but we should probably make that really clear in the commit message and release notes.

justinrlee:
Hey Eric - I need to do some testing on this - I'll try to do testing this weekend.

I agree - I think this would break Kubernetes <1.9, which I think is deprecated?

ethanfrogers:
I think we decided that <1.9 won't be suppported.

ezimanyi:
I'm fine with not supporting <1.9 as long as we make this clear in the release notes.

justinrlee:
Hello!

Okay, so GKE/EKS aren't on 1.16, and neither is K3s (my test environment of choice).  But I stood up a microk8s instance using Kubernetes 1.16, and verified the following:
* Current versions of Halyard weren't able to successfully deploy Spinnaker to it (`error: unable to recognize "STDIN": no matches for kind "Deployment" in version "apps/v1beta2"`)
* A Halyard image built based on this branch *Is* able to successfully deploy Spinnaker to it.
* In addition, a Halyard image built based on this branch *is* able to successfully to deploy Spinnaker to a Kubernetes 1.13 (EKS) cluster.

Happy to do additional testing if you'd like me to @ezimanyi @ethanfrogers.  Sorry for the delay on this.

Also, I don't know how Halyard versions are determined (when we increment x vs. y vs. z in our x.y.z version numbering), but I see this as a potentially breaking change for users on old versions of Kubernetes.  I agree we need to make it clear in the release notes, but should we do more or is that sufficient?

Thanks!

PR #1444, opened 2019-10-03
fix(plugins): plugins should be able to reference their own config values

link108:
Currently, plugin configs are laid down on Orca in the following format:

```
plugins:
  userSuppliedPluginName:
    foo: bar
```

This causes problems for plugin developers who want to reference their own config values. While a plugin can access config values, the plugin developer does not know what key to reference, since the plugin name is user supplied.

This PR moves to a new format:
```
plugins:
  PluginName/fromManifest:
    foo: bar
```

The key used to reference the plugin is the plugin name that resides in the plugin's manifest.yml.


link108:
While we're figuring out next steps for plugins, I'd like to merge this PR so early users of plugins have less issues (this is a bugfix vs added functionality)

link108:
@spinnakerbot assign-issue @ezimanyi @german-muzquiz

PR #1446, opened 2019-10-09
feat(slack): Allow configurable slack endpoint

diasjorge:


diasjorge:
After https://github.com/spinnaker/echo/pull/633 has been merged, I'd like to add support for setting the configuration in halyard.

If someone could please point to me how could I test this locally. I don't really understand how those settings are transformed into the correct yaml structure, so I tried to follow what was in place.

docs/commands.md
```diff
@@ -6579,8 +6579,10 @@ hal config notification slack edit [parameters]
 ```
 
 #### Parameters
+ * `--base-url`: Slack endpoint.
```
german-muzquiz: Can you indicate that this parameter is optional?

diasjorge: Is the comment good like this? I accept other suggestions :)

PR #1449, opened 2019-10-17
fix(kubernetes): remove user-facing references to todo(lwander)

lwander:


duftler:
Looks like the formatting needs to be fixed.

PR #1454, opened 2019-10-25
feat(config): HalconfigDirStruct source of truth, allow override

ncknt:
Halyard uses `halyard.halconfig.directory` for both `halconfigDirectory` and `halconfigPath`. With this PR:
- Halyard only uses `HalconfigDirectoryStructure` to build the correct path.
- We can override the directory used. The goal is to be able to temporarily point Halyard to a separate directory to read the user's configuration.

Note after that `localBomPath` still uses the same property but it's in `halyard-core` so without an easy way to refactor.

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
@@ -36,7 +36,6 @@ import java.io.InputStream;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -61,11 +60,6 @@ import org.yaml.snakeyaml.scanner.ScannerException;
 @Slf4j
 @Component
 public class HalconfigParser {
-
-  @Autowired String halconfigPath;
-
-  @Autowired String halyardVersion;
-
   @Autowired StrictObjectMapper objectMapper;
 
   @Autowired HalconfigDirectoryStructure halconfigDirectoryStructure;
@@ -73,7 +67,6 @@ public class HalconfigParser {
   @Autowired Yaml yamlParser;
 
   private boolean useBackup = false;
-  private String backupHalconfigPath;
 
   /**
    * Parse Halyard's config.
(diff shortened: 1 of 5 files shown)


ncknt:
@german-muzquiz Can you PTAL?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
```diff
@@ -104,8 +100,11 @@ public Halconfig setInmemoryHalConfig(ByteArrayInputStream is) throws IllegalArg
   }
 
   private InputStream getHalconfigStream() throws FileNotFoundException {
-    String path = useBackup ? backupHalconfigPath : halconfigPath;
-    return new FileInputStream(new File(path));
+    String baseDirectory =
+        useBackup
+            ? halconfigDirectoryStructure.getBackupConfigPath().toString()
+            : halconfigDirectoryStructure.getHalconfigDirectory();
+    return new FileInputStream(new File(baseDirectory, "config"));
```
german-muzquiz: `halconfigDirectoryStructure.getBackupConfigPath()` already returns the path including `config` file, leading to a double "config" in case of backups: https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigDirectoryStructure.java#L109

ncknt: Good catch! Updated.

ncknt:
@ezimanyi Could you PTAL?

plumpy:
Can you explain how this will be used? A `ThreadLocal` is an incredibly fragile way to do anything, and I can't see why you need to use it here...

(I was about to put up a PR to delete this code because it's unused and very unusual, but I just noticed it was merged yesterday.)

ncknt:
This is to be used in the context of the Kubernetes operator for Spinnaker. The operator itself sends the config to Halyard to generate the manifests (but not apply them) via a multipart request. The manifest generation is single threaded and the override cleaned up after use.

plumpy:
Can you show me what the calling code looks like and maybe I can help come up with a more idiomatic solution? Like maybe there's a `ParsingScope` object that contains a `HalconfigDirectoryStructure` and gets passed down or something.

While I certainly understand the inclination to just write the least amount of code that will make this work (especially in a codebase like halyard that's already so ugly), jamming config data through using a `ThreadLocal` dramatically breaks encapsulation and is definitely an anti-pattern and this might be a good opportunity to clean things up a little...

ncknt:
Yeah `ThreadLocal` is a cheap way to achieve the result. It's along the same lines as another existing thread local (`DaemonTaskHandler.getContext()` used to cache the parsed config). In this very case, the call to override the directory wraps the processing of the request which is thread bound.

I'm open to alternatives but they will come at a price, either make `DeployService`, `HalconfigParser`, `GenerateService` and quite a few others non Spring components or pass the scope/context everywhere.

ezimanyi:
I would definitely advocate for finding an alternative here...while it might be more effort, I think that effort should be considered part of the cost of implementing this feature.  It's really not great to have the correctness of the code be tightly coupled to the threading model, as this change is causing to happen.

ncknt:
Ok I'll give it a try. Wish me luck :)

PR #1455, opened 2019-10-28
fix(saml): get saml file path instead of file contents in saml validator

KathrynLewis:
The SAML validator currently is getting the file contents of the metadata file instead of the file path, which causes the validator to fail. This PR fixes that bug.


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/security/SamlValidator.java
```diff
@@ -30,10 +31,13 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 @Component
 public class SamlValidator extends Validator<Saml> {
+  @Autowired SecretSessionManager secretSessionManager;
```
german-muzquiz: This is already inherited from `Validator`, you don't need to define it here.

KathrynLewis: Ah, thanks! Updated.

KathrynLewis: Ah, thanks! Updated.

PR #1457, opened 2019-10-30
fix(halyard/macos): Catalina compability

jervi:
Halyard will not start on macOS Catalina because it can't write logs to `/var`. Changing the log directory to the user directory fixes the problem.

### DIFF
--- a/startup/macos/hal
+++ b/startup/macos/hal
@@ -12,8 +12,8 @@ function start_daemon() {
     printf "The halyard daemon isn't running yet... starting it manually"
 
     /opt/halyard/bin/halyard \
-      2> /var/log/spinnaker/halyard/halyard.err \
-      > /var/log/spinnaker/halyard/halyard.log &
+      2> ~/dev/spinnaker/logs/halyard/halyard.err \
+      > ~/dev/spinnaker/logs/halyard/halyard.log &
 
     echo $! > $HALYARD_DAEMON_PID_FILE 
 


german-muzquiz:
Did Catalina upgrade changed ownership of `/var/log/spinnaker` or is there another reason that Halyard cannot write to that dir? 
Logs folder reference is not just in hal startup script, there are other places in the code where we would need to change it that affect all operating systems: https://github.com/spinnaker/halyard/blob/master/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/LocalLogCollectorFactory.java#L62.

jervi:
No, the upgrade cleaned up the entire `/var` directory and moved everything "foreign" there into a "Relocated Items" folder in `/Users/Shared`. So there is no `/var/log/spinnaker` folder anymore. I can recreate it though.

I'm not arguing that the solution in this PR is the best solution, but maybe we could check in the startup script if the folder exist and if not, recreate it?

german-muzquiz:
Creating the folder at startup if it doesn't exist sounds good. @ezimanyi thoughts?

jervi:
On second thought, then we would need to run halyard as root. Maybe just a check and a better error message would suffice

ethanfrogers:
FWIW, looks like this affects anyone doing any local development on Catalina, we well. I was testing another PR and, when running a `hal config` command, saw errors about not being able to save a temporary config file.

german-muzquiz:
How about checking if the directories exist and if not, create them and set permissions with `sudo`, so that `hal` command would automatically ask for root password only in that case, something like this:
```
if [ ! -d "/var/log/spinnaker" ]
then
  sudo mkdir -p /var/log/spinnaker/halyard
  ....
fi
```

ezimanyi:
I think the solution in the last comment above (checking if it exists, and creating it if not) makes sense.

PR #1458, opened 2019-11-01
feat(artifacts): add git repo artifact support

KathrynLewis:
Enables configuration through Halyard for the git repo artifact type added here: https://github.com/spinnaker/clouddriver/pull/4049

@ethanfrogers


ethanfrogers:
I'm having a bit of trouble testing this locally (stupid Catalina) but will give it another go later today. One thing that I want to be sure about - does the config key show up as `gitRepo` or `gitrepo`? The config value is `gitRepo` but I'm not sure if it's case sensitive or not.

KathrynLewis:
It shows up as `gitrepo` but I've tested end-to-end and the account shows up as expected in the UI and the bake manifest stage successfully fetches the repo. But no rush on my end, so definitely test it out locally when you get a chance!

From `~/.hal/default/staging/clouddriver.yml`:
Screen Shot 2019-11-01 at 9 07 15 AM

PR #1462, opened 2019-11-05
fix(config): fix anonymous Google storage credentials

ncknt:
This should fix https://github.com/spinnaker/spinnaker/issues/5089.

The anonymous Google storage credentials using `com.google.auth.oauth2.GoogleCredentials.newBuilder().build()` are breaking on `refreshAccessToken` - probably after the bump to version `v1-rev267-1.25.0`.

I'm not a GCS specialist so I'd appreciate additional insights: @ezimanyi @plumpy


ezimanyi:
Looks like this was broken in #1452 so will let @duftler take a look to make sure this makes sense to him.

PR #1463, opened 2019-11-07
fix(backup): support wildcard truststore: skip backup

ncknt:
This skips the backup of `*` files as found in Igor's `truststore`: https://github.com/spinnaker/igor/pull/363

This should fix https://github.com/spinnaker/spinnaker/issues/5096

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
@@ -486,7 +486,7 @@ public abstract class Node implements Validatable {
                               }
                             }
 
-                            if (fPath == null) {
+                            if (fPath == null || fPath.equals("*")) {
                               return null;
                             }
 


ncknt:
@german-muzquiz Can you PTAL?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Node.java
```diff
@@ -486,7 +486,7 @@ public void makeLocalFilesAbsolute(String halconfigPath) {
                               }
                             }
 
-                            if (fPath == null) {
```
german-muzquiz: This fixes the immediate need, however I think we have a deeper problem because `trustStore` _sometimes_ uses a `@LocalFile` and `@SecretFile` field for something that is not a file reference (this star special value). This means potential bugs everywhere files are processed, like here. I don't see an easy solution without having to separate this special value to its own property in the configuration, thereby impacting also Igor.

Can you verify if the following doesn't break having this star value for trustStore?
1. Backups: `hal backup create` and then `hal backup restore...`
2. Any "mutating" command from CLI, for example change the value of a feature flag.

The above perform some file operations on any `@LocalFile` field, besides the actual `hal deploy apply`, which I assume is working.

kevinawoo:
Hey all, this is a pretty old PR, so I'm going play janitor 🧹 and just close it for now. 

Feel free to create a new PR and we'll take a look again with a new set of eyes 👀 !

PR #1474, opened 2019-12-04
feat(kubernetes): Add ServiceAccount provisioning support

relu:
This is a feature proposal for provisioning ServiceAccount resources for each deployment type through halyard rather than relying on user provisioned ServiceAccounts via the serviceAccountName setting.

On top of that, having the ability to define ServiceAccount annotations enables us to make use of AWS EKS' [OIDC authentication model](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html).


relu:
Any chance to get some attention to this? Since https://github.com/spinnaker/spinnaker/issues/5378 was merged, Spinnaker running on EKS should now be able to make use of OIDC enabled roles using service accounts.

kevinawoo:
@relu would you be able to fix the merge conflicts and we can take another look? 🙏

PR #1476, opened 2019-12-05
feat(huaweicloud): add provider of huaweicloud

zengchen1024:


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/huaweicloud/HuaweiCloudBakeryDefaultsValidator.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Override
+  public void validate(ConfigProblemSetBuilder p, HuaweiCloudBakeryDefaults n) {
+    DaemonTaskHandler.message(
+        "Validating "
+            + n.getNodeName()
+            + " with "
+            + HuaweiCloudBakeryDefaultsValidator.class.getSimpleName());
+
+    if ((new HuaweiCloudBakeryDefaults()).equals(n)) {
```
ezimanyi: I'm not sure I fully understand how this new commit fixes the bug.  Can you explain the logic behind short-circuiting here?  Thanks!

zengchen1024: Like google's [implementation](https://github.com/spinnaker/halyard/blob/master/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/google/GoogleBakeryDefaultsValidator.java#L48-L59), it checks whether the input is empty before doing validations. Here I did the same work, but I use `==` instead of `equals` to compare the input with an empty instance of HuaweiCloudBakeryDefaults wrongly.

ezimanyi: Ok, this looks good to me then, thanks for the explanation!

PR #1479, opened 2019-12-09
feat(stats): Adds Telemetry validator, which prints a new Info level …

ttomsu:
…message that will get printed during all `hal config telemetry enable|disable` and `hal deploy apply` invocations. Samples below:

```
$ hal config telemetry enable
+ Get current deployment
  Success
+ Enable or disable telemetry
  Success
Validation in default.telemetry:
- INFO Telemetry is currently ENABLED. Usage statistics are being
  collected—Thank you!These stats inform improvements to the product, and that
  helps the community.To disable, run `hal config telemetry disable`.To learn more
  about what and how telemetry data is used, please see
  https://www.spinnaker.io/community/stats

+ Successfully enabled telemetry settings.

```

```
$ hal config telemetry disable
+ Get current deployment
  Success
+ Enable or disable telemetry
  Success
Validation in default.telemetry:
- INFO Telemetry is currently DISABLED. Usage statistics are not
  being collected. Please consider enabling statistic collection. These stats
  inform improvements to the product, and that helps the community. To enable, run
  `hal config telemetry enable`. To learn more about what and how telemetry data
  is used, please see https://www.spinnaker.io/community/stats.

+ Successfully disabled telemetry settings.

```

I realize this kind of repurposes the `Problem` and `ProblemSet` classes to print an info message, but the infrastructure is super useful for this kind of task, i.e. only print a message during specific commands. Renaming all instances of Problem and ProblemSet would blow up this PR and mask what is actually trying to be accomplished.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/ResponseUnwrapper.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/ResponseUnwrapper.java
@@ -235,6 +235,9 @@ public class ResponseUnwrapper {
         List<String> options = problem.getOptions();
 
         switch (severity) {
+          case INFO:
+            AnsiUi.info(message);
+            break;
           case FATAL:
           case ERROR:
             AnsiUi.error(message);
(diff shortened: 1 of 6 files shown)


demobox:
> hal config telemetry enable

Perhaps `hal config stats enable`?

/cc @dorbin

ttomsu:
Changing from `stats` to `telemetry` is a much bigger change, and one I'm not a big fan of - telemetry is more semantically accurate at the point of collection, and they're turned into stats on the backend by analysis. 

That being said, I realize we use the term interchangeably through the codebase, and I would prefer to use one term.

dorbin:
@ttomsu , @demobox 
I have  no opinion on "stats" v. "telemetry." Either is ok with me.

ttomsu:
Since changing the wording is out of scope for this PR, I'll merge as is and we can bikeshed on it in another PR.

PR #1493, opened 2020-01-07
feat(features): add support for Cloud Formation

ferwguerra:


docs/commands.md
```diff
@@ -6145,6 +6145,7 @@ hal config features edit [parameters]
  * `--artifacts`: Enable artifact support. Read more at [https://spinnaker.io/reference/artifacts/](https://spinnaker.io/reference/artifacts/)
  * `--artifacts-rewrite`: Enable new artifact support. Read more at [https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/](https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/)
  * `--chaos`: Enable Chaos Monkey support. For this to work, you'll need a running Chaos Monkey deployment. Currently, Halyard doesn't configure Chaos Monkey for you; read more instructions here [https://github.com/Netflix/chaosmonkey/wiki](https://github.com/Netflix/chaosmonkey/wiki).
+ * `--cloud-formation`: Enable Cloud Formation support.
```
gregjones: supernit, but "CloudFormation" for consistency in the description

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/EditFeaturesCommand.java
```diff
@@ -102,6 +102,12 @@
       arity = 1)
   private Boolean gremlin = null;
 
+  @Parameter(
+      names = "--cloud-formation",
+      description = "Enable Cloud Formation support.",
```
gregjones: "CloudFormation" here too

xavileon:
@spinnaker/oss-approvers PTAL

xavileon:
Hello @spinnaker/oss-approvers, can we have some feedback wether this is a valid approach for enabling the CFN feature through halyard?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/ClouddriverProfileFactory.java
```diff
@@ -85,6 +80,14 @@ protected void setProfile(
     if (deploymentConfiguration.getProviders() != null) {
       processProviders(deploymentConfiguration.getProviders());
     }
+    Features features = deploymentConfiguration.getFeatures();
```
ezimanyi: It looks like this PR is creating two places in the hal config to set whether cloudformation is enabled:
* `features.cloudFormation`
* `providers.aws.features.cloudFormation`

The second of these (`providers.aws.features.cloudFormation`) is where we actually want the setting to live as it's where it ends up in the clouddriver config output; this code here looks to be looking at the first location and updating the second to match it.

I don't think we need the `features.cloudFormation` at all; I think that the single source of truth for the value of the setting should be `providers.aws.features.cloudFormation`, which removes the need for this code here and avoids confusingly having two places in the hal config to set it.

In addition to removing the field `features.cloudFormation`, I think you'll also need to change where you've added the command; this should probably be something like `hal config provider aws features ...` and should be registered in `AwsCommand.java`.

ferwguerra: @ezimanyi Thanks for the feedback!
I agree with you on the changes to be made.
As soon as I have time I will implement them. Thanks again!

ferwguerra: @ezimanyi could you take a look to the new solution?  Thank you!

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/providers/AbstractEditFeaturesProviderCommand.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2017 Google, Inc.
```
ezimanyi: This should reference you or your company (and year should be updated).

ferwguerra: @ezimanyi Thank you for the feedback. Changes done :)

PR #1498, opened 2020-01-15
feat(plugins): able to add plugin repositories

link108:
This PR enables configuring plugin repositories for PF4J to consume. See https://github.com/spinnaker/spinnaker/issues/5243 and https://github.com/pf4j/pf4j-update

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,4 +1,4 @@
-FROM alpine:3.10
+FROM alpine:3.11
 LABEL maintainer="delivery-engineering@netflix.com"
 COPY --from=halyard.compile /compiled_sources/halyard-web/build/install/halyard /opt/halyard
 
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -11728,7 +11728,7 @@ hal plugins repository add REPOSITORY [parameters]
 `REPOSITORY`: The name of the plugin repository to operate on.
  * `--deployment`: If supplied, use this Halyard deployment. This will _not_ create a new deployment.
  * `--no-validate`: (*Default*: `false`) Skip validation.
- * `--url`: (*Required*) The location of the plugin repo.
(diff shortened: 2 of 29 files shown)


ezimanyi:
This looks reasonable (and maybe I'm missing this) but where are these settings actually used in the service configs?  (Or is that not part of this PR?)

link108:
@ezimanyi This outputs the following to the service configs:
```
spinnaker:
  extensibility:
    repositories:
      my-repo-id:
        url: https://my.repo.url/
```

In each repository, there should be a file, `plugins.json` that describes the available plugins, see [pf4j-update's repository structure](https://github.com/pf4j/pf4j-update#repository-structure)


Do note, this is not sufficient for halyard supporting PF4J plugins, but is a stepping stone to get there. There will be another Halyard PR that will deprecate the previous plugin configurations and transition to only supporting PF4J plugins. This second PR will be a breaking change to whomever is using the MVP implementation of plugins, as Halyard will output plugin configurations for use with PF4J, not the MVP implementation.

link108:
We discussed this in [slack](https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1578510602005000) and then today in the Platform SIG meeting (see [agenda](https://docs.google.com/document/d/1zGIgEwO9lZfYaI0N6necuH-x0SjmL3hmbbeMWZgqnEw/edit#heading=h.n1w9jnw3t7un)). We will be producing another PR after this is merged to deploy a `plugin-manifest.json` for Deck to consume, while using PF4J-Update's repositories list to inform the backend services which plugins to install. This is an intermediary step for moving off of the MVP implementation, while Plugin Bundles and the Deck<->Front50 interactions are being built.

link108:
@ezimanyi I updated the PR such that the hal config and the service configs are in the same format.

PR #1502, opened 2020-01-19
feat(deployment): Overwrite deploy command's timeout value

ryu1kn:
This PR introduces `--wait-for-completion-timeout-minutes` option for `hal deploy apply` command and allows us to overwrite its 10 minutes default timeout value. The issue is reported in https://github.com/spinnaker/spinnaker/issues/5185.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -11442,6 +11442,7 @@ This guarantees that no configuration will be generated for this deployment. Thi
  * `--prep-only`: (*Default*: `false`) This does just the prep work, and not the actual deployment. Only useful at the moment if you want to just clone the repositories for a localgit setup.
  * `--service-names`: (*Default*: `[]`) When supplied, only install or update the specified Spinnaker services.
  * `--wait-for-completion`: (*Default*: `false`) When supplied, wait for all containers to be ready before returning (only applies to Kubernetes V2 provider).
+ * `--wait-for-completion-timeout-minutes`: Specify timeout for deploy apply command.
 
 
 ---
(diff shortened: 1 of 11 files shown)


ryu1kn:
Hello @ezimanyi @maggieneterval , do you know who can review this PR? I'd like to get feedback on this.

ezimanyi:
@ryu1kn : Thanks for the PR! I'm wondering what value you were thinking of setting this timeout to?  If this is frequently taking a long time, I wonder if we should just increase the default upstream rather than make this configurable.  (I could imagine extending this to 20 or 30 minutes by default.)

ryu1kn:
Thanks @ezimanyi . I’d set it to 30 mins for our case though most of the time it finishes within 20. It takes time as it’s doing rolling update, and the time depends on how many pods we’re running.

ryu1kn:
Hmm... but I am wondering if making it configurable still worth it.

I don't want a deployment to be terminated because it took slightly longer than usual and left the environment broken. If I need to come back to raise another PR to extend the timeout value, I feel I need to suggest a bigger number which can be waste for other people. Maybe they know their Spinnaker deployment is not going to succeed anyway if it takes longer than 10 mins 🤔

ryu1kn:
@ezimanyi Let me know how we want to go with this. If you prefer keeping the feature/code simpler over making the timeout value configurable, I’ll update the PR by just extending the timeout to 30 mins.

ryu1kn:
@ezimanyi Thanks for proceeding this PR! 👍

PR #1507, opened 2020-01-23
fix(kubernetes): Only load kubeconfig if it's a local file

german-muzquiz:
Closes https://github.com/spinnaker/spinnaker/issues/5305

When kubeconfig is stored remotely, Halyard shouldn't try to modify it in place because clouddriver will download it from remote source anyway.


german-muzquiz:
cc @ezimanyi

PR #1510, opened 2020-01-23
config(core): Remove MPTV2 UI feature flag

louisjimenez:
Accompanies this [Deck PR](https://github.com/spinnaker/deck/pull/7804)


halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Features.java
```diff
@@ -22,7 +22,7 @@
 
 @Data
 @EqualsAndHashCode(callSuper = false)
-@JsonIgnoreProperties({"jobs"})
+@JsonIgnoreProperties({"jobs", "managedPipelineTemplatesV2UI"})
```
maggieneterval: Just followed the link from the Deck PR specifically to make sure this was here but you are waaay ahead of me
:heart:

PR #1517, opened 2020-01-28
chore(java11): Switch the containers to use Java 11.

plumpy:
Halyard was accidentally forgotten the [the Java 11 RFC](https://github.com/spinnaker/governance/blob/master/rfc/java11.md). Unlike the other containers, I'm not going to provide Java 8 fallback containers for Halyard, since users can just use 1.30 if they have issues with Java 11. If there's a lot of demand for a Java 8 container, I'll look into providing them.


plumpy:
nevermind, gonna try to make Java 8 containers too. BRB

PR #1520, opened 2020-01-29
feat(plugins): deploy PF4J plugins to orca

link108:
As referenced in https://github.com/spinnaker/halyard/pull/1498#issuecomment-574904828, this PR adds the ability to deploy PF4J plugins to Orca and removes the ability for Halyard to deploy MVP plugins. 

I plan on updating Halyard to deploy plugins to more services, however some services already have `spinnaker` as a top level key in their service config yaml and they fail to startup due to duplicate key. 

We will be producing another PR to deploy the `plugin-manifest.json` to Deck.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/AddPluginCommand.java
```diff
@@ -33,23 +35,33 @@
...
   protected void executeThis() {
     String currentDeployment = getCurrentDeployment();
     String name = getPluginName();
     Plugin plugin =
         new Plugin()
-            .setName(name)
+            .setId(name)
             .setEnabled(isSet(enabled) ? Boolean.parseBoolean(enabled) : false)
-            .setManifestLocation(manifestLocation);
+            .setUiResourceLocation(isSet(uiResourceLocation) ? uiResourceLocation : null)
```
ezimanyi: I'm not sure the `isSet` is doing anything here...this is effectively doing `uiResourceLocation != null ? uiResourceLocation : null` which is just the same as always setting `uiResourceLocation`.

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/AddPluginCommand.java
```diff
@@ -33,23 +35,33 @@
...
+            .setVersion(version);
+    Arrays.stream(extensions.split(","))
+        .distinct()
```
ezimanyi: Do we need a `distinct` here if we're about to put the result as the key of a `Map`?

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/EditPluginCommand.java
```diff
@@ -33,26 +34,42 @@
...
 
     plugin.setEnabled(isSet(enabled) ? Boolean.parseBoolean(enabled) : plugin.getEnabled());
-    plugin.setManifestLocation(
-        isSet(manifestLocation) ? manifestLocation : plugin.getManifestLocation());
+    plugin.setUiResourceLocation(
+        isSet(uiResourceLocation) ? uiResourceLocation : plugin.getUiResourceLocation());
+    plugin.setVersion(isSet(version) ? version : plugin.getVersion());
+
+    for (String pluginExtensionName : extensions.split(",")) {
+      if (!plugin.getExtensions().containsKey(pluginExtensionName)) {
```
ezimanyi: Looks like this would be simpler as `putIfAbsent`

halyard-config/src/test/groovy/com/netflix/spinnaker/halyard/config/model/v1/ExtensibilitySpec.groovy
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package com.netflix.spinnaker.halyard.config.model.v1
+
+import com.netflix.spinnaker.halyard.config.model.v1.node.Extensibility;
+import com.netflix.spinnaker.halyard.config.model.v1.plugins.Plugin
+import com.netflix.spinnaker.halyard.config.model.v1.plugins.PluginExtension
+import spock.lang.Specification
+
+class ExtensibilitySpec extends Specification {
```
ezimanyi: :tada:

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/AddPluginCommand.java
link108: probably not, would potentially reduce the number of objects to be GC'ed, in the case of extensions being added, but I don't see that having a big effect either way 👍

PR #1522, opened 2020-01-30
fix(plugins): Ignore extraneous pluginConfigurations when parsing con…

duftler:
…fig file.


duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/5371.

PR #1530, opened 2020-02-06
Revert "config(core): Remove MPTV2 UI feature flag (#1510)"

louisjimenez:
This reverts commit 759e98840f7d67ec5b2e692a51cb6d8d0cadef56.


louisjimenez:
Reverting this until Spinnaker 1.19 is released so that users can continue to use halyard to enable the mptv2 feature flag

maggieneterval:
Is this because your PR that removed the flag from Deck didn't make it into 1.18?

louisjimenez:
@maggieneterval That's right, the Deck PR targets 1.19 and onwards

louisjimenez:
@spinnakerbot cherry-pick 1.31

PR #1539, opened 2020-02-14
feat(plugins): lays down plugin-manifest for Deck

bpowell:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/PluginManifestProfileFactory.java
```diff
@@ -0,0 +1,61 @@
...
+                  && p.getUiResourceLocation() != "";
+            })
+        .forEach(
+            v -> {
+              Plugin p = v.getValue();
+              profile.appendContents("{");
+              profile.appendContents("id: \"" + p.getId() + "\",");
+              profile.appendContents("version: \"" + p.getVersion() + "\",");
+              profile.appendContents("url: \"" + p.getUiResourceLocation() + "\",");
+              profile.appendContents("},");
```
german-muzquiz: Small suggestion: probably readability could improve by using a string template constant and use `String.format` to build the final string, instead of appending each token individually:
```
private static final String PLUGIN_ENTRY = "{ id: \"%s\", version: \"%s\", url: \"%s\", },"
...
profile.appendContents(String.format(PLUGIN_ENTRY, p.getId(), p.getVersion(), p.getUiResourceLocation()));
```
Similarly, a different string template could be used to build the full string:
```
private static final String PLUGINS_TEMPLATE = "const plugins = [%s]; export { plugins };"
```

link108:
before we merge it, could we update to lay down `plugin-manifest.json` as described in https://github.com/spinnaker/deck/pull/7902 ?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeckService.java
```diff
@@ -99,6 +102,12 @@ public Type getType() {
         apacheSpinnakerProfileFactory.getProfile(
             "apache2/" + filename, path, deploymentConfiguration, endpoints));
 
+    filename = "plugin-manifest.js";
```
link108: should be `.json`

link108:
LGTM 👍

bpowell:
@german-muzquiz can we please get a merge on this?

german-muzquiz:
cc @ezimanyi

PR #1542, opened 2020-02-18
feat(kubernetes): Enable TCP probe for containers

ncknt:
Adding a flag to switch readiness to TCP probe. The reason for that is to enable deployments with mTLS where the `wget` readiness probe won't work. We could also allow users to provide their own `exec` command to use a valid client certificate but that's not covered by this PR :)

To enable the TCP probe, add the following to service-settings/<service>.yml:
```
kubernetes:
  useTcpProbe: true
```

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
@@ -423,7 +423,8 @@ public interface KubernetesV2Service<T> extends HasServiceSettings<T>, Kubernete
 
   default TemplatedResource getProbe(ServiceSettings settings, Integer initialDelaySeconds) {
     TemplatedResource probe;
-    if (StringUtils.isNotEmpty(settings.getHealthEndpoint())) {
+    if (StringUtils.isNotEmpty(settings.getHealthEndpoint())
+        && !settings.getKubernetes().getUseTcpProbe()) {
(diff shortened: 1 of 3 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/v2/KubernetesV2Service.java
```diff
@@ -423,7 +423,8 @@ default String buildContainer(
 
   default TemplatedResource getProbe(ServiceSettings settings, Integer initialDelaySeconds) {
     TemplatedResource probe;
-    if (StringUtils.isNotEmpty(settings.getHealthEndpoint())) {
+    if (StringUtils.isNotEmpty(settings.getHealthEndpoint())
```
maggieneterval: Any thoughts on improving readability by lifting up the TCP case so we can keep the flag closest to the logic it actually gates? Something like "if no health endpoint OR useTcpProbe flag is on, use the TCP probe"

ncknt: Updated!

maggieneterval: You are so fast!!

ncknt: updated the !notEmpty

PR #1543, opened 2020-02-20
feat(telemetry): add ability to report deploy method

castanedaci:
This is a follow-up to spinnaker/kork#503 and /spinnaker/echo#770 where we added deploy method and version. 
This change in halyard allows sending the telemetry information with the deployment method and version.


german-muzquiz:
@ezimanyi that's right, the idea is that halyard will just set two new fields in `echo.yml` (deployment type and version), and echo will send this information to telemetry service.

PR #1550, opened 2020-02-26
feat(stats): Make stats collection opt-out

ttomsu:
Here's the experience:

1. Beginning state of the config:
    ````
    telemetry:
      enabled: false
    ````
1. User updates to new Halyard
1. User runs read-only operation, like `hal config`
    ```
    - INFO Telemetry is now opt-out. Your telemetry preference is
    currently unset, so it is ENABLED....
    ```
    ```
    telemetry:
      enabled: false
    ```
1. User runs a mutating operation, like `hal config version edit --version 1.19.0`. This flips the `enabled` field and adds the `explicitlySet` field. When `explicitlySet` is false, telemetry is enabled, regardless of the `enabled` field's value (hence opt-out). Flipping it in config is meant to reflect this.
    ```
    telemetry:
      enabled: true
      explicitlySet: false
    ```
1. User runs a deploy operation:
    ```
    - INFO Telemetry is now opt-out. Your telemetry preference is
    currently unset, so it is ENABLED.
    ```
1. User runs `hal config telemetry disable`:
    ```
    - INFO Telemetry is now opt-out. Your telemetry preference is
    currently set to DISABLED.... 
    ```
    ```
    telemetry:
      enabled: false
      explicitlySet: true
    ```
1. User runs `hal config telemetry enable`:
    ```
    - INFO Telemetry is now opt-out. Your telemetry preference is
    currently set to ENABLED.
    ```
    ```
    telemetry:
      enabled: true
      explicitlySet: true
    ```

cc @imosquera @ncknt

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/config/v1/HalconfigParser.java
@@ -229,6 +229,20 @@ public class HalconfigParser {
               .build());
     }
 
+    // Added for the the opt-in/opt-out switch for release 1.19
+    // This ensures that telemetry.enabled reflects the enabled/disabled state truthfully in the
+    // written halconfig. Without it, one would see this, yet telemetry would actually be enabled:
+    //   telemetry:
+    //     enabled: false
+    //     explicitlySet: false
+    local.getDeploymentConfigurations().stream()
+        .forEach(
+            dc -> {
+              if (!dc.getTelemetry().getExplicitlySet()) {
+                dc.getTelemetry().setEnabled(true);
+              }
+            });
+
     AtomicFileWriter writer = null;
     try {
       writer = new AtomicFileWriter(path);
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Telemetry.java
@@ -34,7 +34,12 @@ public class Telemetry extends Node {
   @ValidForSpinnakerVersion(
       lowerBound = "1.18.0",
       tooLowMessage = "Telemetry is not available prior to this release.")
-  private Boolean enabled = false;
+  private Boolean enabled = true;
+
+  @ValidForSpinnakerVersion(
+      lowerBound = "1.19.0",
+      tooLowMessage = "Will enable telemetry when not explicitly disabled.")
+  private Boolean explicitlySet = false;
 
   private String endpoint = DEFAULT_TELEMETRY_ENDPOINT;
   private String instanceId = new ULID().nextULID();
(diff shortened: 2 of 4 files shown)


plumpy:
Oof. `explicitlySet`.

What about changing the name from `telemetry` to `metricsCollection` or something, allowing us to start over again fresh?

ttomsu:
That would be a *significantly* bigger change. I'd also say having two blocks that are synonymous would cause confusion, and deleting an old block isn't straightforward (as the `plugin` removal proved).

plumpy:
> That would be a significantly bigger change.

Not sure that's particularly relevant, TBH.

The docs we will have to keep forever now will say "if you want to opt out of metrics collection, add this stanza to your halconfig:
```
telemetry:
  enabled: false
  explicitlySet: true
```

That feels like a Bad Plan to me. People are going to try to opt out and still be sending metrics because we made the opt-out process super confusing. Then they will be Very Mad. Doing more work now to make this less confusing for our users seems well worth the effort, IMO.

ttomsu:
> The docs we will have to keep forever now will say "if you want to opt out of metrics collection, add this stanza to your halconfig:

This is not true. The docs (both on [spinnaker.io](https://www.spinnaker.io/community/stats/#how-is-the-data-collected) and in the Halyard output for *every* deploy command) clearly state:

```
To disable, run `hal config telemetry disable`
```

The primary halconfig interaction is still through the CLI. And besides, with Kleat it seems Halyard is going away long term anyway.

plumpy:
People most certainly hand-edit their halconfigs, though. Pretending it's a binary blob that people don't look at, edit, copy/paste around, etc. just isn't true.

Also, kleat is intending to be almost 100% backwards compatible with existing halconfigs, except now people will _definitely_ be hand-editing these files, so I don't think this is something we can reasonably live with...

ttomsu:
That's fine that they hand edit it, but the _default_ is to use the CLI. This change makes it so that this block is as consistent as can be. It clearly informs the user what the current status is and how to change it. I think replacing it with a different block is completely unnecessary, and would cause more confusion and complication, not less.

plumpy:
> That's fine that they hand edit it, but the default is to use the CLI

That's true for now. That will not be true in the future. We already know that but you seem to be ignoring that fact entirely.

Given that, doesn't this feel a little short-sighted? Or at least can you think about what this looks like a year from now? Does `explicitlySet` still have to be around?

ttomsu:
It doesn't have to be around a year from now. 

This is most a short term patch in order to migrate users who had the previous Halyard versions write out `telemetry.enabled: false` to their halconfigs (because it was by default turned off). In fact, once the first mutating operation happens (one that writes the halconfig back to disk), if I could switch it _then_, there'd be no reason for `explicitlySet`. 

AFAIK, nothing like that exists. The daemon doesn't read the config until the first operation happens.

plumpy:
I'd love to talk through how that migration happens. You have a great plan above for how things get migrated to this new `explicitlySet` world. But what's the plan for removing that config parameter? How do we do it safely and without accidently opting people into data collection?

What if instead of `enabled` it was `sendData` (or something) and your migration removed the `enabled` setting. So it went from
```
telemetry:
  enabled: false
```
to
```
telemetry:
  sendData: true
```

That seems like it would be easy to do (mark `enabled` as an ignored parameter, which would also prevent it from getting written out, then add a new `sendData` which defaults to true) and is less weird than `explicitlySet`

PR #1551, opened 2020-02-26
config(core): Revert "fix(flags): Revert "config(core): Remove MPTV2 …

louisjimenez:
…UI feature flag (#1510)" (#1530)"

The feature flag for the MPTV2 UI was removed from halyard but had to be reverted since 1.19 had yet to be released. This reverts that revert and will go out with 1.19.


louisjimenez:
Delaying the removal of this flag until Spinnaker 1.21. This will give users more time to migrate to >=1.19 versions of Spinnaker that will have the templates UI on by default. This delay is intended to support users that are on the latest version of Halyard but deploying to Spinnaker <1.19 and want to use Halyard to turn on the UI.

gegulati:
Hi @louisjimenez , Is this feature flag for UI available  in 1.24.2 release ?  Halyard version 1.40.

PR #1554, opened 2020-02-27
feat(codebuild): Add support for static credentials

Kaixiang-AWS:
1. Add an AbstractEditCiCommand class that could be inherited from. For now only codebuild account
command extends this class. There could be more in the future.
2. Mark assume-role and account-id as not required so that the static credentials could be used directly
3. Add an API in halyard daemon to set custom fields for CI providers

Example:
```
codebuild:
  enabled: true
  accessKeyId: 'access-key-id'
  secretAccessKey: 'secret-access-key'
  accounts:
  - name: my-codebuild-account
    accountId: '123456789012'
    assumeRole: role/spinnakerManaged
    region: us-west-2
  - name: my-codebuild-eu-central-1
    accountId: '123456789012'
    assumeRole: role/spinnakerManaged
    region: eu-central-1
  - name: my-codebuild-no-assume-role
    region: us-west-2
```
In the config above, `my-codebuild-account` will use static credentials to assume role `spinnakerManaged` in us-west-2. `my-codebuild-eu-central-1` will use static credentials to assume role `spinnakerManaged` in eu-central-1. `my-codebuild-no-assume-role` will directly use static credentials to call CodeBuild in us-west-2

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -184,6 +184,7 @@
  * [**hal config ci codebuild account get**](#hal-config-ci-codebuild-account-get)
  * [**hal config ci codebuild account list**](#hal-config-ci-codebuild-account-list)
  * [**hal config ci codebuild disable**](#hal-config-ci-codebuild-disable)
+ * [**hal config ci codebuild edit**](#hal-config-ci-codebuild-edit)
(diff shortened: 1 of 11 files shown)


clareliguori:
CodeBuild parts LGTM, but I'd like someone else to review the core halyard class changes as well

PR #1559, opened 2020-03-03
feat(plugins): add plugins to all services

link108:
Now that PF4J has been introduced to Spinnaker, Halyard should deploy plugin configuration to the individual services. 

Previous discussion on slack: https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1582934543022800

This is dependent on https://github.com/spinnaker/echo/pull/792/files.


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/Front50ProfileFactory.java
```diff
@@ -59,8 +59,8 @@ public void setProfile(
       throw new HalException(Problem.Severity.FATAL, "No persistent storage type was configured.");
     }
 
+    Map<String, Object> spinnakerYaml = deploymentConfiguration.getSpinnaker().toMap();
```
ezimanyi: I'm not sure I understand the change in this file...are we now embedding the entire `spinnaker.yml` inside `front50.yml`?

Can you show me an example of what the resulting `front50.yml` looked like before the change and what it looks like after?

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
```diff
@@ -41,6 +43,22 @@ protected void setProfile(
 
     profile.appendContents(
         yamlToString(deploymentConfiguration.getName(), profile, spectatorConfig));
+
+    // Front50 contains persistent storage configuration under the spinnaker top level key, so we
+    // add extensibility configuration for Front50 in Front50ProfileFactory
+    if (this.getClass() != Front50ProfileFactory.class) {
```
ezimanyi: Instead of introspecting to figure out what class this is, shouldn't this use an overriden method instead?  In general it's a bad pattern for a parent class to switch on what the implementing class is.

link108: ah, so you're saying to remove this check and update Front50ProfileFactory to override setProfile? That makes sense to me and would be cleaner 👍

ezimanyi: Yes, exactly, or even something `boolean includeExtensibilityConfig()` that defaults to true in the base class but is overriden to false in `Front50` would also work, and this block could check that flag.

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/Front50ProfileFactory.java
link108: ah, so this spinnakerYaml is not `spinnaker.yml`, I can update that so its a bit more clear. This is the `spinnaker` top level key in the hal config, eg this is what contains the plugin configuration. 

`front50.yml` currently contains:
```
## WARNING
## This file was autogenerated, and _will_ be overwritten by Halyard.
## Any edits you make here _will_ be lost.

spectator:
  applicationName: ${spring.application.name}
  webEndpoint:
    enabled: false

spinnaker:
  s3:
    bucket:
    ...
  oracle:
    enabled: false
  gcs:
    rootFolder: front50
    enabled: false
  azs:
    enabled: false
  redis:
    host: spin-redis.cam
    port: 6379
    enabled: false

server:
  port: ${services.front50.port:8080}
  address: ${services.front50.host:localhost}

cassandra:
  enabled: false

hystrix:
  command:
    default.execution.isolation.thread.timeoutInMilliseconds: 15000
  threadpool:
    DefaultNotificationDAO:
      coreSize: 25
      maxQueueSize: 100
    DefaultPipelineDAO:
      coreSize: 25
      maxQueueSize: 100
```


After this change, `front50.yml` would look like:
```
## WARNING
## This file was autogenerated, and _will_ be overwritten by Halyard.
## Any edits you make here _will_ be lost.

spectator:
  applicationName: ${spring.application.name}
  webEndpoint:
    enabled: false

spinnaker:
  extensibility:
    plugins:
      pluginA:
         ...
    repositories:
      repositoryA:
        url: https://example.com
    plugins-root-path: /opt/front50/plugins
  s3:
    bucket:
    ...
  oracle:
    enabled: false
  gcs:
    rootFolder: front50
    enabled: false
  azs:
    enabled: false
  redis:
    host: spin-redis.cam
    port: 6379
    enabled: false

server:
  port: ${services.front50.port:8080}
  address: ${services.front50.host:localhost}

cassandra:
  enabled: false

hystrix:
  command:
    default.execution.isolation.thread.timeoutInMilliseconds: 15000
  threadpool:
    DefaultNotificationDAO:
      coreSize: 25
      maxQueueSize: 100
    DefaultPipelineDAO:
      coreSize: 25
      maxQueueSize: 100
```

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
link108: I like it!

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/Front50ProfileFactory.java
ezimanyi: Oh, I see, this is just because the stuff under `spinnaker` now has more than just persistent stores so you renamed the variable to match what it now contains.  Makes sense!

halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
```diff
@@ -41,6 +43,24 @@ protected void setProfile(
...
+          "plugins-root-path", "/opt/" + this.getArtifact().toString().toLowerCase() + "/plugins");
+      extensibilityYaml.put("extensibility", extensibilityContents);
+      spinnakerYaml.put("spinnaker", extensibilityYaml);
+
+      profile.appendContents(
+          yamlToString(deploymentConfiguration.getName(), profile, spinnakerYaml));
+    }
+  }
+
+  public Boolean addExtensibilityConfigs() {
```
ezimanyi: I think this can be `protected`.  It should also return a `boolean` instead of a `Boolean` because we never expect it to be null.

ezimanyi:
Looks like the branch doesn't allow modifications from maintainers so neither I nor mergify can merge this.

maggieneterval:
Hey link108 ! I don't have a ton of context on this change but I think it is breaking the builds. I'm seeing Echo fail to start up ([full logs](https://builds.spinnaker.io/job/Validate_BomMultiPlatform/4506/CLOUD_PLATFORM=ec2/artifact/logs/service_logs/echo.log)) with the following exception indicating a duplicate key `spinnaker` in its config:

```
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:         at com.netflix.spinnaker.echo.Application.main(Application.java:42) [echo-web-2.9.1-20191216151527.jar:2.9.1-20191216151527]
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]: Caused by: org.yaml.snakeyaml.constructor.DuplicateKeyException: while constructing a mapping
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:  in 'reader', line 5, column 1:
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:     spectator:
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:     ^
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]: found duplicate key spinnaker
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:  in 'reader', line 90, column 1:
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:     spinnaker:
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:     ^
Mar 27 06:38:55 ip-172-31-1-154 echo[21205]:         at org.yaml.snakeyaml.constructor.SafeConstructor.processDuplicateKeys(SafeConstructor.java:100) ~[snakeyaml-1.23.jar:na]
```

I see that this PR is in fact putting some stuff under a `spinnaker` key:

`spinnakerYaml.put("spinnaker", extensibilityYaml);`

Any thoughts on this?

ezimanyi:
I think @maggieneterval is correct and this is breaking deploying any version of Spinnaker <= 1.18 with Halyard.  While spinnaker/echo#792 removed the `spinnaker` stanza from the `echo` `halconfig` block, that only affects versions of Spinnaker 1.19+, so anyone who upgrades Halyard won't be able to deploy older versions of Spinnaker.

We'll need to revert this until the plugins team can come up with a plan that retains backwards compatibility.

PR #1565, opened 2020-03-10
feat(stats): Rename telemetry to stats, enable by default

ttomsu:
This is largely a `s/telemetry/stats/g` change, with some minor tweaks to ignore the old `telemetry` block (see `DeploymentConfiguration.java`). This change enables telemetry stats by default, and makes https://github.com/spinnaker/halyard/pull/1550 obsolete.

One known effect of this will be that all existing instance IDs will be regenerated. This may show a bump in unique instance counts as users upgrade their Halyard instances. Since there is only 118 28-day-active instances anyway, I think it'll largely be swallowed by the expected increase of newly enabled instances reporting metrics.

@ezimanyi Once this is in and Halyard is released, we'll need to indicate that the newly released Halyard version is required for 1.19.0.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/EchoProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/EchoProfileFactory.java
@@ -108,7 +108,7 @@ public class EchoProfileFactory extends SpringProfileFactory {
       stats.setSpinnakerVersion(statsVersion);
       stats.setDeploymentMethod(deploymentMethod());
       profile.appendContents(
-          yamlToString(deploymentConfiguration.getName(), profile, new TelemetryWrapper(stats)));
+          yamlToString(deploymentConfiguration.getName(), profile, new StatsWrapper(stats)));
     }
 
     profile.appendContents(profile.getBaseContents()).setRequiredFiles(files);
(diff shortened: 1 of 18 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/EchoProfileFactory.java
```diff
@@ -154,12 +152,13 @@ private String halyardVersion() {
...
   private static class TelemetryWrapper {
-    private Telemetry telemetry;
+    private Stats telemetry;
 
-    TelemetryWrapper(Telemetry telemetry) {
-      this.telemetry = telemetry;
+    TelemetryWrapper(Stats stats) {
+      this.telemetry = stats;
     }
   }
```
plumpy: Can you _also_ write it out to `stats` and then update echo to read it from that location instead?

ezimanyi: Good point, I had missed that in my review of the PR.  In general, I'd like to minimize the difference between what is in the hal config and what is in each service's config, both to simplify the logic in `kleat` and also to make it easier on anyone trying to understand how these settings map to each other.

ttomsu: Opened up two other PRs in echo and buildtool to make this happen.

PR #1572, opened 2020-03-15
 feat(k8s/deploy): Add custom health check to service-settings

clanesf:
In configuring our services to use mTLS we ran into an issue with the 
standard halyard deployment. The issue is that when https is enabled and
client certificates are required, the services never come up because
the health checks fail and are rejected because of a lack of a client certificate.

This PR adds the ability to create a custom exec-based health check and 
configure each service's health check as needed to handle client certificates
and potentially other network related issues.

For example, by adding the following to `.hal/default/service-settings/clouddriver.yml`
Halyard will deploy a health check that provides the client cert and key provided
that they are present on the pod in the specified location

```
# clouddriver.yml service settings file with custom defined health check
kubernetes:  
  volumes:
  - id: spin-mtls-certs
    mountPath: /opt/certs/
    type: secret
  customHealthCheckExecCommands: 
  - wget
  - --spider
  - --ca-certificate=/opt/certs/k8s-cluster-ca.crt
  - --certificate=/opt/certs/client.cert.crt
  - --private-key=/opt/certs/client.key.pem
  - -q
  - https://localhost:7002/health
```

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-#Fri Mar 13 21:28:06 PDT 2020
+#Sat Mar 14 20:40:20 UTC 2020
 enablePublishing=false
-korkVersion=7.25.3
+korkVersion=7.26.0
 org.gradle.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-#Wed Mar 11 01:03:44 UTC 2020
+#Fri Mar 13 21:28:06 PDT 2020
 enablePublishing=false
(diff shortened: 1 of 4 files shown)


clanesf:
@cfieber - When you get a chance can you review this? We'd like to rollout mTLS next week across our Spinnaker instances and right now not being able to use Halyard to customize the health check to include the client certificates is blocking our forward progress.

cfieber:
this looks reasonable to me but I use neither halyard or k8s so I'd prefer someone else to weigh in

PR #1578, opened 2020-03-23
feat(build): Remove TravisCI and replace with GitHub Actions

jonsie:


ttomsu:
Have you taken those scripts through a trial run?

jonsie:
@ttomsu I don't see anything unique here compared to the service builds, and this is the same pattern we're using elsewhere so this all should work.

PR #1594, opened 2020-04-02
feat(plugins): add plugins to all services

link108:
Now that PF4J has been introduced to Spinnaker, Halyard should deploy plugin configuration to the individual services.

Previous discussion on slack: https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1582934543022800
Previous PR: https://github.com/spinnaker/halyard/pull/1559

I tested this with the `spinnaker-release-1.19.x-latest-unvalidated` service builds and setting the spinnaker version to `1.18.0` to ensure plugin configs weren't sent to the services.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
@@ -17,9 +17,12 @@
 package com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile;
 
 import com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentConfiguration;
+import com.netflix.spinnaker.halyard.core.registry.v1.Versions;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSettings;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.service.ServiceSettings;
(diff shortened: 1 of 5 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringProfileFactory.java
```diff
@@ -41,6 +44,36 @@ protected void setProfile(
...
+        "plugins-root-path", "/opt/" + this.getArtifact().toString().toLowerCase() + "/plugins");
+    extensibilityContents.put("strict-plugin-loading", false);
+    extensibilityYaml.put("extensibility", extensibilityContents);
+    spinnakerYaml.put("spinnaker", extensibilityYaml);
+    return spinnakerYaml;
+  }
+
+  protected boolean spinnakerVersionSupportsPlugins(
+      DeploymentConfiguration deploymentConfiguration) {
+    return Versions.greaterThanEqual(deploymentConfiguration.getVersion(), "1.19.4");
```
link108, line 72: this assumes the 1.19.4 release contains the kork fixes: https://github.com/spinnaker/kork/pull/586/files

ezimanyi: How will this handle cases where people are deploying things like `master-latest-unvalidated` or `release-1.19.x-latest-validated`, etc.?

link108: updated and added tests 👍

link108:
Thats a good point, I don't think there are many/any plugin users right now, but in case there are, I'll add a check to deploy Orca configs.

PR #1596, opened 2020-04-03
fix(container): restore openssl to alpine image

dbyron0:
removed by https://github.com/spinnaker/halyard/commit/43dfc6744ea6dbe6fd65cb349a53696d96329dae

In addition to putting back something that was there previously, openssl is in the ubuntu
container, so let's have it here as well.

Not sure whether `fix` is really fair here?  Maybe this is more `feat` ?

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -10,6 +10,7 @@ RUN apk --no-cache add --update \
   bash \
   curl \
   openjdk8-jre \
+  openssl \
   py-pip \
   python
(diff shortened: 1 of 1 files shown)


plumpy:
Very reasonable, thanks for the PR and sorry for the churn...

dbyron0:
Thank you!!

dbyron0:
Any chance of a 1.33.1 with this in case 1.34 is gonna be awhile?

plumpy:
I shipped a 1.34!

PR #1614, opened 2020-04-14
fix(publishing): No longer need to change ACL on halyard artifact sin…

ttomsu:
…ce the whole bucket is public


ttomsu:
Post-merge tag: this change was related to https://github.com/spinnaker/spinnaker/issues/5434

PR #1618, opened 2020-04-15
Fix Debian install script if sysusers.d is not present

mmlac:
fixes spinnaker/spinnaker#4964

Debian installer fails when the sysusers.d directory is not present. That is the case on e.g. GCP Debian 9 images.


Mahito:
LGTM, but first commit(346587d) does not Commit message conventions.
I think you have to do `git rebase -i` to merge the commits.

PR #1640, opened 2020-04-22
fix(google): Update constructors to account for retry changes

ezimanyi:
The changes to the retry logic updated some fields from Long to int (because we won't be waiting 24 days between retries); this commit just updates the places we called them in Halyard to pass an int.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 clouddriverVersion=5.58.0
 fiatVersion=1.18.3
-front50Version=2.17.1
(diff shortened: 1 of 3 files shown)


ezimanyi:
This will be required once I cut a release of front50 containing spinnaker/front50#783; marked as draft because I need to bump the `front50` version in the same commit.

ezimanyi:
Just pulled in the latest front50, and un-marked as draft.  If the tests pass this should be set to merge.

PR #1643, opened 2020-04-22
chore(dependencies): Autobump front50Version

spinnakerbot:


ezimanyi:
Doing this in #1640, which makes some changes required for the dependency bump.

PR #1644, opened 2020-04-23
fix(versions): Use thread safe yaml parser

german-muzquiz:
According to yaml parser [documentation](https://bitbucket.org/asomov/snakeyaml/wiki/Documentation#markdown-header-threading), instances of `org.yaml.snakeyaml.Yaml` are not thread safe. This means that if several threads try to use the same instance for yaml Marshall/Unmarshall at the same time, operations may fail.

This PR changes the shared `yamlParser` bean to scope `prototype` and uses the application context to get instances of the bean each time it's used.


german-muzquiz:
cc @ezimanyi

german-muzquiz:
Thanks @ezimanyi , could you add the label for me? I don't seem to have permissions for adding labels

PR #1645, opened 2020-04-27
fix(cli): replace artifacts commands

maggieneterval:
Partially reverts: https://github.com/spinnaker/halyard/pull/1620

Given the confusion on Slack and GitHub, let's replace the `artifacts` and `artifacts-rewrite` commands until no supported Spinnaker release consumes these flags.  Users on 1.20 will still get the `tooHighMessage` letting them know that these parameters are no longer necessary.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -6354,6 +6354,8 @@ hal config features edit [parameters]
 ```
 
 #### Parameters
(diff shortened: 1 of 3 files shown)


maggieneterval:

PR #1653, opened 2020-04-29
fix(daemon): Fix startup failure by reverting incorrect ps usage

ezimanyi:
Fixes spinnaker/spinnaker#5705

I'm not sure what I was thinking with this original PR; adding the -e flag will return all processes regardless of PID, and is not necessary when passing a PID.

This is causing halyard to fail to startup because we think that the daemon is always running.

This reverts commit 4a686e906fbfce8ef565c3c1aeadffa74ad6535e.


ezimanyi:

PR #1619, opened 2020-04-15
feat(cli): remove Travis, Wercker, Gremlin, and InfrastructureStages feature flags

maggieneterval:
We [replaced](https://github.com/spinnaker/deck/pull/8175) Deck's stage-specific feature flags with a single `hiddenStages` setting , so we should remove their corresponding Halyard commands.

* feat(cli): remove Travis feature flag 

* feat(cli): remove Wercker feature flag 

* feat(cli): remove Gremlin feature flag 

* feat(cli): remove infrastructureStages feature flag

Corresponding docs PR: https://github.com/spinnaker/spinnaker.github.io/pull/1807

PR #1655, opened 2020-04-30
fix(cli): replace gremlin commands for 1.18 and 1.19 compatibility

maggieneterval:
Partially reverts https://github.com/spinnaker/halyard/pull/1619

We need to support Gremlin in Halyard until we release Spinnaker 1.22, because 1.19 Orca still reads this flag.


maggieneterval:

PR #1662, opened 2020-05-06
fix(front50): Remove duplicate spinnaker key

link108:
Halyard is generating duplicate, nested spinnaker keys in the front50 configs, this removes the duplicated key


link108:
@ezimanyi ah good catch, I updated to fix that issue. I tested with a spinnaker version that does support plugins (1.19.6) and one that does not (1.18.3) 👍

ezimanyi:
@link108 : Looks like you'll have to merge in `master` yourself as the branch doesn't allow edits.  It should auto-merge once you've done that.

PR #1672, opened 2020-05-11
refactor(deploy): Make kubectl deployer the only option for a distributed installation on Kubernetes 

maggieneterval:
* refactor(deploy): remove clouddriver-kubernetes dependency from KubernetesService 

  KubernetesService currently calls out to clouddriver-kubernetes-v1 to assemble an image registry, repository, and tag into a KubernetesImageDescription object, and then assemble those components into a Docker image ID. Since clouddriver-kubernetes-v1 will not be included in the next Clouddriver release, let's instead include the relevant string formatting logic inline.

* refactor(kubernetes): remove Kubernetes V1 account kubeconfig validation 

  The legacy Kubernetes (V1) provider is deprecated, and will not be included in Spinnaker 1.21. Given that no new legacy accounts should be added, let's remove the V1-specific kube config validation that we never applied to V2 accounts anyway. This will unblock us from completely decoupling Halyard's distributed installation and Clouddriver's Kubernetes provider. This commit also updates my TODO to remove the remaining V1 validation logic when no supported Spinnaker release includes the V1 provider (tracking this work in this [GitHub issue](https://github.com/spinnaker/spinnaker/issues/5749)).

* refactor(deploy): Make kubectl deployer the only option for a distributed installation on Kubernetes 

  Spinnaker 1.21 will not include support for using legacy (V1) accounts to deploy to Kubernetes. The next release of Halyard will not include support for using a legacy (V1) account to deploy Spinnaker itself using Halyard. It should be relatively straightforward for users to migrate from the V1 deployer to the kubectl deployer, and can be done independently of migrating any pipelines or other Spinnaker accounts to the V2 provider. Given that we don't publish a Halyard changelog, I will be sure to communicate this in the #halyard Slack channel when we do the next Halyard release.

* refactor(deps): remove dependency on clouddriver-kubernetes 

  The previous commits removed all clients of clouddriver-kubernetes and clouddriver-kubernetes-v1. Moving forward, this will prevent the class of bug where contributors do not realize that breaking changes to Clouddriver could impact Halyard (at least in the Kubernetes provider).


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/distributed/kubernetes/KubernetesService.java
```diff
@@ -18,8 +18,7 @@
 
 import static com.netflix.spinnaker.halyard.config.model.v1.node.DeploymentEnvironment.ImageVariant.SLIM;
 
-import com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.KubernetesUtil;
-import com.netflix.spinnaker.clouddriver.kubernetes.v1.deploy.description.servergroup.KubernetesImageDescription;
+import com.google.common.base.Strings;
```
ezimanyi, line 21: 🥳

halyard-config/halyard-config.gradle
```diff
@@ -4,7 +4,6 @@ dependencies {
 
   implementation "com.netflix.spinnaker.clouddriver:clouddriver-docker:$clouddriverVersion"
   implementation "com.netflix.spinnaker.clouddriver:clouddriver-google:$clouddriverVersion"
-  implementation "com.netflix.spinnaker.clouddriver:clouddriver-kubernetes-v1:$clouddriverVersion"
```
ezimanyi, line 7: 🥳 🥳 🥳 🥳 💯

PR #1678, opened 2020-05-13
fix(plugins): remove edit plugin command.

edgarulg:
Currently, the command `hal plugins edit` return null and is not working, the `edit command` did not support updating the config for the plugin.  So let's just edit by hand for now.


halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/EditPluginCommand.java
```diff
@@ -1,65 +0,0 @@
...
-    for (String pluginExtensionName : extensions.split(",")) {
-      plugin
-          .getExtensions()
-          .putIfAbsent(pluginExtensionName, new PluginExtension().setId(pluginExtensionName));
-    }
-
-    new OperationHandler<Void>()
-        .setFailureMesssage("Failed to edit plugin " + plugin.getId() + ".")
-        .setSuccessMessage("Successfully edited plugin " + plugin.getId() + ".")
-        .setOperation(Daemon.setPlugin(currentDeployment, plugin.getId(), !noValidate, plugin))
```
danielpeach, line 62: Does anything else use `Daemon.setPlugin`? If not, let's rip it out and whatever else is exercised when you call it (there should be a corresponding controller method).

halyard-cli/src/test/groovy/com/netflix/spinnaker/halyard/cli/command/v1/CommandTreeSpec.groovy
```diff
@@ -38,10 +38,9 @@ import com.netflix.spinnaker.halyard.cli.command.v1.config.security.authn.x509.X
 import com.netflix.spinnaker.halyard.cli.command.v1.config.security.authz.AuthzCommand
 import com.netflix.spinnaker.halyard.cli.command.v1.config.security.ui.UiSecurityCommand
 import com.netflix.spinnaker.halyard.cli.command.v1.config.stats.StatsCommand
-import com.netflix.spinnaker.halyard.cli.command.v1.config.stats.StatsEnableDisableCommandBuilder
 import com.netflix.spinnaker.halyard.cli.command.v1.plugins.AddPluginCommand
 import com.netflix.spinnaker.halyard.cli.command.v1.plugins.DeletePluginCommand
-import com.netflix.spinnaker.halyard.cli.command.v1.plugins.EditPluginCommand
+
```
danielpeach, line 43: Please remove this line.

danielpeach:
Just want to point out the background on this - this code never worked, and we don't want to invest the time in supporting it.

halyard-cli/src/test/groovy/com/netflix/spinnaker/halyard/cli/command/v1/CommandTreeSpec.groovy
edgarulg: i removed it

halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/plugins/EditPluginCommand.java
edgarulg: The `Daemon.setPlugin` is not used by anything else, the only place where it is use is in the EditPluginCommand that I already removed

edgarulg: I removed the unused code for the `Daemon.setPlugin` on the service, also I removed the endpoint on the Web project that edit a plugin via Rest API.

maggieneterval:
@edgarulg @danielpeach CI for github.spinnaker.io is [failing at HEAD](https://github.com/spinnaker/spinnaker.github.io/runs/785406488) because there's a lingering reference to the now-non-existent edit plugin command; mind updating the plugin docs to remove that reference?

edgarulg:
Hello @maggieneterval my apologize for that, I already remove the documentation on this PR: https://github.com/spinnaker/spinnaker.github.io/pull/1885 Thanks for taking a look

PR #1692, opened 2020-05-29
refactor(kubernetes): Allow cacheThreads 0. This will disable cache for an account.

dcodix:
I am using 0 cacheThreads on one of my accounts to disable caching as it uses a lot of resources and fills redis up.
On previous halyard versions it did not complain, but now, setting it to 0 (zero) complains about it.
Deploying with `--no-validation` works fine, but it is not elegant.

If there is no other way to disable caching for some account, could you allow to use 0 (zero) ? It can still default to 1, but the use of 0 helped us a lot.


Emma-Black:
🙏 🙏 🙏

PR #1717, opened 2020-06-16
feat(travis): Support for numberOfJobs and buildResultLimit

costimuraru:
Fixes https://github.com/spinnaker/spinnaker/issues/5847

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -4417,10 +4417,12 @@ hal config ci travis master add MASTER [parameters]
 `MASTER`: The name of the master to operate on.
  * `--address`: (*Required*) The address of the travis API ([https://api.travis-ci.org](https://api.travis-ci.org)).
  * `--base-url`: (*Required*) The base URL to the travis UI ([https://travis-ci.org](https://travis-ci.org)).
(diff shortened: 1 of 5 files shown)


costimuraru:
CC @jervi @gardleopard @ezimanyi

jervi:
Thank you @costimuraru!
FYI, there’s this lingering PR from a year ago https://github.com/spinnaker/halyard/pull/1355 that never got merged. That PR doesn’t include the `buildResultLimit`, so I think it can be replaced by this PR instead. But I guess the comments made on the old PR are still valid, but support for Spinnaker <1.17 is maybe not that important anymore? WDYT @plumpy?

costimuraru:
Thanks, @jervi. I followed the discussion on your initial PR, that's why I decided to also keep the `numberOfRepositories` param (to maintain compatibility with Spinnaker < 1.17). 
I believe we should keep the old param for a while to not break integrations (it's already marked as deprecated). We could remove it in the future along with the code from Igor [1].


[1] https://github.com/spinnaker/igor/blob/master/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisProperties.java#L76-L80

PR #1730, opened 2020-06-24
feat(travis): Support for filteredRepositories parameter

costimuraru:
Depends on https://github.com/spinnaker/igor/pull/790

### DIFF
--- a/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/travis/TravisMaster.java
+++ b/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/travis/TravisMaster.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.halyard.config.model.v1.node.CIAccount;
 import com.netflix.spinnaker.halyard.config.model.v1.node.NodeIterator;
 import com.netflix.spinnaker.halyard.config.model.v1.node.NodeIteratorFactory;
 import com.netflix.spinnaker.halyard.config.model.v1.node.Secret;
+import java.util.List;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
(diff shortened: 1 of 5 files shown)


costimuraru:
Thanks, @maggieneterval. Added the annotation @ValidForSpinnakerVersion, as suggested.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/ci/travis/TravisMaster.java
```diff
@@ -37,4 +38,5 @@ public NodeIterator getChildren() {
   private Integer numberOfRepositories;
   private Integer numberOfJobs;
   private Integer buildResultLimit;
+  private List<String> filteredRepositories;
```
maggieneterval, line 41: I think you want the `ValidForSpinnakerVersion` annotation here rather than duplicated in the add/edit commands. Were you able to test this by building the daemon and CLI locally"

costimuraru: Oops, yeah, thanks for the heads up. Made the change.
I tested locally:

```bash
❯ ./hal config ci travis master add myaddress2 --filtered-repositories alloy --address example.com --base-url test.com

+ Get current deployment
  Success
+ Add the myaddress2 master
  Success
Validation in default.ci.travis.myaddress2:
- WARNING Field TravisMaster.filteredRepositories not supported for
  Spinnaker version 1.21.0: Filtered repositories is not supported prior to this
  release.
? Use at least 1.22.0 (It may not have been released yet).
```

Will/does kleat also have support for this feature? How about deprecating fields and marking them as obsolete for spinnaker versions > x?

maggieneterval: Great question! I think the tentative plan once kleat is GA is to just version it along with the top-level Spinnaker version, so that all supported fields in that version of kleat are guaranteed to be supported in that version of Spinnaker. But if you have any thoughts please post in the #kleat Slack channel, we are very interested in all feedback!

PR #1737, opened 2020-06-26
fix(local-bom): Fix issues blocking use of locally hosted configuration files

jaydee864:
Addresses bug with local BOM deployments where the 'local:' is not pruned from the version string before converting to a SemVer object. See https://github.com/spinnaker/spinnaker/issues/5703 for details.

Also addresses bug with orderBySemVer where versions which were supposed to be treated as equivalent were not actually being treated as such if they weren't the exact same string. For example, according to the spec, 1.0.0-00000 is supposed to be equal to 1.0.0-99999 since lessThan is not supposed to care about the part after the hyphen. However, 1.0.0-00000 was treated as lessThan 1.0.0-99999 because of the secondary naturalOrder sorting.

Also addresses bug with LocalDiskProfileReader where it does not add the ".tar.gz" file extension to profileName when readArchiveProfile like GoogleProfileReader does


glauco:
I came across this same problem today.
It would be great to have this fix available on the next halyard version.

PR #1741, opened 2020-07-06
feat(authn/saml): New optional SAML parameter for signing messages

german-muzquiz:
Halyard changes for https://github.com/spinnaker/gate/pull/1269

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -11177,6 +11177,7 @@ hal config security authn saml edit [parameters]
  * `--metadata`: The address to your identity provider's metadata XML file. This can be a URL or the path of a local file.
  * `--no-validate`: (*Default*: `false`) Skip validation.
  * `--service-address-url`: The address of the Gate server that will be accesible by the SAML identity provider. This should be the full URL, including port, e.g. [https://gate.org.com:8084/](https://gate.org.com:8084/). If deployed behind a load balancer, this would be the laod balancer's address.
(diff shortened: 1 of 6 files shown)


german-muzquiz:
cc @ezimanyi maybe this will need to be added to kleat

ezimanyi:
@german-muzquiz : Thanks for thinking about `kleat`!  Yes, I definitely think this should also be added to `kleat` to keep its feature parity with Halyard.

Is this something you'd want to do as a first commit to `kleat`?  I'm happy to help with this as well.  I think for `kleat` we'd want to make an `enum` of available encryption types in the `proto`...here's a [good example](https://github.com/spinnaker/kleat/blob/1754772f46c6c8957cb68e8832e84a2862548eb4/api/proto/storage/s3.proto#L47).

german-muzquiz:
> @german-muzquiz : Thanks for thinking about `kleat`! Yes, I definitely think this should also be added to `kleat` to keep its feature parity with Halyard.
> 
> Is this something you'd want to do as a first commit to `kleat`? I'm happy to help with this as well. I think for `kleat` we'd want to make an `enum` of available encryption types in the `proto`...here's a [good example](https://github.com/spinnaker/kleat/blob/1754772f46c6c8957cb68e8832e84a2862548eb4/api/proto/storage/s3.proto#L47).

Created kleat PR :)

PR #1742, opened 2020-07-06
refactor(kubernetes): deprecate ProviderVersion

maggieneterval:
Related to: https://github.com/spinnaker/spinnaker/issues/5749

* Since ProviderVersion was only ever relevant to KubernetesAccount, move the enum from Provider to KubernetesAccount before deprecating it. Hopefully this will prevent any future cloud providers from needlessly including the providerVersion field in every account.
* We have included warnings that V1 accounts cannot be used for a Distributed install for a couple of Halyard releases; let's remove these for the next release.


maggieneterval:
> So the plan then is that in the next version of Halyard, you can still use the --provider-version flag to set an account to V1 (if deploying a version of Spinnaker before the deprecation)

💯 ✅ 🙏 

> then once no supported versions still support V1 we'll remove that flag entirely?

I'm actually leaning towards letting it linger indefinitely, since it isn't actually very harmful as Halyard-isolated tech debt. Originally I wanted to document a policy of "A given Halyard release is only guaranteed to work with a Spinnaker version that was supported at the time of its release" but without additional Halyard functionality to enforce that, it doesn't seem worth it to potentially confuse users this late in Halyard's life (sorry, Halyard). Instead I'm working on improving the existing contributor docs on deprecation to better document the `@ValidForSpinnakerVersion` annotation as the way to deprecate config; PR coming soon!

ezimanyi:
Right, of course...at this point in Halyard's lifecycle leaving the parameter with `@ValidForSpinnakerVersion` makes perfect sense!

PR #1744, opened 2020-07-06
feat(docs): add deprecation documentation

maggieneterval:
Related to: https://github.com/spinnaker/spinnaker/issues/5757, https://github.com/spinnaker/spinnaker/issues/5749

Given that Halyard investment is waning, let's document `ValidForSpinnakerVersion` as the path to field deprecation, rather than adding a policy of only supporting fields that exist in a supported Spinnaker version (as described in the above GitHub issues).

* feat(docs): improve discoverability of contributor docs 

* feat(docs): add field deprecation docs 

* feat(docs): document how to add minimum Spinnaker version to new fields

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,3 +64,12 @@ make
 > WARNING! Since this project uses one gradle project for two executables, you
 > currently can't run `make` and `./gradlew` at the same time (one of the two
 > will likely hang indefinitely).
+
+## Adding Functionality
+
+* [Adding a config parameter](/docs/adding-a-config-parameter.md)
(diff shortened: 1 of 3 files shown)


maggieneterval:
Yeah I thought about adding that as well...I guess the only case we'd want folks doing that is if they discover a field that has never been read by any known Spinnaker version (which sounds crazy, but I think we've each discovered several over the years). Slightly in favor of not documenting just to further discourage it? But could definitely be persuaded otherwise.

ezimanyi:
I think I'm convinced to leave out documenting that case!

PR #1747, opened 2020-07-07
fix(cli): ignore providerVersion in all non-Kubernetes account types

maggieneterval:
Instead of ignoring `providerVersion` in the base Account class, ignore it in each non-Kubernetes subclass, so that `providerVersion: v1` is respected for legacy Kubernetes accounts.


maggieneterval:
Remember that time I said deleting fields was so dangerous we shouldn't even document it? 🙈

PR #1762, opened 2020-07-17
fix(install script): Don't be so particular about Java versions

plumpy:
We should be able to run with any version of Java >=8. Java 7 has been EOL for over five years now, so if they still have that installed as their primary JDK, we probably have other issues. Let's just skip the version check entirely.

There are many valid ways someone could install Java. I think it's a fair assumption that someone installing Halyard should know how to install Java. So let's remove the part where we try to install an unnecessarily old version for them.

@maggieneterval (as build cop): there's some chance that the end-to-end tests are depending on `InstallHalyard.sh` to install Java. I don't really know how to figure that out, so I'll just try to fix it ASAP if it all blows up.

Fixes GoogleCloudPlatform/spinnaker-for-gcp#196.

### DIFF
--- a/install/debian/InstallHalyard.sh
+++ b/install/debian/InstallHalyard.sh
@@ -73,7 +73,7 @@ function process_args() {
         ;;
       *)
         echo "ERROR: Unknown argument '$key'"
-        exit -1
+        exit 1
     esac
   done
 }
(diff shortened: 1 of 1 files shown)


plumpy:
Oh great, thanks so much for the pointers! I'll fix the deploy script before I merge this.

GCP just seems to offer the base OS VM images. I think I'll probably just stick with those and add a command to install Java, like you suggest.

PR #1763, opened 2020-07-20
feat(kubernetes): allow adding custom service annotations

KeisukeYamashita:
## What

This allows adding customized annotations to Kubernetes service.

## Why

There are many cases that we want to add annotations to Kubernetes Service.
One example is Google Cloud Platform's [Network endpoint group (NEG)](https://cloud.google.com/load-balancing/docs/negs). This required to add a NEG configuration in Kubernetes Service annotation.
But currently, halyard doesn't allow to add it. It allows only `podLabels`, `podAnnotations`, `serviceLabels`. (ref: [podAnnotations, podLabels and serviceLabels, Spinnaker](https://spinnaker.io/reference/halyard/custom/#podannotations-podlabels-and-servicelabels))
Therefore, I've added to use this in many cases.

After this PR has merged, I will like to update the spinnaker.github.io document as well.


KeisukeYamashita:
@ezimanyi  Thank you for your review! 

Yeah, I am a big lover of kleat.
But it is still under development so I will continue to use Haylard in the meantime. That's why I raise this PR😊

With kleat, we can easy enable NEG. Super great🎉

PR #1769, opened 2020-08-06
chore(dependencies): Autobump korkVersion

spinnakerbot:


plumpy:
@srekapalli looks like this was caused by spinnaker/halyard#735

srekapalli:
> @srekapalli looks like this was caused by [spinnaker/kork#735](https://github.com/spinnaker/kork/pull/735)

Fix for this: https://github.com/spinnaker/halyard/pull/1770

PR #1782, opened 2020-08-27
chore(dependencies): Autobump clouddriverVersion

spinnakerbot:


ezimanyi:
This is failing because Halyard consumes clouddriver as a library, and the removal of Azure from clouddriver in #4809 now breaks the parts of Halyard that are doing Azure account validation and require `clouddriver-azure`.

Normally the fix would be to also remove Azure from clouddriver, but given that there is an approved Azure SIG now (spinnaker/governance#172) I expect that we'll re-instate Azure to clouddriver soon, which will resolve this issue.  So based on that I'm going to just wait at this point until we can bump to a clouddriver version that again has Azure.

PR #1787, opened 2020-09-01
fix(5995): Fix non-existent nexus node in halyard config

jmalvarezf-lmes:
Hi,

This should be fixing https://github.com/spinnaker/spinnaker/issues/5995.

I have tested it in our pre-production environment and it works.

Thanks,


ezimanyi (approved):
Thanks!

PR #1795, opened 2020-09-17
feat(cli): add upperBound for liveManifestCalls flag

maggieneterval:
The `liveManifestCalls` flag was [removed](https://github.com/spinnaker/clouddriver/pull/4904) from Clouddriver; update Halyard to indicate that it is no longer necessary when upgrading to 1.23.


ezimanyi (approved):
🆙 🐇 🎏 💯

PR #1806, opened 2020-10-02
chore(redis): Move Halyard's redis image into the CDF-owned GCP project

plumpy:
The `kubernetes-spinnaker` project seemingly exists solely to host this one image. I have no idea why. Let's move it into its own Artifact Registry repository in the CDF-owned `spinnaker-community` project.

Also, the version of Redis in this image is 3.2.11, which is three years old, but I suppose that's a problem for another day.


maggieneterval (approved):
⛵ 🔴 📸 ➡️ ☁️ 💻 📁 💯 🍡

PR #1813, opened 2020-10-16
fix(halyard): Bugfix 6112: Populate settings.js with default settings for Oracle provider

anupmpatil:
fix(halyard): settings.js doesn't have Oracle default settings
This settings.js is being used as template by halyard and gets hydrated with default settings. Without these settings, oracle default settings were not getting populated, causing at least one know failure while creating OCI load balancer template.

After this fix, default settings for oracle provider will be populated properly

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/deck/DeckProfileFactory.java
@@ -143,6 +143,10 @@ public class DeckProfileFactory extends RegistryBackedProfileFactory {
     bindings.put("kubernetes.default.namespace", "default");
     bindings.put("kubernetes.default.proxy", "localhost:8001");
 
+    bindings.put("oracle.default.account", "Default");
+    bindings.put("oracle.default.bakeryRegions", "us-phoenix-1");
+    bindings.put("oracle.default.region", "us-phoenix-1");
+
     // Configure GCE
     GoogleProvider googleProvider = deploymentConfiguration.getProviders().getGoogle();
     bindings.put("google.default.account", googleProvider.getPrimaryAccount());


anupmpatil:
@maggieneterval Could you please review this PR?

maggieneterval:
Hi @anupmpatil! We've been trying to move config defaults into the microservices themselves rather than have Halyard supply them. Can you please set these defaults in Deck's Oracle provider instead?

Feel free to follow the [Kleat](https://github.com/spinnaker/kleat) repo and join the #kleat Spinnaker Slack channel for updates on Halyard's replacement (no definite timeline yet).

Thanks!

anupmpatil:
@maggieneterval Thank you for the explanation, I have made changes in Deck's Oracle provider, I will ensure if any more changes are needed in Deck. So, just to get more clarification regarding this fix, do we still need to go ahead and merge it?

maggieneterval:
@anupmpatil Thanks for letting me know; if you've already handled these defaults in Deck's Oracle provider then I'll go ahead and close this PR.

anupmpatil:
@maggieneterval 
Till the alternative service comes up, the users who will (or are still using) use halyard with oracle provider are still going to see the issues because they will not have default settings and will see the same bug. I have 2 questions regarding this:
1. Can we request to let this change in and in case if it does goes in, are you planning any new halyard version release that accommodate this?
2. In case this change does not go in, we have work around using the settings-local.js configuration. We can create document explaining the steps needed for this work around. Where should we publish this document, so that users trying to use oracle provider with halyard can clearly know how they should configure settings-local.js?

maggieneterval:
> Till the alternative service comes up, the users who will (or are still using) use halyard with oracle provider are still going to see the issues because they will not have default settings

Sorry if my previous comment wasn't totally clear; I meant that we should handle setting these Oracle default settings in Deck itself (wherever we are reading them in the Deck code), rather than having Halyard write them or having users configure them in their `settings-local.js`.

Kleat will also not set any defaults for users; see [here](https://github.com/spinnaker/kleat#defaulting) for more details.

PR #1819, opened 2020-11-06
feat(validations/cf): Updated CF client for validations

german-muzquiz:
Adjusted CloudFoundry validations due to changes on CF code used to query spaces introduced by https://github.com/spinnaker/clouddriver/pull/5071

Now the validator doesn't depend on clouddriver internal objects, and instead depends on retrofit API interfaces, still exposed by clouddriver.


zachsmith1 (approved):
LGTM! thanks for making this more resilient to change

PR #1820, opened 2020-11-06
fix(boms): change default docker registry to spinnaker-community

plumpy:


halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/WriteableProfileRegistryProperties.java
```diff
@@ -24,5 +24,5 @@
 @ConfigurationProperties("spinnaker.config.input.gcs")
 public class WriteableProfileRegistryProperties {
   private String jsonPath = "";
-  private String project = "spinnaker-marketplace";
+  private String project = "spinnaker-community";
```
ezimanyi, line 27: I was confused by where we were using this...it looks unused (even since it was introduced in #201).  I guess it's not really worth deleting the variable here and it doesn't hurt to just re-point it to `spinnaker-community`.

ezimanyi (approved):

halyard-core/src/main/java/com/netflix/spinnaker/halyard/core/registry/v1/WriteableProfileRegistryProperties.java
plumpy: That was exactly my thought. I was going to delete it, but I figured why risk breaking something. Whereas if there _is_ something still using the `spinnaker-marketplace` project, I guess I'm okay with breaking that because I'd really like to know.

Sorry, I should have written this in the PR message.

PR #1848, opened 2021-02-10
chore(dependencies): Autobump Front50 Version

link108:

PR #1866, opened 2021-05-10
fix(gcs): Fix validation error when bucketLocation is not specified

cristhian-castaneda:
When the Front50 version was updated #1848 a change for GcsStorageService was introduced.

Prior to the rewrite of GcsStorageService, front50 accepted either null or empty string as the bucket location; now that GcsStorageService is in kotlin and does not use a String? for the field, an error occurs on trying to create the GcsStorageService.

PRs related spinnaker/front50#934


tunderwood (approved):

link108:
@mergifyio backport release-1.42.0

link108:
@Mergifyio backport release-1.42.x

PR #1880, opened 2021-07-23
fix(oauth2): add roles to user info mapping

estebangarcia:
When using an "OTHER" oauth2 provider we need to specify the user roles mapping field.

[Gate supports this](https://github.com/spinnaker/gate/blob/master/gate-oauth2/src/main/groovy/com/netflix/spinnaker/gate/security/oauth2/OAuth2SsoConfig.groovy#L111) but halyard doesn't so it has to be configured with a profile file.

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/oauth2/EditOAuth2Command.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/config/security/authn/oauth2/EditOAuth2Command.java
@@ -81,6 +81,11 @@ public class EditOAuth2Command extends AbstractEditAuthnMethodCommand<OAuth2> {
   private String userInfoMappingUsername;
 
   @Parameter(
+      names = "--user-info-mapping-roles",
+      description = "The roles field returned from your OAuth provider.")
+  private String userInfoMappingRoles;
+
+  @Parameter(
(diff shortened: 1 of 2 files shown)


VayshaYsno:
Any updates? There is still no option to pass roles through Oauth2...

VayshaYsno (approved):

VayshaYsno:
Found a solution! Spinnaker allows you to override it's configuration through `*-local.yml` files. All you have to do is to configure connection part in `config`, but `userInfoMapping` and `scope` in `gate-local.yml`, like this:

```
config:
   security:
      authn:
         oauth2:
            enabled: true
            client:
               clientId: XXXXXX
               clientSecret: XXXXXXXX
               accessTokenUri: XXXXXXX
               userAuthorizationUri: XXXXXXX
            resource:
               userInfoUri: XXXXXX
            userInfoMapping: {}
            provider: OTHER
```

And this part in `gate-local.yml`:

```
security:
   oauth2:
      userInfoMapping:
         email: email
         firstName: given_name
         lastName: family_name
         username: username
         roles: groups
      client:
         scope: opened,email,profile,groups
```
And if you will login, and check your Spinnaker Api like this: `https://domain-of-api.example/auth/user`, you will see that `roles:` are passed successfully now!

Note, that your mappers can be different of described above. I used to do it with `k8s` and Keycloak as OpenID idp.
Hope it will be helpful for someone!

PR #1953, opened 2022-05-19
Fix bugs in systemd service preventing halyard daemon from starting

ashleykleynhans:
1. Fixed bugs in systemd service.
2. Create spinnaker user in postinst script if it does not exist.
3. Set default user in postinst script to spinnaker instead of ubuntu.

### DIFF
--- a/halyard-web/lib/systemd/system/halyard.service
+++ b/halyard-web/lib/systemd/system/halyard.service
@@ -4,9 +4,9 @@ Description=A tool for managing the lifecycle of a Spinnaker installation
 [Service]
 Restart=always
 RestartSec=10
-Type=forking
-ExecStart=
-ExecStart=sudo -u $(cat /var/spinnaker/config/halyard-user) -g $(cat /var/spinnaker/config/halyard-user) /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &
+Type=simple
+ExecStart=/opt/halyard/bin/halyard
+User=spinnaker
 
 [Install]
 WantedBy=user.target
--- a/halyard-web/pkg_scripts/postInstall.sh
+++ b/halyard-web/pkg_scripts/postInstall.sh
@@ -5,13 +5,21 @@ echo '/opt/halyard/bin/hal "$@"' | sudo tee -a /usr/local/bin/hal > /dev/null
 
 chmod +x /usr/local/bin/hal
 
-HAL_USER=""
-if [ -f "/opt/spinnaker/config/halyard-user" ]; then
+if [ -f "/opt/spinnaker/config/halyard-user" ];
+then
   HAL_USER=$(cat /opt/spinnaker/config/halyard-user)
-fi
-
-if [ -z "$HAL_USER" ]; then
-  HAL_USER="ubuntu"
+else
+  HAL_USER=spinnaker
+  getent passwd spinnaker > /dev/null
+  if [ $? -gt 0 ]; then
+    useradd -s /bin/bash $HAL_USER
+  fi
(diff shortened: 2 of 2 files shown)


halyard-web/lib/systemd/system/halyard.service
```diff
@@ -4,9 +4,9 @@ Description=A tool for managing the lifecycle of a Spinnaker installation
 [Service]
 Restart=always
 RestartSec=10
-Type=forking
-ExecStart=
-ExecStart=sudo -u $(cat /var/spinnaker/config/halyard-user) -g $(cat /var/spinnaker/config/halyard-user) /opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log &
+Type=simple
+ExecStart=/opt/halyard/bin/halyard
```
kskewes-sf, line 8: Looks good, where does Halyard log to by default? stdout?

Do we still need this: `/opt/halyard/bin/halyard 2>&1 > /var/log/spinnaker/halyard/halyard.log` ?

kskewes-sf: Or does `journald` pick it up?

halyard-web/pkg_scripts/postInstall.sh
```diff
@@ -11,7 +11,11 @@ if [ -f "/opt/spinnaker/config/halyard-user" ]; then
 fi
 
 if [ -z "$HAL_USER" ]; then
-  HAL_USER="ubuntu"
+  getent passwd spinnaker > /dev/null
```
kskewes-sf, line 14: Great. 
It seems various scripts in the repo all support supplying a user for Halyard to run as.
`spinnaker` is a sensible default, matching other services like clouddriver, etc.

Do we need to also write the user into `/opt/spinnaker/config/halyard-user` ?
there look to be a few "update" related scripts in the repo that might depend on this as well?

Thanks again for tackling this. Big help.

halyard-web/lib/systemd/system/halyard.service
ashleykleynhans: The systemd syntax does dot allow redirection of output to a log file, redirecting the output to the log file was one of the issues that was causing the service not to start.

https://manpages.ubuntu.com/manpages/bionic/en/man5/systemd.service.5.html

```
 This syntax is inspired by shell syntax, but only the meta-characters and expansions
       described in the following paragraphs are understood, and the expansion of variables is
       different. Specifically, redirection using "<", "<<", ">", and ">>", pipes using "|",
       running programs in the background using "&", and other elements of shell syntax are not
       supported.
       ```
       
 But as you mentioned, journald picks it up and the logs can be viewed using `journalctl`.

File: halyard-web/pkg_scripts/postInstall.sh (hunk 2 above)
Comment (ashleykleynhans) [reply halyard-web/pkg_scripts/postInstall.sh:13]: I've now updated my changes to include this.

File: halyard-web/pkg_scripts/postInstall.sh [hunk 3]
```diff
@@ -5,13 +5,21 @@ echo '/opt/halyard/bin/hal "$@"' | sudo tee -a /usr/local/bin/hal > /dev/null
 
 chmod +x /usr/local/bin/hal
 
-HAL_USER=""
```
kskewes-sf, line 8: Shell is weird. 
Regardless of this removal `HAL_USER` will be available to `install ... --owner=$HAL_USER`.
This is because the variables are effectively global in bash unless defined in a function using `local` keyword.

kskewes-sf (commented):
LGTM. Thank you!

dbyron-sf (approved):

PR #1954, opened 2022-05-25
fix(provider/google): Added support to service account impersonation

kirangodishala:
This is to address the issue [ #6660](https://github.com/spinnaker/spinnaker/issues/6660)

Two new fields(serviceAccountId & serviceAccountProject) need to be added to google account fields in hal config which will be used in clouddriver for service account impersonation. 

Updated hal commands documentation as well.


sanopsmx (approved):
lgtm

PR #1955, opened 2022-05-26
fix(cli): Fix health response type

jfrabaute:
Hal /health endpoint can return:
{
  "status" : "UP",
  "groups" : [ "liveness", "readiness" ]
}

So the return type must be Map<String, Object>
instead of Map<String, String>

Fixes: https://github.com/spinnaker/spinnaker/issues/6655

### DIFF
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/Daemon.java
@@ -51,7 +51,17 @@ import retrofit.converter.JacksonConverter;
 @Slf4j
 public class Daemon {
   public static boolean isHealthy() {
-    return getService().getHealth().get("status").equalsIgnoreCase("up");
+    // Hal /health endpoint can return:
+    // {
+    //   "status" : "UP",
+    //   "groups" : [ "liveness", "readiness" ]
+    // }
+    // So the return type must be Map<String, Object>
+    Object status = getService().getHealth().get("status");
+    if (status instanceof String) {
+      return ((String) status).equalsIgnoreCase("up");
+    }
+    return false;
   }
 
   public static String shutdown() {
--- a/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
+++ b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonService.java
@@ -37,8 +37,14 @@ import retrofit.client.Response;
 import retrofit.http.*;
 
 public interface DaemonService {
+  // Hal /health endpoint can return:
+  // {
+  //   "status" : "UP",
+  //   "groups" : [ "liveness", "readiness" ]
+  // }
+  // So the return type must be Map<String, Object>
   @GET("/health")
-  Map<String, String> getHealth();
+  Map<String, Object> getHealth();
 
   @POST("/shutdown")
   Map<String, String> shutdown(@Body StringBodyRequest _ignore);


dbyron-sf:
These changes LGTM, thanks.  I don't see any existing automated tests for anything in this neighborhood.  Would you mind looking into adding one?  I'd love to be able to detect this kind of problem sooner in the future.  If it's too painful, I'm OK to merge this as is.

jfrabaute:
Hi

Thanks for the review.
I checked regarding adding tests. I didn't find any for the halyard-cli. I'm not familar with the code nor the framework, so I bet trying to add a test for this change will take a while.
Would it be possible to merge this as a first step?

Thanks.

dbyron-sf:
FWIW, I think a test for this would end up in halyard-web/src/test/java somewhere, but testing static methods is often painful.

The more I look into this, the more I'm torn.  It's https://spring.io/blog/2020/03/25/liveness-and-readiness-probes-with-spring-boot#exposing-kubernetes-probes-with-spring-boot-actuator that makes me wonder if a different approach is more appropriate.

From what I can tell, the `Daemon.isHealthy` method is only called from one place...[here](https://github.com/spinnaker/halyard/blob/351ed7521836ab8e74d1282681c511960efbd27b/halyard-cli/src/main/java/com/netflix/spinnaker/halyard/cli/command/v1/HalCommand.java#L93).  From
```
  @Parameter(
      names = "--ready",
      description =
          "Check if Halyard is up and running. Will exit with non-zero return code when it isn't.")
  private boolean healthy;
```
and specifically the name of the parameter (`--ready`), it seems pretty clear that the desired thing here is readiness...so wouldn't it be clearer to rename `Daemon.isHealthy` to `Daemon.isReady` (and the healthy private member of HalCommand to ready), and similarly rename `DaemonService.getHealth` to DaemonService.getReadiness`...and have it hit `/health/readiness` instead of `/health` ?  I haven't seen exactly what that endpoint returns.  We might still need the casting, but we might not...and if we do need the casting, it'll still be more clearer what's going on....

After a bit of testing though, I have a feeling there's something else going on...
In one terminal window:
```
$ docker run --name halyard --rm us-docker.pkg.dev/spinnaker-community/docker/halyard:1.45.0
```
and in another:
```
$ docker exec -it halyard curl http://localhost:8064/health
{
  "status" : "UP"
}
```
I haven't found a link to that OpsMx helm chart that https://github.com/spinnaker/spinnaker/issues/6655 references, but I suspect it's setting
```
management.health.probes.enabled=true
```
so I bet the way forward is to set that in [halyard-web/config/halyard.yml](https://github.com/spinnaker/halyard/blob/351ed7521836ab8e74d1282681c511960efbd27b/halyard-web/config/halyard.yml) as well...but all of this says to me that we really do need a test.

jfrabaute:
Hi,

Thanks for checking the details.
Yes, I didn't know/realize that the endpoint is also exposing /health/{readiness/liveness} (IIUC, I haven't checked).
I'm using a very old helm chart and the install pod which is connecting to halyard to generate the config is doing `hal --ready` in a loop into it succeeds.
As it keeps faling because of the se-serialization of the response on the hal client, the install is stuck in the loop.
That's why I first sent this change.
I agree with ouy that it would be better to fix this in the right way both on the client and server side.
I won't have the time to work on this in the next few days, but I'll try to revisit later when I have some time.

jfrabaute:
Thanks to your comment, I realized I could just disable the probes:

```
management.health.probes.enabled=false
```

And now deployment is working. So I have a workaround.

Regarding this PR, I'll close it for now as it's probably not the right thing to do.

jfrabaute:
We are using an old helm chart, and I fixed it.
The official helm chart is similar.
What is did is add a line here:
https://github.com/helm/charts/blob/master/stable/spinnaker/templates/configmap/halyard-init-script.yaml#L13

```
    printf 'management.health.probes.enabled: false\n' >> /tmp/config/halyard-local.yml
```

I (or a member of my team) will be able to send a PR for the chart with this change next week, if nobody has done it.

After adding this line, we could deploy with our old helm chart (We are planning to upgrade to the new helm chart, so we'll need this change).

Thanks.

jfrabaute:
@snorlaX-sleeps Is the new helm chart here?: https://github.com/OpsMx/spinnaker-helm/tree/main/charts/spinnaker

I'd like to send a PR for the fix, so I suppose I should send it to the OpsMx chart, correct?

snorlaX-sleeps:
@jfrabaute - we still use the stable chart, as there was never an official maintainer listed in the repo for the chart.
The OpsMx chart was mentioned in the original issue.

Have you noticed roll-outs progressing slower with HealthChecks disabled? The one installation I tried it with took a very long time, in comparison with the previous 3-4mins

jfrabaute:
> Have you noticed roll-outs progressing slower with HealthChecks disabled?
I haven't tried with HealthChecks disabled.

I'm not sure the stable chart is accepting PRs anymore. The rpeo is archived, so I guess, and that's why I was thinking about sending a PR for the OpsMx chart (We still use the stable chart as well, we probably nede to migrate to the OpsMx chart).

PR #1966, opened 2022-08-24
(fix)install: Cleaned up the Debian install script

ashleykleynhans:
1. Added the GPG signing key for the spinnaker-community repo when adding the repo.
2. Removed Java installation since Java would already need to be installed to run Halyard, and it is not possible to run `hal deploy apply` if the Halyard daemon is not running.
3. Fixed deprecation warnings in Redis installation script.

### DIFF
--- a/halyard-deploy/src/main/resources/services/redis/install.sh
+++ b/halyard-deploy/src/main/resources/services/redis/install.sh
@@ -1,2 +1,1 @@
-add-apt-repository -y ppa:chris-lea/redis-server
-apt-get -q -y --force-yes install redis-server={%version%} redis-tools={%version%}
+apt-get -q -y --allow-unauthenticated --allow-downgrades --allow-remove-essential --allow-change-held-packages install redis-server={%version%} redis-tools={%version%}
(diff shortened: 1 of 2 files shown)


halyard-deploy/src/main/resources/services/redis/install.sh
```diff
@@ -1,2 +1 @@
-add-apt-repository -y ppa:chris-lea/redis-server
```
ashleykleynhans, line 1: The install.sh script already handles this.

halyard-deploy/src/main/resources/services/redis/install.sh
```diff
@@ -1,2 +1 @@
...
-apt-get -q -y --force-yes install redis-server={%version%} redis-tools={%version%}
\ No newline at end of file
+apt-get -q -y --allow-unauthenticated --allow-downgrades --allow-remove-essential --allow-change-held-packages install redis-server={%version%} redis-tools={%version%}
```
ashleykleynhans, line 1: Fixes deprecation warnings for `--force-yes`

PR #1967, opened 2022-08-24
fix(install): Fixed various bugs in the Debian installation scripts

ashleykleynhans:
1. Fixed various issues with the Debian installation scripts.
2. Removed Java installation from the script that halyard runs when applying a deployment, because Java would already need to be installed to install Halyard and start the daemon.
3. Changed the minimum Ubuntu requirement from 14.04 to 18.04 since anything prior to 18.04 is out of support.
4. Removed option for specifying a user and home directory, since the Halyard package depends on the user being `spinnaker` and creates the `spinnaker` user if it does not exist.

### DIFF
--- a/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/debian/LocalDebianRedisService.java
+++ b/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/debian/LocalDebianRedisService.java
@@ -54,7 +54,7 @@ public class LocalDebianRedisService extends RedisService implements LocalDebian
 
   @Override
   public String installArtifactCommand(DeploymentDetails deploymentDetails) {
-    return "apt-get -q -y --force-yes install redis-server && (systemctl start redis-server.service || true)";
+    return "apt-get -q -y --allow-unauthenticated --allow-downgrades --allow-remove-essential --allow-change-held-packages install redis-server && (systemctl start redis-server.service || true)";
   }
 
   @Override
--- a/halyard-deploy/src/main/resources/debian/init.sh
+++ b/halyard-deploy/src/main/resources/debian/init.sh
@@ -1,23 +1,4 @@
-source /etc/os-release
-
-if [ "$VERSION_ID" = "14.04" ]; then
-  cat > /etc/init/spinnaker.conf <<EOL
-description "spinnaker"
-start on filesystem or runlevel [2345]
-stop on shutdown
-pre-start script
-  for i in {%services%}
-  do
-    if [ ! -d "/var/log/spinnaker/\$i" ]; then
-      echo "/var/log/spinnaker/\$i does not exist. Creating it..."
-      install --mode=755 --owner=spinnaker --group=spinnaker --directory /var/log/spinnaker/\$i
-    fi
-    service \$i start
-  done
-end script
-EOL
-else
-  cat > /lib/systemd/system/spinnaker.service <<EOL
+cat > /lib/systemd/system/spinnaker.service <<EOL
 [Unit]
 Description=All Spinnaker services
 After=network.target
@@ -29,4 +10,3 @@ RemainAfterExit=yes
 [Install]
 WantedBy=multi-user.target
 EOL
-fi
--- a/halyard-deploy/src/main/resources/debian/install.sh
+++ b/halyard-deploy/src/main/resources/debian/install.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 ## auto-generated debian install file written by halyard
+## which is executed when running 'hal deploy apply'.
 
 set -e
 set -o pipefail
@@ -8,7 +9,7 @@ set -o pipefail
 # install redis as a local service
 INSTALL_REDIS="{%install-redis%}"
 
-# install first-time spinnaker dependencies (java, setup apt repos)
+# install first-time spinnaker dependencies (setup apt repos)
 PREPARE_ENVIRONMENT="{%prepare-environment%}"
 
 REPOSITORY_URL="{%debian-repository%}"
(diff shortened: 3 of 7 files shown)


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/local/debian/LocalDebianRedisService.java
```diff
@@ -54,7 +54,7 @@ LogCollector getLocalLogCollector() {
 
   @Override
   public String installArtifactCommand(DeploymentDetails deploymentDetails) {
-    return "apt-get -q -y --force-yes install redis-server && (systemctl start redis-server.service || true)";
+    return "apt-get -q -y --allow-unauthenticated --allow-downgrades --allow-remove-essential --allow-change-held-packages install redis-server && (systemctl start redis-server.service || true)";
```
ashleykleynhans, line 57: Fixes deprecation warning for `--force-yes`

halyard-deploy/src/main/resources/debian/init.sh
```diff
@@ -1,23 +1,4 @@
-source /etc/os-release
-
-if [ "$VERSION_ID" = "14.04" ]; then
```
ashleykleynhans, line 3: Removed since this is for ancient distros that are out of support, only Ubuntu 18.04 LTS onwards still has support.

halyard-deploy/src/main/resources/debian/install.sh
```diff
@@ -40,74 +41,35 @@ else
...
-
-function add_java_apt_repository() {
-  # Only Ubuntu prior to 18.04 LTS requires this PPA
-  if [ "${DISTRIB_RELEASE%%.*}" -lt "18" ]; then
-    echo "Adding Java PPA repository for Ubuntu version less than 18.04"
-    add-apt-repository -y ppa:openjdk-r/ppa
-  fi
-}
-
-function install_java() {
```
ashleykleynhans, line 79: No longer needed because Java would have already have had to be installed in order to get Halyard installed, so if someone is able to run `hal deploy apply`, they would already have Java installed.

ashleykleynhans:
Fixes https://github.com/spinnaker/spinnaker/issues/6657

install/Installer.sh
```diff
@@ -288,35 +248,16 @@ function configure_bash_completion() {
...
+  echo "$(tput bold)Halconfig will be stored at /home/spinnaker/.hal/config$(tput sgr0)"
 
   mkdir -p /opt/spinnaker/config
   chmod +rx /opt/spinnaker/config
 
   cat > /opt/spinnaker/config/halyard.yml <<EOL
 halyard:
   halconfig:
-    directory: $halconfig_dir
+    directory: /home/spinnaker/.hal/config
```
dbyron-sf, line 260: Wasn't `$halconfig_dir` set to `/home/spinnaker/.hal` before this PR?

install/debian/InstallHalyard.sh
```diff
@@ -78,81 +54,29 @@ function process_args() {
...
 
   mkdir -p /opt/spinnaker/config
-  chmod a+rx /opt/spinnaker/config
+  chmod +rx /opt/spinnaker/config
 
   cat > /opt/spinnaker/config/halyard.yml <<EOL
 halyard:
   halconfig:
-    directory: $halconfig_dir
+    directory: /home/spinnaker/.hal/config
```
dbyron-sf, line 79: Same here.

install/debian/InstallHalyard.sh
```diff
@@ -181,23 +103,25 @@ if [ "\$yes" != "y" ] && [ "\$yes" != "Y" ]; then
...
 
+apt purge -y spinnaker-halyard
+
 rm /opt/halyard -rf
 rm /var/log/spinnaker/halyard -rf
 rm -f /usr/local/bin/hal /usr/local/bin/update-halyard
 
 echo "Deleting halconfig and artifacts"
 rm /opt/spinnaker/config/halyard* -rf
-rm $halconfig_dir -rf
```
dbyron-sf, line 190: Don't we still need this, or at least the equivalent without a $halfconfig_dir variable?  Seems tempting to leave that variable around...

install/Installer.sh
ashleykleynhans: Nope, the variable was set using the command line options that I removed, which is why I removed the variable and hard-coded it instead, since the Debian package installs it to /home/spinnaker anyway.

install/debian/InstallHalyard.sh
ashleykleynhans: Same reason as above,

install/Installer.sh
dbyron-sf: I was looking a few lines above at
```
local halconfig_dir="$home/.hal"
```
I'm all for hard-coding it.  I was really asking about the `/config` at the end since it seems like we didn't have it before.

ashleykleynhans: Fixed

install/debian/InstallHalyard.sh
ashleykleynhans: Fixed

ashleykleynhans:
@dbyron-sf Thanks for taking the time to review, it is very much appreciated.  The issues you pointed out have now been resolved, thank you and nicely spotted!

install/Installer.sh
```diff
@@ -328,67 +269,38 @@ spinnaker:
...
   echo "Aborted"
   exit 0
 fi
 
-apt-get purge spinnaker-halyard
+apt-get purge -y spinnaker-halyard
 
 echo "Deleting halconfig and artifacts"
 rm /opt/spinnaker/config/halyard* -rf
 rm $halconfig_dir -rf
```
dbyron-sf, line 291: I don't see anything that defines `$halconfig_dir` anymore.

ashleykleynhans: Apologies, I was  fast asleep when I tried to put it back 🤦 .  Its fixed now, thank you.

dbyron-sf (approved):

PR #1972, opened 2022-08-26
fix(halyard): cli boolean flags not working

sbeaulie:
Some of the boolean flags that are parameterized via jcommander are not working
See new unit test added that are failing with the current jcommander version
Parameters that are saved as private boolean like `--help` work, but not the
ones attached to more complex method_x(booalean y). There is not much documentation
in jcommander for those so I suspect a bug, since attaching them to methods with
a signature using Boolean instead of boolean do work.
This fixes issue reported https://github.com/spinnaker/spinnaker/issues/6700
in addition to `--quiet` it also fixes `--debug`, `--alpha` and `--color`


dbyron-sf (approved):

dbyron-sf:
Thank you @sbeaulie !

PR #1974, opened 2022-09-13
fix(install): Enable Halyard and Spinnaker services to start on reboot, and fixed Debian artifacts URL

ashleykleynhans:
### Resolves the following issues

1. Spinnaker and Halyard services are not enabled once they are installed, and do not start automatically on reboot.
2. The Halyard service is added as a dependency to a non-existent unit user.target in the systemd unit file.
3. The Debian artifact repository URL still points to Bintray.

### DIFF
--- a/halyard-web/lib/systemd/system/halyard.service
+++ b/halyard-web/lib/systemd/system/halyard.service
@@ -9,4 +9,4 @@ ExecStart=/opt/halyard/bin/halyard
 User=spinnaker
 
 [Install]
-WantedBy=user.target
+WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


halyard-web/lib/systemd/system/halyard.service
```diff
@@ -9,4 +9,4 @@ ExecStart=/opt/halyard/bin/halyard
 User=spinnaker
 
 [Install]
-WantedBy=user.target
+WantedBy=multi-user.target
```
ashleykleynhans, line 12: Resolves this issue:
```
Unit /lib/systemd/system/halyard.service is added as a dependency to a non-existent unit user.target.
```
Which also prevents the service from starting automatically on reboot.

dbyron-sf (approved):

PR #1984, opened 2022-10-11
Autobump clouddriver version master and use the same azure dependency as in clouddriver

ovidiupopa07:
- update imports
- add retrofit2 in `halyard-config.gradle` (before updating the library it was fetched from the azure-runtime library which is no longer available)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-clouddriverVersion=5.78.0
+clouddriverVersion=5.79.0
 fiatVersion=1.36.0
 korkVersion=7.149.0
 front50Version=2.27.0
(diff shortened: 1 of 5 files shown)


ovidiupopa07:
Hi @dbyron-sf.  I fixed the issue raised in the following PR: https://github.com/spinnaker/halyard/pull/1983
Not sure about the workflow, though, if it is ok or not that I opened the PR in this project

dbyron-sf:
Thanks much @ovidiupopa07 .  Making a PR like this is great.  I ended up cherry-picking the fixed into https://github.com/spinnaker/halyard/pull/1983 so it can all in together.

PR #1994, opened 2022-11-22
fix: accept GitHub prop overrides

boonware:
# Issue
https://github.com/spinnaker/spinnaker/issues/6766

# Details
Allow GitHub auth properties to be overridden when using the `github` OAuth2 provider. This allows users to use the `github` provider for configuring GitHub Enterprise, rather than having to use provider type `other`; using provider type `other` disables other features, see the parent issue for details.


boonware:
@dbyron-sf @ovidiupopa07 @diasjorge Could I get an approval for the build workflow, and a review?

ovidiupopa07:
> @dbyron-sf @ovidiupopa07 @diasjorge Could I get an approval for the build workflow, and a review?

Hi @boonware, I do not have permission to approve workflows. But I can help with the review 😃

boonware:
@ovidiupopa07 Thank you, much appreciated.

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java
```diff
@@ -200,4 +213,8 @@ public static Provider fromString(String name) {
               + Arrays.toString(Provider.values()));
     }
   }
+
+  private static boolean isSet(String config) {
+    return config != null;
```
ovidiupopa07, line 218: maybe you can reuse a common library like the one from apache or from the SpringFramework (StringUtils)

ovidiupopa07 (changes requested):
Are there any tests that you can add to verify this new behavior?

halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/security/OAuth2.java
mattgogerly: I don't think this method is necessary at all.. 

`if (isSet(x)) === if (x)`

PR #1996, opened 2022-11-30
fix(deck): Permissions for the custom profile must be the Apache user and not spinnaker user

ashleykleynhans:


halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/service/DeckService.java
```diff
@@ -22,10 +22,7 @@
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerArtifact;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.SpinnakerRuntimeSettings;
 import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.*;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.deck.ApachePassphraseProfileFactory;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.deck.ApachePortsProfileFactory;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.deck.ApacheSpinnakerProfileFactory;
-import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.deck.DeckProfileFactory;
+import com.netflix.spinnaker.halyard.deploy.spinnaker.v1.profile.deck.*;
```
dbyron-sf, line 25: Mind reverting to the non-* imports please?

dbyron-sf:
Would love to see info somewhere (e.g. a comment in the code, or in the commit message) about why the perms for the custom profile need to be Apache.  In some ways it seems obvious, at least at this moment...but I suspect that's going to fade with time.  In one version of the world, we'd have an automated test here that would fail until we got the permissions right, but I suspect that's a ton of work and so not worth doing given that there's still a glimmer of hope that halyard will disappear.

ashleykleynhans:
Apologies, I messed up this PR and closed it by accident, so I have created a new one.

I've added comments to the new PR regarding the Apache user as suggested.

dbyron-sf:
OK, looking at https://github.com/spinnaker/halyard/pull/1997 now...

dbyron-sf:
And thank you!!

PR #2099, opened 2023-10-17
chore(dependencies): Autobump korkVersion

spinnakerbot:


dbyron-sf (changes requested):
needs some other changes to go with it

PR #2181, opened 2024-09-09
refactor(dependency): replace groovy coordinates during upgrade of groovy 4.x

j-sandy:
Replacing the groovy coordinates from `org.codehaus.groovy` to `org.apache.groovy` supported by groovy 4.x and above versions.


dbyron-sf:
included in https://github.com/spinnaker/halyard/pull/2182

PR #2189, opened 2024-10-23
chore(java17): Upgrade java runtime to JRE17 and compilation using JDK17, but target remains Java11

j-sandy:


dbyron-sf:
@mattgogerly how does this look?

Dockerfile.slim
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.11
+FROM python:3.7-alpine3.16
```
dbyron-sf, line 1: The changes here are the only diffs from https://github.com/spinnaker/halyard/pull/2171

dbyron-sf (approved):

